PR #39, opened 2015-04-06
Add some basic transitions that can be added to any element.

jeffz-optimizely:
@tomgenoni Finally added some basic transition helper classes to LEGO, as we discussed months ago. I did my best guess regarding where in LEGO they should live, so if you think they don't belong in base let me know. I also wasn't sure if I should have placeholders or mixins alongside these classes, so if you think those are useful let me know.


tomgenoni:
I'm going to close this since we discussed plans for animations in the coming months.

jeffz-optimizely:
@tomgenoni good call.

PR #42, opened 2015-04-09
Create a Compass Polyfill for Libsass Compatibility

dtothefp:
This PR addresses making Lego Core Libsass compatible

- created an index.js and reference as `main` in package.json. This file exports `includePath` for use in grunt/gulp/webpack sass and exports a deprecated legoJs path.
- used [Bourbon](http://bourbon.io/) mixins and one Bourbon function to polyfill missing Compass mixins/functions

This is probably not the file directory structure that is correct but the functionality is there. If you would like to test lego core functionality with grunt/gulp/webpack sass see [this repo](https://github.com/dtothefp/lego-libsass-gulp-grunt-webpack)

@tomgenoni @chestone @danoc Please have a look and let me know what you think.

Not sure about the `@charset 'UTF-8'` it was in the Bourbon mixin files so I left it for now.

#### Update

- Added `box-sizing` mixin for Libsass compatibility with lego desktop

Libsass chokes on this map in desktop variables:

```
$z-index: (
  notification       : 1002, // Higher than nav bar. Legacy fix.
  poptip             : 3001, // Higher than old dialogs. Legacy fix.
  project-drawer     : 2999
);
```

The keyword `notification` for some reason throws the error `Unordered_map::at: key not found in undefined (line undefined, column undefined)`.

`node-sass` recently updated and `grunt-sass` has applied the update. The `notification` keyword does not throw an error in this newest version of `node-sass/grunt-sass` but will still throw the error in `gulp-sass` and webpack `sass-loader` until these packages update. 

Therefore, if you would like to test the lego core/desktop build you can go to [this repo](https://github.com/dtothefp/lego-libsass-gulp-grunt-webpack) and run `grunt`.

#### Update 2

Found a [branch](https://github.com/dlmanning/gulp-sass/tree/2.x) on `gulp-sass` where they are making upgrades to 3.0.0. `notification` error does not happen in this branch. I added it to the Lego package.json and added a `libsass` task to the Lego gulpfile.  Run `gulp libsass` to compile to `dist` with Libsass


danoc:
Closing because this is already complete.

PR #48, opened 2015-05-27
deleting css that no one knows why we have

kwalker3690:
@tomgenoni


tomgenoni:
@tscanlin I doubt this affects you but just in case... This rule has a relatively high specificity and is causing more problems than it solves. We add classes to anchors without hrefs if this styling is needed.

tscanlin:
LGTM, feel free to delete.

PR #54, opened 2015-06-05
adding disabled tab and fixing size of external icon

kwalker3690:
@tomgenoni


src/core/partials/objects/_tabs.scss
```diff
@@ -177,6 +177,15 @@
       background: map-fetch($color, background brand);
       bottom: -1px;
     }
+
+    &.tab-disabled {
+      @extend %link--disabled;
+      cursor: default;
+
+      &.tab-active::after { // Hides the blue bar under the active tab.
```
tomgenoni: We should probably only ever have `.tab-disabled` or `.tab-active` but not both at the same time. So we probably don't need this.

PR #55, opened 2015-06-08
Add a line-height to the help object so it works better in most applications.

danoc:
@tomgenoni – Needed this to use in a part of the product where the `line-height` was `10px`.

Merge if it looks good!

### DIFF
--- a/src/core/partials/objects/_help.scss
+++ b/src/core/partials/objects/_help.scss
@@ -14,6 +14,7 @@
   color: map-fetch($color, text white);
   width: 16px;
   height: 16px;
+  line-height: 16px;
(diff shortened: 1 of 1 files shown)


src/core/partials/objects/_help.scss
```diff
@@ -14,6 +14,7 @@
   color: map-fetch($color, text white);
   width: 16px;
   height: 16px;
+  line-height: 16px;
   text-align: center;
   display: inline-block;
```
tomgenoni: I'll merge this but we should try `display: inline-flex;` and then center the question mark with the flex options.

PR #56, opened 2015-06-15
refactor form label optional to not use pseudo element content

tscanlin:
@tomgenoni This is to address this JIRA ticket for i18n: https://optimizely.atlassian.net/browse/UXD-72

There are only about a dozen places this is used in the product and I can make a PR there as well to accompany this.

Just wanted to check and see if this is an ok way to address or if you'd prefer a different approach. This is also pretty similar to the .tagger class and I'm wondering if we would want to consolidate and make something like '.deemphasized-note' (terrible name, I know..) or something along those lines.

Let me know if you have ideas around this.

### DIFF
--- a/src/core/partials/base/_form.scss
+++ b/src/core/partials/base/_form.scss
@@ -12,7 +12,10 @@
 //     <fieldset>
 //       <ol class="lego-form-fields">
 //         <li class="lego-form-field__item">
-//           <label class="lego-label lego-label--optional">Single Input</label>
+//           <label class="lego-label">
+//             Single Input
+//             <span class="lego-label__optional">(Optional)</span>
+//           </label>
 //           <input type="text" class="lego-text-input">
 //           <div class="lego-form-note">This is a form note.</div>
 //         </li>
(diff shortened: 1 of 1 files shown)


tomgenoni:
This looks fine. Should I wait to merge until after the A/B PR is ready?

Also, regarding the `.tagger` class... it may be the same now but I'd prefer not to consolidate as the styling could eventually change and, semantically, it's nice to have different classes even though they may be applying the same styling. `text--right` and `numerical` do the same things but in tables its nice (IMHO) to have a separate, more meaningful class to use there.

tscanlin:
Yeah, lets wait until I get that PR ready to go before merging this. I'll link to it here when it's good to go. Agree with you about keeping the classes separate for semantics purposes.

tscanlin:
Opened a pr/diff for the product changes:
https://phabricator.optimizely.com/D4527

Please review when convenient @tomgenoni no rush

tomgenoni:
LGTM

4vanger:
can you merge it please so we can close https://optimizely.atlassian.net/browse/UXD-72 and https://optimizely.atlassian.net/browse/INTL-357

PR #63, opened 2015-06-25
Add travis-ci to LEGO build

danoc:
Testing to see if this works.


danoc:
@tomgenoni – This adds Travis CI integration to this repo and fixes the current linter warnings and errors.

tomgenoni:
Would you like me to test this?

danoc:
I don't think there is a way to test this since it will only be useful once
merged.

I have tested this though and you should be able to see that it says the
build is passing when you see this PR on the web.

PR #65, opened 2015-07-13
popover fix

kwalker3690:
@danoc This coincides with the work done on the guiders

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ This file follows the format suggested by [Keep a CHANGELOG](https://github.com/
 - [Patch] Fix the broken links in `CHANGELOG.md`. (#60)
 
 ### Changed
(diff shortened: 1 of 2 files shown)


tomgenoni:
@kwalker3690 Did @danoc go over the new versioning/change process?

kwalker3690:
@tomgenoni yep, just added my change to the change log

danoc:
@kwalker3690 – Can you clarify the text in the in the Markdown so that someone without context could also understand the change? Normally we would link to the issue so it provides context, but there is no issue in this case. Also, when referencing classes, can you use Markdown to format it like `.class-name`? Just clarifies things a bit.

Your change also caused a linter error. Click on "Details" on the PR page in GitHub to see the console output.

danoc:
LGTM.

PR #77, opened 2015-08-04
Tim/add move cursor trump

tscanlin:
@tomgenoni This is for the audience thing I was pinged about earlier today.


tomgenoni:
@tscanlin it looks like there are conflicts. Can you resolve?

tscanlin:
Yep, resolving and merging.

PR #80, opened 2015-08-05
Allow `.lego-overlay` to be extended with `%lego-overlay`

danoc:
@tomgenoni – Am I doing this right? The new class that extends LEGO's overlay will look like this:

```css
.info-loader-wrap {
  @extend %lego-overlay;
  @include flex-direction(column);
}
```

### DIFF
--- a/src/core/partials/objects/_spinner.scss
+++ b/src/core/partials/objects/_spinner.scss
@@ -27,8 +27,8 @@
 
 // [---]
 
-.lego-overlay,
-%lego-overlay {
+%lego-overlay,
+.lego-overlay {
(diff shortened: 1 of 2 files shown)


src/core/partials/objects/_spinner.scss
```diff
@@ -27,7 +27,8 @@
 
 // [---]
 
-.lego-overlay {
+.lego-overlay,
+%lego-overlay {
```
tomgenoni: The placeholder should be first, then the class.

tomgenoni:
LGTM

Issue #69, opened 2015-07-21
Width's should follow BEM syntax and have more consistant naming


danoc:
Width's should be renamed from `.width-1-2` to `width--1-2`, for example. Also, `.width-1` should be renamed to `.width--1-1` since `.width--100` has a pixel value while `.width--1` has a percent value.

danoc:
Implemented in this PR: https://github.com/optimizely/lego/pull/84

PR #84, opened 2015-08-10
Renaming sizing trumps to be BEM consistent #69

tomgenoni:


danoc:
LGTM but wait to merge the breaking changes.

PR #88, opened 2015-08-10
Adding overflow trumps.

tomgenoni:
Adding overflow-x: auto; and overflow-y: auto;

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This file follows the format suggested by [Keep a CHANGELOG](https://github.com/olivierlacan/keep-a-changelog).
 
 ## [Unreleased][unreleased]
+### Added
+- [Patch] Adding `overflow` trumps.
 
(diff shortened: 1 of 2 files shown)


danoc:
LGTM.

danoc:
Actually this should be a `Feature`, not a `Patch`.

tomgenoni:
Why Feature? These did not exist before.

danoc:
I did a little research today to understand the difference, and basically:

> Patch version Z (x.y.Z | x > 0) **MUST be incremented if only backwards compatible bug fixes are introduced**. A bug fix is defined as an internal change that fixes incorrect behavior.
>
> Feature version Y (x.Y.z | x > 0) **MUST be incremented if new, backwards compatible functionality is introduced to the public API**. It MUST be incremented if any public API functionality is marked as deprecated. It MAY be incremented if substantial new functionality or improvements are introduced within the private code. It MAY include patch level changes. Patch version MUST be reset to 0 when minor version is incremented.

There's a bunch that we mislabeled, but I'll update `CONTRIBUTING.md` to make this more clear.

PR #93, opened 2015-08-11
Add namespacing to LEGO

tomgenoni:


tomgenoni:
@danoc please review. Note that with three mixins — grid, media, matrix — I had to remove `lego-` from them because I can't use variables in mixin names.

danoc:
LGTM :+1:

Issue #92, opened 2015-08-11
Update the buttons to reflect the new styles


danoc:
Creating this ticket to track @kwalker3690's work on buttons.

danoc:
@kwalker3690 – How has the progress been on this?

kwalker3690:
@danoc Great! I'm working with Linda to finalize the styles. They can be seen rendered here, we've almost finalized them: http://optimizely.github.io/proto/lego-v2/

danoc:
Closing since it's merged.

Issue #85, opened 2015-08-10
Buttons should have: line-height = height


tomgenoni:
This is a better fix for cross-browser centering of text in the button.

Issue #74, opened 2015-07-21
Change 'lego-button--brand' to another name


kwalker3690:
Ideas are: 'lego-button--primary', 'lego-button--base'...

danoc:
@kwalker3690 – Can we close this?

PR #94, opened 2015-08-11
New buttons

kwalker3690:


kwalker3690:
#92 #85 #74

src/core/partials/objects/_icons.scss
```diff
@@ -76,12 +76,6 @@
     width: 32px;
     height: 32px;
   }
-
-  // This is a vertical alignment fix when used to the right of `.lego-icon`.
```
danoc: Why is this no longer needed?

danoc:
Have you clicked around the application to see how this will look?

src/core/partials/objects/_icons.scss
kwalker3690: I think it may have to do with the line-height change, I'm not 100% sure, but I checked with @tomgenoni and he thought it was fine to delete.

kwalker3690:
I haven't yet, because the names of the button classes changed significantly, so it will take more work to update the classes to see it in the app. But I did make a [test sheet] (http://optimizely.github.io/proto/lego-v2/) to see how it would look, and checked that across all browsers (including ie)

danoc:
Awesome. Sweet job with the examples.

Please update the `CHANGELOG.md` with the change. Feel free to make it long so that it includes instructions on how to upgrade the buttons (e.g. class names that are now different)

kwalker3690:
@danoc I didn't update any of the docs comments in the files, should I do that in this PR or are the docs shifting too much to do so?

PR #96, opened 2015-08-17
Add explanations to Patch, Feature, and Release in `changelog.md`.

danoc:
@kwalker3690 – Here are the changes you requested.


kwalker3690:
We talked about this a while ago in person, but can we change the description of release? It's not clear to me what incompatible API changes means in this case

danoc:
@kwalker3690 – What do you think?

kwalker3690:
Love it, looks perfect now, thanks!

PR #98, opened 2015-08-17
Removed old JS targets from gitignore

chestone:
To prevent, namely myself from having an old version of the repo and accidentally committing js code.


.gitignore
```diff
@@ -1,7 +1,4 @@
 node_modules
-dist
```
danoc: Why remove the `dist` folder? I think `gulp sass` writes to it.

chestone: Woops! Didn't realize that was the current target. Fixed in rev.

danoc:
LGTM!

PR #108, opened 2015-09-01
Flexbox changes, updating contrib.

tomgenoni:


tomgenoni:
@danoc do you think the mixins should be its own file, maybe even trumps? Or should we keep as is? I think it's probably okay as is.

src/core/partials/components/_layout.scss
```diff
@@ -1,107 +0,0 @@
...
-
-  &--scroll-x {
-    overflow-x: auto;
-  }
-
-  &--scroll-y {
-    overflow-y: auto;
-  }
-
-  &--empty {
```
danoc: I think this one is helpful and would love if we kept it in some form.

tomgenoni: How about we create a partials just for centering classes/techniques? There are at least three I can think of, each with their own pros/cons.

danoc: Yeah, I think that's good. :+1:

tomgenoni:
@danoc @kwalker3690 any corrections/concerns?

danoc:
LGTM as long as we explore a partial for centering classes.

tomgenoni:
Yup. https://github.com/optimizely/lego/issues/116

src/core/partials/components/_layout.scss
danoc: Turns out this is only used twice in the Optimizely app.

PR #118, opened 2015-09-04
Select height fix

tomgenoni:
screen shot 2015-09-04 at 2 48 10 pm
`.lego-select` was too short. This makes it behave like new buttons by calculating the same way buttons do. Any changes to `.lego-button` height will now apply correctly to `.lego-select`.


danoc:
LGTM

PR #120, opened 2015-09-09
Fixing add/remove table row.

tomgenoni:
@danoc This is a more robust, simpler solution to this problem.

### DIFF
--- a/src/core/partials/base/_tables.scss
+++ b/src/core/partials/base/_tables.scss
@@ -256,6 +256,7 @@
   }
 
   // Add/Remove Rows
+  // In button rows use a placeholder button with `visibility--hidden` in the `--controls` cell to maintain spacing.
 
   &--add-row {
 
@@ -268,6 +269,7 @@
       padding-bottom: spacer(1);
     }
 
+    //
     &__controls {
       width: 1%;
       white-space: nowrap;
--- a/src/core/partials/base/_tables.scss
+++ b/src/core/partials/base/_tables.scss
@@ -269,21 +269,8 @@
     }
 
     &__controls {
-      vertical-align: middle !important;
-
-      .flex {
-        width: 60px;
-
-        > * {
-          @include flex(none);
-        }
-
-        // TODO (kelly): remove link--bad-news once all forms in the product have been converted to use the new button
-        .link--bad-news,
-        .#{$namespace}button--remove {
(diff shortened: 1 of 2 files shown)


src/core/partials/base/_tables.scss
```diff
@@ -268,22 +269,10 @@
...
-          @include flex(none);
-        }
-
-        // TODO (kelly): remove link--bad-news once all forms in the product have been converted to use the new button
-        .link--bad-news,
-        .#{$namespace}button--remove {
-          margin-left: auto;
-        }
-      }
+      width: 1%;
```
danoc: What is this line doing?

src/core/partials/base/_tables.scss
```diff
@@ -256,6 +256,7 @@
   }
 
   // Add/Remove Rows
+  // In button rows use a placeholder button with `visibility--hidden` in the `--controls` cell to maintain spacing.
```
danoc: Where is `visibility--hidden` being used?

src/core/partials/base/_tables.scss
tomgenoni: The width constrains the table cell to the narrowest possible width. The `nowrap` prevents the buttons from wrapping.

danoc:
LGTM.

src/core/partials/base/_tables.scss
tomgenoni: On any placeholder buttons.
screen shot 2015-09-09 at 11 34 35 am

src/core/partials/base/_tables.scss
danoc: I don't think the example HTML on the design dot site is following the CSS here:

http://design.optimizely.com/oui/core/forms.html#add-remove-row

It doesn't seem to use `.controls` and `.lego-table-add-row` isn't doing anything, I think. Maybe because it's an older version of LEGO?

danoc: My comment was totally wrong. `.lego-table--add-row` is working, but it looks like `.lego-table--add-row__controls` isn't used.

danoc: Also, the buttons on the documentation should have `type="button"` on them since button's in a `form` default to `type="submit"`. This confused me for a sec when I was implementing one of them.

PR #121, opened 2015-09-09
Removing outline:none from buttons.

tomgenoni:
@danoc Review


danoc:
Niceee! Glad to see this. Thanks!

PR #124, opened 2015-09-10
Adding consistent disabled states for inputs.

tomgenoni:
See https://jsfiddle.net/rxu4empz/1/ for examples.


tomgenoni:
@danoc Note that I removed the class referenced in https://github.com/optimizely/lego/issues/117

danoc:
Are you saying you removed the `.lego-input--disabled` class?

tomgenoni:
yes I removed `.lego-input--disabled`.

danoc:
Cool let's make this a release then since it's a breaking change.

danoc:
Actually, would it just work the same with the new Sass? As in everything that had `.lego-input--disabled` will still keep the same styles?

tomgenoni:
It might but I want to keep the naming consistent.

danoc:
Ok cool LGTM.

PR #128, opened 2015-09-15
Added back a reference to the `_borders.scss` file.

danoc:
Relates to this issue: https://github.com/optimizely/lego/issues/127

Going to immediately release this as v4.0.1 to fix broken borders in the app.


tomgenoni:
LGTM

danoc:
@tomgenoni pointed out that this doesn't fix the borders in apps that use LEGO since trumps are opt-in. Merging it but not going to immediately release a version since it's not critical.

PR #133, opened 2015-09-17
Deprecated flex

tomgenoni:
@danoc Please review. These add back the older flex classes. Note that I removed the old mixins as they were dupes of existing ones, except for the `min-height` change but that shouldn't be a problem.

### DIFF
--- a/src/core/partials/elements/_flexbox--deprecated.scss
+++ b/src/core/partials/elements/_flexbox--deprecated.scss
@@ -1,0 +1,76 @@
+// # Flexbox
+// Flexbox support is full on Chrome, Firefox and Safari (with prefix) and nearly full for IE 10.
+//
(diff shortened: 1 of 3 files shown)


src/core/partials/elements/_flexbox--deprecated.scss
```diff
@@ -0,0 +1,76 @@
...
+
+
+// ## Flex classes
+
+.flex {
+  @include display(flex);
+}
+
+.flex-1 {
+  @include flex(1);
```
danoc: This produces slightly different behavior than the old functionality because `min-height: 0` and `min-width: 0` will now be added. I don't think it would break anything though so feel free to leave as-is.

danoc: Actually your description already mentioned that. :smile:

danoc:
LGTM.

PR #145, opened 2015-10-23
Component javascript

tomgenoni:
@tscanlin @kwalker3690 @danoc 

This is NOT ready to merge, I'll need to update Contributing.md but before I did that I want you all to review this stuff. Please pull this branch, npm install, and run `gulp html-tests`. What I was trying to do:

- rewrite the JS so that it can be used/modified by any project
- use jQuery so it can be edited by a broader audience
- contain it so there won't be any conflicts with existing frameworks, jQuery.noConflict()
- fix the long-standing differences between popovers/poptips 

Main things I'd like reviewed:

- overall structure: could these be packaged better? do they need to be?
- syntax: I tried to rely on data-attrs as much as possible to keep the styling/js separation. It makes for more HTML but it seems okay
- can the 'editable' component be integrated with Vue so that it can pass the value into the app?
- what documentation system should we use here?

Please pick apart any of the code. Let me know if you'd like any explanation for how things work or why I did what I did. Note that I changed some Sass as well so this will involve breaking changes.


gulpfile.js
```diff
@@ -88,11 +91,28 @@ gulp.task('sass', function() {
     .pipe(browserSync.stream());
 });
 
+// Concatenate and uglify js
+gulp.task('js', ['js:dependencies'], function() {
+  gulp.src('src/js/**/*.js')
+    .pipe(uglify('core.min.js').on('error', gutil.log))
```
tscanlin: does uglify concatenate on its own?

src/core/partials/objects/_accordion.scss
```diff
@@ -8,43 +8,13 @@
 // - The parent div must be `display:flex`;
 // - The parent div must have height, usually accomplished with a `flex: 1`;
```
tscanlin: With all these examples getting moved out of these files? Where will they end up coming from in the future?

gulpfile.js
tomgenoni: Yup.

src/core/partials/objects/_accordion.scss
```diff
@@ -8,43 +8,13 @@
...
 
-// The wrapping `<div class="doc-accordion-example">` here has been added for purposes of the documentation and should be removed.
```
kwalker3690: I personally still find it really useful to have these examples at the tops of files, is your thought instead to have people look at the html examples?

src/core/partials/objects/_accordion.scss
tomgenoni: They would live in design-dot or in @danoc future sass doc engine. Either way we'd definitely have documentation.

src/core/partials/objects/_accordion.scss
tomgenoni: We can leave them here. Just answered similar question from Tim. I don't have a preference. We can pick whatever works best. I pulled them out for now because they HTML was changing.

src/core/partials/objects/_accordion.scss
kwalker3690: Maybe let's just leave the examples in here then until we have other solutions, I use these examples all the time

src/js/components/accordion.js
```diff
@@ -0,0 +1,35 @@
...
+
+    // If clicking the non-active [data-oui-accordion-item].
+    if ( !$target.hasClass(ACTIVE_CLASS) ) {
+
+      // Get parent accordion.
+      var $accordion = $(this).closest("[data-oui-accordion]");
+      var $curActive = $accordion.find("." + ACTIVE_CLASS);
+      var paneHeight = $curActive.outerHeight(); // Get height of opened pane.
+
+      $target.animate({
```
tscanlin: is there a reason to use jquery.animate rather than css transitions? I'd guess the latter would be slightly more performant.

src/js/components/disclose.js
```diff
@@ -0,0 +1,14 @@
+(function($) {
+
+  $(document).on( "click", "[data-oui-disclose-link]", function(e) {
```
tscanlin: probably not anything we need to worry about now, but would we want to make 'data-oui-' a variable?

src/js/components/accordion.js
tomgenoni: I couldn't find a way to make that work since you can't CSS animate to 'auto' or to an inline value. You can animate max-height but you have to pick an arbitrarily high number which may not be high enough but can also cause delays when closing. But if you know of a way let me know.

tests/editable-v4.html
```diff
@@ -0,0 +1,195 @@
...
+    </symbol>
+    <symbol id="toolbar-size-widescreen" viewBox="0 0 24 24">
+      <g><path class="st0" d="M20.8 19H4.2c-.9 0-2.2-.7-2.2-1.8V6c0-1 1.3-2 2.2-2h16.6c1 0 2.2 1 2.2 2v11.2c0 1.1-1.3 1.8-2.2 1.8zM4.2 5C3.8 5 3 5.6 3 6v11.2c0 .4.7.8 1.2.8h16.6c.5 0 1.2-.4 1.2-.8V6c0-.4-.7-1-1.2-1H4.2z"></path><path class="st0" d="M8 20h9v1H8z"></path><path class="st0" d="M12 19h1v1h-1z"></path><circle class="st0" cx="12.6" cy="16.4" r=".9"></circle><path class="st0" d="M3 14h19v1H3z"></path></g>
+    </symbol>
+    <symbol id="trash" viewBox="0 0 96 96">
+      <path d="M84 22H68v-4c0-6.63-5.37-12-12-12H40c-6.63 0-12 5.37-12 12v4H12c-2.21 0-4 1.79-4 4s1.79 4 4 4h4v48c0 6.63 5.37 12 12 12h40c6.63 0 12-5.37 12-12V30h4c2.21 0 4-1.79 4-4s-1.79-4-4-4zm-48-4c0-2.21 1.79-4 4-4h16c2.21 0 4 1.79 4 4v4H36v-4zm36 60c0 2.21-1.79 4-4 4H28c-2.21 0-4-1.79-4-4V30h48v48z"></path>
+    </symbol>
+    <symbol id="upload" viewBox="0 0 96 96">
```
tscanlin: I wish we could avoid repeating this stuff so much, but +1 for all the iterations. I like options :)

tests/editable-v1.html
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+
+  </style>
+
+</head>
+<body>
+
+
+<h2><span data-pop-type="over" data-location="1" class="edit" data-pop-html="edit--input">Bring thousands of optimization ideas to life in minutes</span></h2>
```
tscanlin: What does 'data-pop-html' do? Are you sanitizing this input if it gets inserted with .html() or .innerHTML?

tomgenoni: My bad, I should have removed v1-v3. Version 4, the one linked to in the index.html is the only one I'm proposing.

src/js/common/pop.js
```diff
@@ -0,0 +1,105 @@
...
+
+  // Show at bottom if not enough room.
+  if ( trigger.top - pop.outerHeight < scrolled ) {
+    location = 7 // Position 8
+  }
+
+  // This handles the positioning/showing of the different types of pops: tips, overs.
+  // Locations correspond to position that the 'pop' element should show up relative to the trigger.
+
+  //    1    2    3
```
kwalker3690: Maybe just start with 0 instead of 1 so that there's no need to do subtracting of 1, ie no need to figure out that location = 1 really signifies Position 2?

Also, this is an awesome little graphic!

tomgenoni: Yeah, good point. I was trying to be user friendly but it's more confusing than it's worth.

tscanlin:
Love the overall structure, it makes sense to me. Also, +1 for data attributes, they make things way more flexible.

The editable component can be integrated with Vue fairly easily as a directive. The one thing you might want to think about here is the full life cycle of these components. For instance, I notice that there's a lot of $el.on('...', doStuff..), but those event handlers aren't ever removed. This is fine in a multipage app because things get re-rendered. But for single page apps we need to be more careful about cleaning up after ourselves and removing these event handlers (eg: $el.off()). Vue has a directive API that is basically an object with the methods: 'bind', 'unbind', 'update' (update is when the directive gets a new value set on it from outside). The more closely we can sync up the API between these components and our Vue directives the easier it will be to make use of these (at least bind/unbind would be helpful).

I'm open to different ways to document this stuff though. Maybe have a page with iframes that load all of the /test/*.html files? Not super elegant but it would probably work ok and be low maintenance.

src/js/components/disclose.js
```diff
@@ -0,0 +1,14 @@
+(function($) {
```
kwalker3690: This may be a little sugar to add later, not needed now, but it would be nice if the content slid down from the top instead of just show/hide

tomgenoni: Agreed.

src/js/components/disclose.js
tomgenoni: Yeah, a related question I had was how to declare variables that are NOT in the global scope.

src/js/components/editable.js
```diff
@@ -0,0 +1,77 @@
+(function($) {
+
+  $(document).on( "click", "[data-oui-editable]", function(e) {
+
+    var ACTIVE_CLASS = "edit-text-trigger";
+
+    $("[data-oui-editable]").removeClass(ACTIVE_CLASS).show();
+
+    // Remove any existing editing box.
+    $(".edit-text").remove();
```
kwalker3690: make this a variable like you did ACTIVE_CLASS cause you use it more than once

src/js/components/poptip.js
```diff
@@ -0,0 +1,29 @@
...
+  $(document).on({
+    mouseenter: function () {
+
+      var ARROW_SIZE = 7;
+
+      // Get the properties of the trigger.
+      var trigger = ouiGetProps( $(this) );
+
+      // Insert the poptip.
+      $("body").append("<div class='pop pop--tip'>");
```
kwalker3690: Do you not need to close the div here?

src/js/components/disclose.js
tscanlin: yeah, you probably either have to have a global or switch to using require.. :/

src/js/components/poptip.js
tscanlin: jQuery does it for you automagically :)

src/js/components/dropdown.js
```diff
@@ -0,0 +1,48 @@
+(function($) {
```
kwalker3690: Same thing here, a slide down or something would be cool, similar to how you've done for pops

src/js/components/poptip.js
kwalker3690: Oh cool good to know :)

tomgenoni:
@tscanlin Yes, we should talk more about unbinding and those other implementation details. I defer to you on how best to manage that.

src/core/_core-variables.scss
```diff
@@ -301,12 +301,12 @@ $accordion-link-height       : 26px;
...
 
 $popover: (
-  max-width                  : 250px
+  max-width                  : 350px
 );
 
 // ## Poptip
 
-$pop-tip: (
+$poptip: (
```
danoc: Could this break OUI instances that read that `pop-tip` `max-wdith`?

tomgenoni: Not sure the question... I'm proposing changing the variable name so, yes, it would break if was set in other places.

src/core/partials/objects/_edit-text.scss
```diff
@@ -0,0 +1,57 @@
+.edit-text {
+  display: none;
+
+  &__wrap {
+    position: relative;
+  }
+
+  &__input {
+    position: absolute;
+    top: -8px; // Moving border back for spacing around text.
```
danoc: Can the 16px be a variable and this be `-($var / 2)`. Or are they not related?

src/js/common/pop.js
danoc: :+1:

src/js/components/disclose.js
danoc: Are you sure the pattern you are using is definitely creating globals?

```js
(function(){
  var foo = 'Hello';
  var bar = 'World!'
  
  function baz(){
      return foo + ' ' + bar;
  }
})();

 //These all throw exceptions:
console.log(foo);
console.log(bar);
console.log(baz());
```

Example from: http://markdalgleish.com/2011/03/self-executing-anonymous-functions/

danoc:
Looking pretty good. Left some comments. One nit is that I'm not a huge fan of putting the examples in `/tests` because I'd expect those to be some sort of automated unit tests. Maybe `/examples` is better?

src/core/partials/objects/_edit-text.scss
tomgenoni: Yeah, I could make it a var.

tomgenoni:
`examples` is fine with me.

src/js/components/disclose.js
tomgenoni: I don't think I wasn't worried about that for the variables. I didn't want the variables, like ACTIVE_CLASS, to be in the same scope as the other components.

tomgenoni:
@tscanlin @kwalker3690 @danoc I pushed my changes that addressed the PR. @kwalker3690 Proper HTML documentation will come in another pass. We should discuss this at OBA.

CHANGELOG.md
```diff
@@ -5,6 +5,19 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This file follows the format suggested by [Keep a CHANGELOG](https://github.com/olivierlacan/keep-a-changelog).
 
 ## [Unreleased][Unreleased]
+### Added
+- [Release] Adding local JS to power components.
+- [Patch] Adding additional HTML examples to test component javascript.
```
danoc: Not a huge deal, but this would technically be a feature since it is a backwards compatible new "feature" whereas "patch" is for backwards compatible bug fixes. We haven't been great at following that though since I didn't understand the difference for a while.

tomgenoni: fixed

tomgenoni:
@tscanlin @danoc @kwalker3690 I believe I addressed all the comments. Can you give it one last look?

tscanlin:
LGTM

As far as using the components here in the product I would probably defer to Jordan for that. I'm just not sure if it makes sense given that the functionality would basically be the same. But if some of these components are substantially different I can look into migrating them, just let me know.

tomgenoni:
Yes, agreed. For now just getting into OUI so that it's available for use and will work on its own is phase 1.

tscanlin:
sounds good to me

danoc:
LGTM!

PR #146, opened 2015-10-27
Table scroll

tomgenoni:
@tscanlin @kwalker3690 @danoc 

Opening PR for discussion, don't merge quite yet. :)

This version of the table scrollable is the cleanest I've come up with. But the downsides are:

- It requires JS to test for "always on" scroll bars in OS X. This could be done once and stored in cookie...? This appears to be unavoidable, regardless of implementation.
- Changes the default behavior of the table completely, converting to flexbox. No specific concern here other than it could lead to unknown bugs or unexpected behavior.
- Requires a width on all `th` and `td` elements. 

This is the cleanest implementation I could come up with but it makes me mildly nervous. Do we want to proceed?

### DIFF
--- a/src/core/partials/base/_tables.scss
+++ b/src/core/partials/base/_tables.scss
@@ -284,7 +284,7 @@
 }
 
 // Scrolling Tables
-// Uses two tables, one for the header and one for the body.
+// Uses flexbox to change the default behavior of the table elements.
 
 // - The parents of `table-scroll` MUST be `display:flex`.
 // - All `<th>` and `<td>` elements must have matching widths.
@@ -292,14 +292,14 @@
 // <table class="table table--scroll table--rule">
 //   <thead>
 //     <tr>
-//       <th class="width--1-2">one</th>
+//       <th class="width--2-4">one</th>
 //       <th class="width--1-4">two</th>
 //       <th class="width--1-4">three</th>
 //     </tr>
 //   </thead>
 //   <tbody>
 //     <tr>
-//       <td class="width--1-2">one</td>
+//       <td class="width--2-4">one</td>
 //       <td class="width--1-4">two</td>
 //       <td class="width--1-4">three</td>
 //     </tr>
--- a/src/core/partials/base/_tables.scss
+++ b/src/core/partials/base/_tables.scss
@@ -287,7 +287,7 @@
 // Uses two tables, one for the header and one for the body.
 
 // - The parents of `table-scroll` MUST be `display:flex`.
-// - Both `<th>` and `<td>` elements must have matching widths.
+// - All `<th>` and `<td>` elements must have matching widths.
 
 // <table class="table table--scroll table--rule">
 //   <thead>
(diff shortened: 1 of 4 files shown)


src/core/partials/base/_tables.scss
```diff
@@ -282,3 +282,59 @@
 .#{$namespace}table-row--active {
   background-color: map-fetch($color, background faint);
 }
+
+// Scrolling Tables
+// Uses two tables, one for the header and one for the body.
```
tscanlin: does this actually use 2 tables?

tests/table-scroll.html
```diff
@@ -0,0 +1,142 @@
...
+
+      function setOverflowScroll() {
+
+        $(".table--scroll").each(function(){
+
+          var tableWidth =  $(this).width();
+          var rowWidth = $(this).find("tbody tr:first-child").width();
+          var diff = tableWidth - rowWidth;
+
+          console.log(diff);
```
tscanlin: nit: probably want to remove this before merging.

src/core/partials/base/_tables.scss
```diff
@@ -282,3 +282,59 @@
...
+
+  tr {
+    @include display(flex); // All table rows changed to flexbox.
+    @include flex(none);
+    width: 100%;
+  }
+
+  th,
+  td {
+    display: block;
```
tscanlin: any reason you don't wanna flex these too? :)

tscanlin:
minor comments but this looks good to me! I wonder if it makes more sense for us to use percentages or pixels for fixed column widths, but I guess it depends on the situation and both would be possible.

Nice work :)

tomgenoni:
Yeah, you can use either fixed or percentages.

src/core/partials/base/_tables.scss
danoc: Yeah I feel like that sentence and the example below are contradicting.

danoc:
LGTM once you address Tim's comments.

kwalker3690:
LGTM

tests/table-scroll.html
tomgenoni: Removed

src/core/partials/base/_tables.scss
tomgenoni: Originally that's what I had (`flex:1`) but that required using trumps for `min-widths` (`width` is ignored when a `flex` value is present). That would mean we would have to create a complete set of `min-width` classes and explain when to use which. So I figured just requiring a `width` trump on all `th/td` was easier to understand/implement. But I'm open to another approach if we feel it's important.

tscanlin: Makes sense to me!

tscanlin:
LGTM

src/core/partials/base/_tables.scss
tomgenoni: Yes, that should have been updated. Fixed now.

PR #147, opened 2015-10-29
Lock down version of node-sass

tomgenoni:
Here are the steps I took. This is a one-time thing, developers cloning the repo do not need to do anything special.

**package.json**

- `gulp-sass` is the only entry in “dependencies”
- all others gulp packages are listed in “devDependencies”

**at oui root** 

- remove `node_modules` or from a fresh clone run `npm install` 
- edit into `node_modules/gulp-sass/package.json` by changing `"node-sass": "^3.4.1"` to version `"3.3.3"`
- in `node_modules/gulp-sass/` directory run `npm install --production` to reinstall `gulp-sass` "dependencies" (and not the "devDependencies") to install the earlier version of `node-sass`
- cd back to `oui` root and run `npm shrinkwrap`
- this creates `npm-shrinkwrap.json` file at the root. It is generated by reading through all "dependencies" and, in this case, writing out only the `gulp-sass` dependencies. 
- commited `npm-shrinkwrap.json` into git 

To test, remove repo’s `node_modules` and run `npm install`. You can confirm it worked if `node_modules/gulp-sass/node-sass/package.json` installed version 3.3.3.

Future package installation should be done with `npm install [package-name] --dev` so that entries for them will show up in `"devDependencies"` and will not trigger a mismatch between `package.json` and `npm-shrinkwrap.json`. 

https://docs.npmjs.com/cli/shrinkwrap

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This file follows the format suggested by [Keep a CHANGELOG](https://github.com/
 ### Added
 - [Feature] Add a license to the repository. (#125)
 - [Feature] Adding scrolling variation for tables.
+- [Feature] Adding `shrinkwrap` file to lock down `node-sass` version used by `gulp-sass`.
 
 ### Fixed
 - [Patch] Fix incorrect links in the changelog.
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -9,7 +9,6 @@ var bump        = require('gulp-bump'),
     scsslint    = require('gulp-scss-lint'),
     symlink     = require('gulp-symlink'),
     sass        = require('gulp-sass'),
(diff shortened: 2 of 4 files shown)


tomgenoni:
@danoc @kwalker3690 @tscanlin In order to get around a libsass bug that caused some problems with `@extend` I'm using `npm shrinkwrap` to lock down a `gulp-sass` dependency. We may not need this for too long if the bug is fixed but for now this seems like the best/easiest solution.

danoc:
@tomgenoni – So this should work if we kill our `node_modules/` then `npm i`?

tomgenoni:
Yup. It'll install the versions it finds in the npm_shrinkwrap.json. In this case only gulp-sass has been locked.

danoc:
I don't think we really need this anymore?

PR #157, opened 2015-12-02
add current color background trump

tscanlin:
@danoc @tomgenoni @kwalker3690 

I would like to add this trump to oui so that elements can inherit a background color from a parent element. What do you all think?


tomgenoni:
LGTM.

danoc:
Please add it to the Changelog! See `CHANGELOG.md`. 

LGTM otherwise.

PR #160, opened 2015-12-03
Method for adding any svg icons into inputs.

tomgenoni:
screen shot 2015-12-03 at 2 15 47 pm

This adds an input variation to show a calendar and dropdown arrow.


tomgenoni:
@danoc @tscanlin Currently this variation contains both the dropdown and the date icon. Would we ever need just the date?

tscanlin:
I can't think of a time we'd just want the date. Seems like it would be good to be consistent with that.

Does this use a jquery datepicker under the hood or a native html5 date picker? Or is this just for styling purposes?

Also, what's the reason for the 2 bg images?

If you did want the icon to be optional this could be implemented similar to the date picker in the scheduler (with svg icons the normal way, not as a bg).

tomgenoni:
This is purely styling. I'm using two base64 background images to limit the HTML needed to just throwing a class on there. But, yes, I could use svgs normally with extra HTML. So it would look something like:

```
<div class="datepick">
   <input class="input-datepick" />
  <svg... />
  <svg.../>
</div>
```

where `datepick` is `display:inline-block` and `position: relative` and the svgs are positioned absolute. @danoc you have an opinion? It's probably "better" to do it this way but the overhead seemed a bit much.

tscanlin:
I probably slightly prefer the second way since it gives a bit more control and it's easier to update the icons, but it's not a huge deal either way.

LGTM. Thanks Tom!

tomgenoni:
Yeah, the more I think about it the more I think the verbose HTML way is more expected/consistent. I will redo and update.

tomgenoni:
@danoc @tscanlin I've updated the CSS/HTML so that we can now add any icon to the left and/or right of an input. This HTML will output the image attached to the first comment:

```
<div class="lego-input-icon">
  <input type="text" class="lego-text-input soft-triple--sides width--150">
  <svg class="lego-icon lego-input-icon__left lego-input-icon__muted">
    <use xlink:href="#calendar-16"></use>
  </svg>
  <svg class="lego-icon lego-input-icon__right">
    <use xlink:href="#arrow-down-16"></use>
  </svg>
</div>
```

Viewable at http://optimizely.github.io/proto/schedule/

If we go with this approach I will refactor our search inputs as well.

tscanlin:
Thanks so much for making this improvement, I love that it's so much more general and flexible, it should make our lives much easier. Thanks Tom!

Even with the icon being a bit weird I'd say LGTM =)

src/core/partials/objects/_input-icon.scss
```diff
@@ -0,0 +1,37 @@
...
+//   <input type="text" class="lego-text-input soft-triple--sides width--150">
+//   <svg class="lego-icon lego-input-icon__left lego-input-icon__muted">
+//     <use xlink:href="#calendar-16"></use>
+//   </svg>
+//   <svg class="lego-icon lego-input-icon__right">
+//     <use xlink:href="#arrow-down-16"></use>
+//   </svg>
+// </div>
+
+.#{$namespace}input-icon {
```
tscanlin: It feels way nicer to not have that long, difficult to maintain base64 string. +100!

PR #164, opened 2015-12-08
Updated select

tomgenoni:
This simplifies the select code quite a bit. The original code was complicated because I was working around some strange SVG behavior in IE that turned out to be an svg bug. Using non-expanded SVG as source for base64 background image fixed in in IE.

@danoc @tscanlin


tscanlin:
+1 for simplicity. Nice work! LGTM

danoc:
Awesome! Can you close this one and merge it into `devel` please? There's no way to switch for old branches, unfortunately.

PR #165, opened 2015-12-08
Add a deploy script to NPM and Gulpfile

danoc:
Still need to update the `CHANGELOG`, but just putting this out there.


tomgenoni:
I'd like to clean up the gulp tasks and some variables but that can wait until after this is in. LGTM.

PR #171, opened 2015-12-09
Updated select

tomgenoni:
This is a hack-free simpler implementation of the select dropdown for IE. There is a odd bug in IE that it turns out I can sidestep.


tomgenoni:
@danoc @tscanlin You both approved this previously but please do again. Opening this against devel this time.

tscanlin:
LGTM

danoc:
LGTM!

PR #172, opened 2015-12-09
Icon align

tomgenoni:
Adding `.icon--text-align` class to allow inline icons to visually align with the text. E.g., help icons.

Before:
screen shot 2015-12-09 at 3 08 57 pm

After:
screen shot 2015-12-09 at 3 09 11 pm

@danoc @kwalker3690


danoc:
LGTM 

(Although I'd think we should eventually discuss using more `em` or `rem` for accessibility reasons.)

PR #173, opened 2015-12-10
Js updates

tomgenoni:
This just cleans up the JS a bit more, most importantly it adds `document` click handlers only when they are needed, and removes them when they are not.

@danoc @kwalker3690


danoc:
LGTM

tomgenoni:
Opening different PR for these changes. Closing.

PR #187, opened 2016-01-11
oui-extras.css file output only icons.

tomgenoni:
@danoc Can you confirm that the build system still runs? Feel free to pull this branch and make changes.

PR #189, opened 2016-01-12
Rename `core` to `oui`

tomgenoni:
@danoc This changes all references of `core` to `oui`.


danoc:
Going to wait until we merge #187 to do a pass on this.

danoc:
@tomgenoni – I made some changes. Take a look.

Although I don't think renaming `src/core/` to `src/oui/` (instead of `src/scss/`) makes it more clear for users importing OUI since the imports already point into `node_modules/oui/`.

tomgenoni:
It makes it clearer for users integrating it with gulp/grunt because now their root.scss file will look like:

```
@import 'oui/oui-variables';
@import 'design-dot/design-dot-variables';

@import 'oui/oui-partials';
@import 'design-dot/design-dot-partials';
```

The partial paths will now all start with their origins.

danoc:
Hmmm... Those imports would work because of the `importPath` is set to point to `node_modules/`?

tomgenoni:
Yes, but it's not hardcoded. OUI exports the path to itself for other projects to use. `includePath` points to its `src`.

PR #205, opened 2016-01-14
Remove comment about space-around that isn't needed

### LINKED ISSUE #180: `justify-content: space-around` note is the wrong place
https://github.com/optimizely/oui/blob/1fe580e06cf56ea1ce40fbd0fde6d3d8e14d32da/src/core/partials/elements/_flexbox.scss#L132-L133

@tomgenoni Think we should add the class or delete this?

danoc:
Fixes #180.

cc: @tomgenoni


tomgenoni:
This comment is in the wrong place but it's a valid warning. Should it be moved to the `justify-content` mixin?

danoc:
Yeah, good call.

Issue #202, opened 2016-01-13
Break up gulpfile.js into individual files


tomgenoni:
Our gulpfile.js is starting to get a little unruly. It'd be nice to divide up the tasks into separate files.

danoc:
Yeah, I agree. There's probably a bunch of other stuff in there that isn't actually needed.

PR #206, opened 2016-01-14
Gulp tasks

tomgenoni:
@danoc I was able to check that all the tasks still work as before except for these two:

```
deploy.js
versioning.js
```

Please pull the repo and check as best you can. Is there a way to do dry runs with these?

Also, I'll do another pass at adding variables (like for deploy) but in a later pass.


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ This file follows the format suggested by [Keep a CHANGELOG](https://github.com/
 
 ## [Unreleased][Unreleased]
 ### Added
+- [Feature] Moving gulp tasks into individual files. Adding documenation for each one. (#177)
```
danoc: This should say #202.

tomgenoni: Good catch. I'll update.

danoc:
LGTM!

Issue #156, opened 2015-12-01
Change `_borders.scss` to use a border color that can be overridden without changing the base color 


danoc:
`_borders.scss` uses the base color on all of the trumps. I'd like to use a lighter color for ScribeSass border trumps without having to modify `$color, ui base`. It would be great if border color was a variable in `_core-variables.scss` that could be overwritten by applications using OUI.

cc: @tomgenoni

PR #219, opened 2016-01-26
Creating border object to hold variables.

tomgenoni:
@danoc This addresses #156. I've created a variable and applied it to all border colors. This isn't specific to `_borders.scss` but you can now amend the `$border` object with new colors and adjust as needed in your application.


danoc:
LGTM!

Issue #220, opened 2016-01-27
Select left-padding is inconsistent with inputs


danoc:
See screenshot:
<img width="711" alt="screenshot 2016-01-26 16 36 30" src="https://cloud.githubusercontent.com/assets/916038/12600087/b9396a88-c44b-11e5-82b7-51c35fed40bc.png">

tomgenoni:
![screen shot 2016-01-26 at 4 54 16 pm](https://cloud.githubusercontent.com/assets/1171072/12600403/5acb080a-c44e-11e5-94e1-47362f4b9b0f.png)

tomgenoni:
In PR #221

PR #221, opened 2016-01-27
Decreasing padding around selects.

tomgenoni:
@danoc Addresses #220

### DIFF
--- a/src/oui/partials/base/_select.scss
+++ b/src/oui/partials/base/_select.scss
@@ -11,12 +11,12 @@
   height: map-fetch($button, size base height);
   border-radius: map-fetch($border-radius, base);
   font-size: map-fetch($font, size base);
-  padding: 0 spacer(3) 0 7px;
+  padding: 0 spacer(3) 0 map-fetch($form, input padding);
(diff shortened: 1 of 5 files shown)


src/oui/partials/base/_select.scss
```diff
@@ -11,12 +11,12 @@
   height: map-fetch($button, size base height);
   border-radius: map-fetch($border-radius, base);
   font-size: map-fetch($font, size base);
-  padding: 0 spacer(4) 0 spacer(1.5);
+  padding: 0 spacer(3) 0 7px;
```
danoc: Might make sense to make these variables?

https://github.com/optimizely/oui/blob/9d063587dfcfef0fe6bf806e823d1b3d5fa94799/src/oui/partials/base/_form.scss#L285

tomgenoni: Yeah, I should since it's shared by all the inputs.

tomgenoni:
@danoc adding padding variable

danoc:
LGTM, thanks!

PR #227, opened 2016-02-01
Update the oui-icons reference in package.json

danoc:
This doesn't retroactively fix old versions, but fixes it for the future release. 

cc: @tomgenoni


tomgenoni:
LGTM

PR #261, opened 2016-02-19
Add other input types to input reset.

### LINKED ISSUE #217: Remove outline from inputs of type number
screenshot 2016-01-25 13 34 15

It would be great if `input`'s with `type="number"` were right-aligned automatically and the default padding was changed to better suit the input.

danoc:
Fixes #217.

cc: @tomgenoni


tomgenoni:
LGTM

PR #266, opened 2016-02-19
Rename an incorrect width class referenced within a comment.

### LINKED ISSUE #256: Table example code uses `width-1` instead of `width--1-1`
Search for:

```
<table class="lego-table lego-table--add-row width-1">
```

danoc:
Fixes #256.

cc: @tomgenoni


tomgenoni:
LGTM

PR #274, opened 2016-02-19
Add step at end of CONTRIBUTING to include merging `master` back into `devel`

### LINKED ISSUE #273: Add step at end of CONTRIBUTING to include merging `master` back into `devel`.
This is more important when deploying with the "hotfix" approach because it causes `master` and `devel` to diverge.

danoc:
Fixes #273.

cc: @tomgenoni


tomgenoni:
LGTM

PR #276, opened 2016-02-19
Autoprefix

### LINKED ISSUE #241: Display mixin is obsolete
Display mixin: https://github.com/optimizely/oui/blob/devel/src/oui/partials/elements/_flexbox.scss#L22-L26.

```scss
@mixin display($value) {
  display: -webkit-#{$value};
  display: -ms-#{$value}box;
  display: $value;
}
```

As far as I can tell, there is something wrong here. Either you want the `display(..)` mixin to be a wrapper around the `display` CSS property, and there is no reason to output WebKit and Microsoft specific variations without checking for the `flex` value. Or you want this to be a helper to initialize a flex container —which seems like given the filename, the code comments and the `box` suffix on line 3— and you don’t need an argument at all.

```scss
@mixin flex-container {
  display: -webkit-flex;
  display: -ms-flexbox;
  display: flex;
}
```

### LINKED ISSUE #242: Flex mixin is likely to break
Flex mixin: https://github.com/optimizely/oui/blob/devel/src/oui/partials/elements/_flexbox.scss#L28-L35.

The condition in the `flex` mixin is likely to break. `$num != '0'` will evaluate to true even if you pass `0` as a number, which is not what you want. Either you need to cast the parameter to a string (which I would not necessarily recommend), or you need to also check for the numeric version of the value (if there is).

```scss
@mixin flex($num) {
  -webkit-flex: $num;
  -ms-flex: $num;
  flex: $num;
  @if ($num != '0' and $num != 0) { // http://stackoverflow.com/questions/29398259/css-flexbox-issue-in-firefox-v36-and-greater
    min-height: 0;
    min-width: 0;
  }
}
```

tomgenoni:
This adds autoprefixer and removes a bunch of mixins no longer needed. Only questionable addition is the class `.flex-overflow-fix` that addresses the overflow issues in Firefox. In some situations developers would need to add this class, or otherwise add a `min-height` value into their CSS. We should probably provide information about this "bug" somewhere since it only applies to 100% height layouts.

Fixes #242
Fixes #241 

@danoc


src/oui/partials/elements/_flexbox.scss
```diff
@@ -6,169 +6,93 @@
...
+  align-self: center;
 }
 
 .flex-self--end {
-  @include align-self(flex-end);
+  align-self: flex-end;
+}
+
+// Flex Firefox layout fix
+// This is triggered when children of flex elements require overflow. Firefox requires a `min-height/width`.
```
danoc: I think this could use an explanation about what it means in practice. AKA does this class need to be added everywhere we use flex?

tomgenoni: Ok, I'll add more explanation. It does not need to be added everywhere, only in a very specific situation.

danoc:
:+1:

tomgenoni:
Fixes #276

PR #279, opened 2016-02-23
Move functions, mixins, and Sass-y things to `partials/sass/`

### LINKED ISSUE #259: Move mixins and functions into a directory of Sass things
A few benefits:

* This directory will be loaded before the rest of the partials so everything will have access to it.
* Clear separation between the mixins and the SCSS that uses it. You can think of the mixins/functions as an API.
* We probably want to keep the number of mixins low to minimize complexity. Keeping them in a dedicated location helps us keep an eye on them.
* Bonus: Keeping it separate makes it easier to test using a tool like [True](https://github.com/oddbird/true).

Files that contain mixins that can be moved:
* partials/elements/_flexbox.scss – Won't need to be moved once we have autoprefixer set up since the mixins will be killed. See #240.
* partials/elements/_functions.scss
* partials/components/_grid.scss 
* partials/components/_matrix.scss
* partials/components/_arrows.scss – Placeholders for `_arrows_inline.scss`
* partials/base/_fonts.scss – This is a weird one, but might make sense in this directory.

cc: @tomgenoni @daverau-optimizely

danoc:
Fixes #259.

cc: @tomgenoni – Feel free to be nit-picky!


tomgenoni:
I think `sass` is okay for the functions/mixins but it doesn't seem appropriate for things like "media". Could they be "objects"? I know that word is loaded.

danoc:
I see how `object/` makes sense, but "media" is a mixin in the same way as `@mixin arrow(...)` (currently in `_mixins.scss`).

Let's chat in person tomorrow!

danoc:
@tomgenoni – Can you take another look at this? I've consolidated mixins, functions, and placeholders into their own files and moved the fonts file back to where it was as we spoke about in person.

src/oui/oui.scss
```diff
@@ -10,9 +10,10 @@ $include-fonts:(
   italics: true
 );
 
-/// OUI functions and mixins
-@import 'partials/elements/functions';
-@import 'partials/elements/mixins';
+/// OUI placeholders, functions, and mixins
+@import 'partials/sass/placeholders';
```
tomgenoni: I think we could put three inside `_partials.scss`. That way variables is first and these files can use them if needed.

tomgenoni:
If you can address the note above then LGTM.

PR #280, opened 2016-02-23
Namespacing option for JS

### LINKED ISSUE #255: Add namespace option for classes in javascript
Currently the javascript injects popovers and poptips with no namespace. They will fail if the Sass has been namespaced, e.g., `.lego-pop`.

tomgenoni:
Running `gulp js --namespace=lego-` where 'lego-' is the namespace, will add it to poptips/popovers classes that are injected onto the page and need to match namespaced CSS. 

@danoc 

Fixes #255


danoc:
LGTM!

PR #285, opened 2016-02-26
add overflow scroll trump

tscanlin:
@danoc this is needed to fix the weird hover bug you and @tjj5036 found. Let me know if you have questions.


danoc:
Can you add a line to the CHANGELOG please? Otherwise LGTM.

tscanlin:
Sure, and should I merge this into devel or master? Ideally I'd like this to go in ASAP to fix the Audience builder and it should just be a patch level change.

danoc:
@tscanlin Just merge into `devel` because we're still 2 versions behind in the app. So this would have to be 9.0.2 but the app is on 7.

I'd recommend just inline styling it in the app TBH. Or add it to an existing class if there is one.

tscanlin:
I will just add the same class to the _targeting.scss file as well for now. Expect a diff in the app with that change soon.

PR #294, opened 2016-03-03
Dropdown lock styling

### LINKED ISSUE #292: Dropdown should assert more styling properties to insulate from inherited styles
screen shot 2016-03-03 at 1 57 36 pm

Here the dropdown is nested inside a `th` and its line-height/font-weight is applying to the dropdown items.

tomgenoni:
Fixes #292 

@danoc


danoc:
LGTM!

cc: @kwalker3690 for reference.

PR #304, opened 2016-03-14
Fixing visual bug with anchors as button

### LINKED ISSUE #303: Anchors as buttons fix
There are some slight differences in the text color between buttons and anchor links using the `button` class. 

These is the current hover states of anchors as buttons.
screen shot 2016-03-14 at 11 14 02 am

These are if `a.button` is added to _button.scss.
screen shot 2016-03-14 at 11 13 50 am

tomgenoni:
Fixes #303

@danoc


danoc:
Shouldn't the styles for `<a>` get loaded before (and get overwritten by) the `button`s?

tomgenoni:
Ah, I found a simpler solution. `a:hover` text color wasn't being set in the button. I've reverted my initial change and added the hover text color.

danoc:
Sweeeet! LGTM.

PR #317, opened 2016-03-22
Removed `appearance` and `keyframes` mixins since we now use Autoprefixer

### LINKED ISSUE #282: Appearance and keyframes mixins can be removed since we use autoprefixer
See `_mixins.scss`.

danoc:
Fixes #282.

cc: @tomgenoni


tomgenoni:
Sweet. LGTM.

PR #322, opened 2016-03-23
Offload deployment tasks to Travis CI and change the commands used to deploy

danoc:
This PR moves a lot of the deployment process to Travis CI. It will deploy whenever a release tag is created and pushed to master.

@tomgenoni – I have no idea if this works because I can't test it until it is merged into `master`, but this is my best guess.

cc: @daverau-optimizely since this fixes some of the problems we ran into when I last showed you how the deploy works.


danoc:
Going to close this and branch off of `master` instead so I can test it out sooner.

PR #327, opened 2016-03-23
Set ACL on Travis CI uploads so that the files are publicly viewable.

### LINKED ISSUE #326: ACL should be set in Travis CI configuration file
This ensures that the uploaded files are publicly viewable.

danoc:
cc: @tomgenoni 

Fixes #326


tomgenoni:
I don't have insight into how this works but I don't see anything off about it.

PR #350, opened 2016-04-04
Updating brand-dark variable color

### LINKED ISSUE #349: Brand-dark is not in color map in design dot
`brand-dark` variable does not match any of the values here: http://design.optimizely.com/brand/colors.html

We should replace it with the closest available value

kwalker3690:
Fixes #349 

@danoc @tomgenoni


danoc:
Sweet, thanks!

Issue #352, opened 2016-04-04
Add dropdown class to support dropdown with sentences of text inside


kwalker3690:
![image](https://cloud.githubusercontent.com/assets/2287470/14260830/3e38fa2c-fa62-11e5-8c51-97ac1226d95e.png)

danoc:
This has been merged and released in v11.0.0. It is now available in the app.

PR #353, opened 2016-04-04
Adding new dropdown class for descriptive dropdowns

kwalker3690:
Adds #352

@danoc @tomgenoni


src/oui/partials/components/_dropdown.scss
```diff
@@ -114,4 +136,23 @@
...
+  &--descriptive {
+    .#{$namespace}dropdown__item {
+      white-space: normal;
+    }
+
+    .#{$namespace}dropdown__block-link {
+      cursor: pointer;
+    }
+
+    &__header {
```
danoc: Should these all have the namespace thing?

kwalker3690: No because they are nested

kwalker3690: They inherit the namespace from above

src/oui/partials/components/_dropdown.scss
```diff
@@ -114,4 +136,23 @@
     right: 0;
     left: auto;
   }
+
+  &--descriptive {
+    .#{$namespace}dropdown__item {
+      white-space: normal;
+    }
+
+    .#{$namespace}dropdown__block-link {
```
danoc: This should probably be applied to all `/dropdown__block-link`, not just within `.dropdown--descriptive`.

PR #362, opened 2016-04-08
Implement eslint for OUI and fix errors/warnings

danoc:
@tscanlin – Can you take a look at this please?

This will be more important as we move to React and replace/refactor our existing JS. I'd like to eventually share the eslintrc file that the Optimizely app uses, but that is in a private repo that needs some work before it is opened up.


tscanlin:
All of Tom's formatting is gone! But it's consistent with the rest of the js code style in the app so this is awesome!! Thanks for making this change Dan! LGTM 👍

Issue #137, opened 2015-09-24
Rounded buttons have round corners next to inputs


kwalker3690:
<img width="302" alt="screenshot 2015-09-23 18 51 00" src="https://cloud.githubusercontent.com/assets/2287470/10063619/d6656b7c-6224-11e5-8fd7-cf97db424513.png">

tomgenoni:
The problem here is that the button/input are double nested inside a `lego-button-group`. Need to remove the outer one.

danoc:
Reopening this one because I'm reverting it until we have time to fix a regression it introduced.

<img width="603" alt="screenshot 2016-04-11 17 47 03" src="https://cloud.githubusercontent.com/assets/916038/14446682/75effece-000d-11e6-8cf0-05113dede49f.png">

The change to the selector in `_button-group.scss` was causing it to target items within the dropdown that is within a button group. The position relative (added on hover and focus) breaks the dropdowns.

PR #367, opened 2016-04-12
Revert part of a commit to `.button-group` that introduced a bug in v10.0.0

danoc:
Related to #137 and BUG-2825.

See the issues for context.

cc: @kwalker3690


kwalker3690:
LGTM

PR #369, opened 2016-04-12
Updated styling of popovers and changed size of line-height tight variable

kwalker3690:
@danoc


CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This file follows the format suggested by [Keep a CHANGELOG](https://github.com/olivierlacan/keep-a-changelog).
 
 ## [Unreleased][Unreleased]
+### Added
+- [Feature] Updated styling of popover content
```
danoc: Can you make this a bit more descriptive?

CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
...
+
+### Changed
+- [Release] Changed size of ($line-height, tight) variable
```
danoc: I'd just consolidate those two things into one under `### Changed` that is a `[Feature]`. I wouldn't say this is breaking since it is a small design tweak.

danoc:
Two small comments and this has a lint issue. See Travis CI log for details.

src/oui/partials/components/_popover.scss
```diff
@@ -132,6 +132,17 @@
     font-weight: 500;
   }
 
+  &__body {
+    margin: spacer(0.5) 0;
```
danoc: Nit: Do the `left` and `right` margins need to be set? If not. then we should do `margin-top: spacer(0.5)` and `margin-bottom: spacer(0.5)`.

danoc:
One nit then LGTM once fixed or if those left/right need to be there.

PR #371, opened 2016-04-12
Restyle button focus to be more consistent with other inputs.

### LINKED ISSUE #355: Styling for :focus
@zleach you mentioned wanting to get rid of the default styling that is shown on `:focus` for things like buttons. We should not get rid of it because it is essential to accessibility, but instead design a style for it that is better than the standard.

danoc:
This is the SCSS portion of the improvements to our `:focus` styles for buttons.

It is purposely scoped to the OUI `button` class, not all buttons, since I think that better aligns with our existing coding patterns.

(Fixes #355)

@kwalker3690 or @tscanlin – Can one of you take a look? See the issue for context.


kwalker3690:
Love it, ship it

PR #373, opened 2016-04-13
Remove linter pre-commit hook to avoid annoying Ruby dependency.

### LINKED ISSUE #370: Remove pre-commit hook
The pre-commit hook runs `npm test` which has an annoying Ruby dependency for scss-lint. (There is no good SCSS linter written in JS.)

Travis CI runs the linters when branches are pushed or pull requests are created. It will prevent merging if there are lint issues, so it's safe to remove the pre-commit hook.

cc: @kwalker3690

danoc:
Fixes #370.

cc: @kwalker3690


kwalker3690:
yay! LGTM

Issue #382, opened 2016-04-21
Research accessibility best practices, share findings, and enforce in pull requests


danoc:
I'll be doing a deep dive into accessibility as we move towards building React components. We can greatly benefit all users if we bake accessibility into these components. A little bit of work will go a long way since we will no longer be duplicating HTML.

### Step 1: Research best practices

This begins with researching best practices. These are a few resources that come to mind, but I'd love pointers towards other places to look.

* Read the [WCAG 2.0](https://www.w3.org/TR/WCAG20/) – These are the latest recommendations from the W3C to make the web more accessible.
* Look at [Salesforce's LDS](https://www.lightningdesignsystem.com/) and 18F's [U.S. Web Design Standards](https://standards.usa.gov/) – These are examples of two design systems that place a huge focus on accessibility.
* [An Alphabet of Accessibility Issues](https://the-pastry-box-project.net/anne-gibson/2014-july-31) – Great list of accessibility issues, some more obvious than others.
* [HTML5 accessibility browser support](https://stevefaulkner.github.io/HTML5accessibility/) – Great checklist of accessible HTML5 features and browser support.

### Step 2: Share findings

There are a few ways to do this, but I'll likely share notes at an upcoming meeting and add an accessibility section to our `CONTRIBUTING.md` file.

### Step 3: Review accessibility in pull requests

I'll start off by creating a checklist of accessibility concerns to review then investigate how we can programmatically enforce some rules. This can likely be done to some extent in automated tests and perhaps there is a accessibility linter I can add to our build system. I'll look into large UI libraries such as Bootstrap and see if they test for accessibility.

daverau-optimizely:
Too broad to do now.

PR #383, opened 2016-04-22
Make the Attention component accessible

danoc:
There are four parts to this:

1. `role="alert"` is added to all Attention components. This causes screen readers to skip to that part of the page.
2. `aria-label` is added to the Attention component and the value depends on the color class (`bad-news`, `brand`) that is supplied to the component. This label conveys the the meaning that we get from the color of the component.
3. The dismiss link was changed to a button. Links are best for navigating to different resources while buttons are best for performing actions that change the underlying page.
4. `ariaLabel` was added to the button because `&times;` doesn't convey that the button will close the alert.

I also added tests for these where appropriate.

This relates to #382.

cc: @tscanlin, @kwalker3690, @daverau-optimizely


.eslintrc.js
```diff
@@ -180,5 +181,8 @@ module.exports = {
     "no-bitwise": "off",
     "no-plusplus": "off",
     "jsx-quotes": ["warn", "prefer-double"],
+
+    // OUI rules
+    "valid-jsdoc": ["warn"],
```
tscanlin: Nice!

tscanlin:
LGTM, nice work!!

PR #384, opened 2016-04-22
Make ArrowsInline more accessible

danoc:
This applies `aria-hidden="true"` to the inline arrows (chevrons) because, on their own, they don't provide much information. `aria-label`'s or helpful text should be provided by the components that are using this component.

cc: @daverau-optimizely @kwalker3690 @tscanlin 

Relates to #382.


tscanlin:
+1 LGTM

PR #385, opened 2016-04-22
Create test helper functions and add tests for them.

danoc:
This creates helper functions for tests that adds `data-test-section` support and abstracts away some React quicks with stateless components.

It also updates the tests to use he helper functions and adds new tests for the helper functions.

***
cc: @tscanlin @kwalker3690 @daverau-optimizely


src/utils/test-helpers.js
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Grab the first child of a ReactElement. This is useful after calling the
+ * `renderIntoDocument` helper function that wraps a component in a `div`. This
+ * is needed because of a limitation of stateless React components.
+ * https://github.com/facebook/react/issues/4839
+ * @param {ReactElement} component - JSX returned in a function
+ * @returns {HTMLElement} Orignal component wrapped in a `div` and rendered
+ * into the document.
+ */
+export const getComponentNodeFromComponent = (component) => {
```
tscanlin: Can we maybe rename this to `getNodeFromComponent` since that's shorter than `getComponentNodeFromComponent`?

tscanlin:
One comment about a shorter method name, otherwise this looks good!! :)

src/components/Button/tests/index.js
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable no-unused-vars */
 import React from 'react';
 import ReactDOM from 'react-dom';
-import TestUtils from 'react-addons-test-utils';
+import ReactTestUtils from 'react-addons-test-utils';
```
tscanlin: Should lines 3 & 4 here just become `import * as testHelpers from '../../../utils/test-helpers';` ?

tscanlin:
LGTM!

src/utils/test-helpers.js
danoc: Good point. Done!

src/components/Button/tests/index.js
danoc: Yeah, good point! I've actually changed it to include both since this is just boilerplate code and there's a chance that both would be needed.

PR #386, opened 2016-04-25
Add tests for code sample

danoc:
This adds a few tests for code samples and prevents them from being wrapped in `div`s.

cc: @tscanlin @daverau-optimizely @kwalker3690

### DIFF
--- a/src/components/Code/index.js
+++ b/src/components/Code/index.js
@@ -15,12 +15,11 @@ let blockCode = (children) => {
 };
 
 const Code = ({ children, type }) => {
-  return (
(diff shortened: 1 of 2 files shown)


tscanlin:
Looks like travis isn't liking the new es6 string format. Maybe we just need to add / change a rule for that.

Otherwise LGTM!

danoc:
@tscanlin – Yeah, I realized it's because ESLint is only cool with that format if it is being used as a multi-line string (since that is the main benefit).

Thanks for the review!

PR #387, opened 2016-04-25
add react Input component

tscanlin:
Starting to work on an Input component.

@danoc @daverau-optimizely @kwalker3690 let me know if you have any feedback.

I'm going to add my own questions inline.


src/components/Input/index.js
```diff
@@ -0,0 +1,61 @@
...
+  defaultValue,
+  placeholder,
+  required,
+  readonly,
+  disabled,
+  onInput,
+  onChange,
+}) => {
+  return (
+    <input type={type}
```
tscanlin: I'm thinking it may just make sense to use `{... this.props}` here instead. Using the spread operator like that would cut down on boilerplate / repetition. What do you think?

src/components/Input/index.js
```diff
@@ -0,0 +1,61 @@
...
+      disabled={disabled}
+      onInput={onInput}
+      onChange={onChange}
+    />
+  );
+};
+
+Input.propTypes = {
+  type: React.PropTypes.oneOf([
+    'text',
```
tscanlin: Not sure if all these are necessary or if they should all be supported from the get go, but they were taken from: https://www.w3.org/TR/html-markup/input.html

Maybe we should comment some out and include a note in the source for why we won't support for now. (For example because browser support isn't good enough yet).

src/components/Input/tests/index.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import * as testHelpers from '../../../utils/test-helpers';
+import Input from '../index';
+
+describe('components/Input', () => {
+  it('should render "text" input when type: "text" is passed', () => {
```
tscanlin: Gonna need more tests for this.. but gotta start somewhere ;)

tscanlin:
This is a work in progress by the way.

src/components/Input/index.js
kwalker3690: Agreed, I think we should only include the ones we already use in the product, which I don't think is all of those? Then we can add more if needed later

danoc: Yeah, I agree with @kwalker3690. Especially since a component doesn't need to map to an HTML element. We can decide what functionality we do or don't want to support.

src/components/Input/index.js
```diff
@@ -0,0 +1,62 @@
...
+    'search',
+    'tel',
+    'color',
+  ]).isRequired,
+  value: React.PropTypes.string,
+  defaultValue: React.PropTypes.string,
+  placeholder: React.PropTypes.string,
+  required: React.PropTypes.bool,
+  readOnly: React.PropTypes.bool,
+  disabled: React.PropTypes.bool,
```
danoc: I think the other files call this `isDisabled` and `isReadOnly`, `isRequired`, etc... I don't have a strong preference but think we should be consistent.

src/components/Input/index.js
```diff
@@ -0,0 +1,62 @@
...
+    'number',
+    'range',
+    'email',
+    'url',
+    'search',
+    'tel',
+    'color',
+  ]).isRequired,
+  value: React.PropTypes.string,
+  defaultValue: React.PropTypes.string,
```
danoc: How is this different than `value`?

src/components/Input/index.js
```diff
@@ -0,0 +1,62 @@
...
+    'text',
+    'password',
+    'number',
+    'date',
+    'email',
+    'url',
+    'search',
+    'tel',
+  ];
+  if (textInputs.indexOf(props.type) !== -1) {
```
danoc: We can kill this if we trim that list of supported types.

src/components/Input/index.js
```diff
@@ -0,0 +1,62 @@
...
+  );
+};
+
+Input.propTypes = {
+  type: React.PropTypes.oneOf([
+    'text',
+    'password',
+    'checkbox',
+    'radio',
+    'button',
```
danoc: Should we require using the `Button` component for buttons? I don't see a case where `input[type=button]` is better.

src/components/Input/index.js
tscanlin: I think we will still need some way of determining if the `text-input` class should be added. For instance checkboxes and radio input's won't need it. How would you imagine that would work?

src/components/Input/index.js
tscanlin: I will trim down the list.

src/components/Input/index.js
tscanlin: +1 I will make sure to add that one to the unsupported list :)

src/components/Input/index.js
tscanlin: See the react docs:
http://facebook.github.io/react/docs/forms.html#default-value

Basically defaultValue is useful for setting the initial value of an input when it's an *uncontrolled* component.

src/components/Input/index.js
tscanlin: Agreed. I will change to add "is".

tscanlin: actually, the nice thing about leaving the "is" part off is that it allows you to do `{...props}` and have everything just work since it's using native react props. Otherwise there is an extra mapping step.

src/components/Input/index.js
danoc: Oh yeah, good point! FWIW though, a bunch of the other libraries split radio and checkboxes into separate components (Rebass and React Bootstrap come to mind).

src/components/Input/index.js
danoc: Oh, neat. Thanks!

src/components/Input/index.js
danoc: Fair, but I'd argue that it's better for us to have a whitelist of props we want to support well instead of allowing all of them and setting restrictions on the ones we specify in `propTypes`.

One of the nice things about whitelisting the props is that we can ensure that the component will behave perfectly in all of the "states" we decide to support. This is no longer true if [we allow all props](https://facebook.github.io/react/docs/tags-and-attributes.html#html-attributes).

Also, FWIW, I think it would be a bit more confusing for our users. In the current implementation, for example, a parent component that passes in a `className` prop would overwrite the class that you have decided the component should use. It's really nice when all of the props are listed in `propTypes` because it becomes a great way to document the component's "API."

src/components/Input/index.js
tscanlin: That's a good point. I will split those two out into their own components.

src/components/Input/index.js
tscanlin: That is true @danoc. And I'm open to other options regarding this, you bring up some really good points.

As far as the `className` example you mention. That is true in this case, and by design. But if we want to make it so className can't be overridden, I think we could just reverse the ordering of the props I'm passing in like so: `{...props} className={classes}`

Also I do want the propTypes validation object to be the limit of which props are supported for exactly the reasons you mention. It's really nice to know all the options that are available and also know that other options which aren't declared can't be passed in. I'm going to see if there is any way to declare that any props not defined in propTypes should be forbidden.

Thanks for the feedback, this in particular is a great discussion point!! :+1:

tscanlin: @danoc @kwalker3690 @daverau-optimizely What do you think about an approach such as:

```
  var validProps = {};
  for (var prop in Input.propTypes) {
    validProps[prop] = props[prop];
  }

  return (
    <input className={classes} {...validProps} />
  );
```

That way all the valid props for a component should be validated in someway and declared on the propTypes object. I think this limits repetitive / boilerplate code, while also satisfying our earlier requirement that unspecified props shouldn't be allowed. If we do decide to go this route I can write a helper for this. Let me know what you think.

danoc: I feel like that's adding a lot of complexity and making the code harder to read when the original intention of `{...validProps}` is to make the code easier to read. It also means that we'd have to define the props in two places since they will also appear in `propTypes`.

As far as the `className` prop, I think that we should discourage people from passing in their own classes (or overriding ours) if we want OUI to encourage consistent and high quality design.

src/components/Input/index.js
danoc: Feel free to do this in a separate PR if you'd like!

tscanlin: Yeah, definitely. This one is big enough. I'll do the other components separately.

src/components/Input/index.js
tscanlin: @danoc So you think this is going to be easier to read?

```
return (
    <input type={type}
      value={defaultValue}
      placeholder={placeholder}
      required={required}
      readonly={readonly}
      disabled={disabled}
      onInput={onInput}
      onChange={onChange}
   />
);
```

I agree that declaring props like that is more explicit, but it also could be error prone if you forget to pass the prop or if you misspell a prop. And it seems to be adding more boiler plate / repetition because I need to define the prop in `propTypes` but also make sure to pass it explicitly. It could get ugly if we start passing a lot of props..

@daverau-optimizely @kwalker3690 do either of you have an opinion?

tscanlin: I've convinced myself that the explicit way is probably better so I'm going to switch to doing that. I hope this encourages us to be more deliberate and avoid passing too many props.

tscanlin:
I need to look into the lint stuff, but I'll let you guys know tomorrow when this is good for another pass.

tscanlin:
@danoc @kwalker3690 @daverau-optimizely I've fixed the lint errors and added some more tests for this component. Can you give this another pass when you get a chance? Thanks!

tscanlin:
rebasing..

src/components/Input/index.js
```diff
@@ -0,0 +1,77 @@
...
+      readOnly={props.isReadOnly}
+      disabled={props.isDisabled}
+      onInput={props.onInput}
+      onChange={props.onChange}
+    />
+  );
+};
+
+Input.propTypes = {
+  type: function(props, propName, componentName) {
```
danoc: Although I see the benefit of stating the unsupported types, I think that is best left to the documentation system we end up using. We're currently adding a lot of extra complexity to this file (50% of the file) for the sake of documentation, but this custom `propType` will likely make it harder to automatically generate documentation compared to the built in `oneOf[...].required` `propType`.

tscanlin: That's true, it would simplify this quite a bit. I'll make that change.

src/utils/test-helpers.js
```diff
@@ -35,7 +35,7 @@ export const getNodeFromComponent = (component) => {
  * Find an element that contains an attribute data-test-section equal to the
  * provided value.
  * @param {ReactElement} component - JSX returned in a function
- * @param {string} testSection - unique label identifying an element
+ * @param {String} testSection - unique label identifying an element
```
danoc: Can you please add this to the lint rules? http://eslint.org/docs/rules/valid-jsdoc#prefertype

tscanlin: Sure, what should we enforce though? Should we go all uppercase? Do any of you have a preference?

src/utils/test-helpers.js
```diff
@@ -44,3 +44,5 @@ export const getTestSectionFromComponent = (component, testSection) => {
 
   return element;
 };
+
+export const simulate = ReactTestUtils.Simulate;
```
danoc: I don't really know if this is a good or bad pattern, but if we're going to do it I think we should add JSDocs that explain what the function does. 

One way to look at this is that by adding it to our `test-helper.js` we're choosing to provide a simulate function to people that use this file. It happens that `ReactTestUtils.Simulate` currently fills that need, but if it became deprecated or changes functionality, we would still need to "support" the simulate function by either implementing it ourselves or finding another solution. The JSDocs help clarify the expected behavior of our simulate function.

src/utils/test-helpers.js
danoc: I don't have a preference 😄 

(Neither does JSDoc, too: https://github.com/jsdoc3/jsdoc/issues/1046)

src/components/Input/index.js
danoc: Awesome! This is looking really good.

src/utils/test-helpers.js
tscanlin: I'm gonna go with all caps then. I think we tend to do that more in the product so hopefully this makes it more consistent.

src/utils/test-helpers.js
tscanlin: I think it helps us avoid having to import a ton of different modules to get test helpers. If this file becomes unwieldy we can figure out ways to split it up, but for now I think it makes sense to keep this functionality here.

src/components/Input/tests/index.js
```diff
@@ -0,0 +1,90 @@
...
+      <Input type="number" />
+    );
+
+    const componentNode = testHelpers.getNodeFromComponent(component);
+
+    expect(componentNode.type).toBe('number');
+  });
+
+  it('should output an error if an invalid "type" is passed', () => {
+    spyOn(console, 'error').and.stub();
```
danoc: Do we have to unstub these or does that happen automatically?

tscanlin: it happens automatically because it only remains stubbed inside the `it` block

danoc:
LGTM! There seems to be a Travis CI issue with NPM install, but that seems unrelated. I'll look into those later.

tscanlin:
Thanks @danoc!

Yeah, I noticed the Travis CI issue with NPM install, but it looks like everything built ok anyway. Some quick googling pointed me here: https://github.com/travis-ci/travis-ci/issues/4771

Which seems to indicate that we just need to add this line:
https://github.com/redis/hiredis-node/blob/master/.travis.yml#L4

PR #389, opened 2016-04-28
add Label component and some tests

tscanlin:
@danoc @kwalker3690 @daverau-optimizely 

Here is a Label component. Let me know if you have any thoughts!


src/components/Input/index.js
```diff
@@ -17,7 +18,26 @@ const Input = (props) => {
   );
 };
 
+const Input = (props) => {
```
tscanlin: Rather than having Input also able to accept a label, I think it might make more sense to have `InputWithLabel` as its own component that combines these two. I know this is a different approach than Rebass has taken, but what do you think? Should Input just handle it all, or should we make another component that composes both of them?

daverau-optimizely: Is there a downside to having input and label as their own, and a third that combines them?

tscanlin: I think it separates concerns better and is a good example of functional composition. The downside is that there is another component to think about and that we need to be explicit in our docs / examples to show that the primary way to build UIs is by using higher level components that combine these smaller pieces for you.

danoc: I think the current implementation is better IMO! Although `Label` is it's own component, to the developer it feels like a string that is passed to an input or a checkbox.

src/components/Label/index.js
```diff
@@ -0,0 +1,21 @@
...
+
+  return (
+    <label className={classes}>{ props.children }</label>
+  );
+};
+
+Label.propTypes = {
+  children: React.PropTypes.oneOfType([
+    React.PropTypes.string.isRequired,
+    React.PropTypes.array.isRequired,
```
danoc: Why array?

tscanlin: If DOM elements are inside a react component then it treats them as an array. It only seems to treat props.children as a string if text content is passed in. I think it makes sense to support both.

src/components/Input/index.js
tscanlin: That's fair. I'm ok with keeping this as it is and we can always revisit later if it is getting too complex. Thanks for the feedback @daverau-optimizely & @danoc!

src/components/Label/index.js
danoc: Okay cool gonna test this out locally a bit.

tscanlin:
@danoc @daverau-optimizely @kwalker3690 

Tests and lint are passing now. Could you guys give this another pass?

src/components/Input/index.js
```diff
@@ -36,6 +52,8 @@ Input.propTypes = {
   isDisabled: React.PropTypes.bool,
   onInput: React.PropTypes.func,
   onChange: React.PropTypes.func,
+  label: React.PropTypes.string,
 };
+renderInput.propTypes = Input.propTypes;
```
danoc: Is this needed? I think `Input.propTypes` already does the trick.

tscanlin: It is needed because otherwise I get eslint errors since the other component / function doesn't have props validation.

danoc:
I did a bit of digging and, FWIW, wrapping the children of `label` break if we want a layout where a button is next to a text input.

This code, for example, is invalid HTML:

```html
<label>
  Hello Label
  <input type="text">
  <button>Button</button>
</label>
```

screenshot 2016-04-29 17 09 22

Chrome assumes that the label refers to the first labelable element, but that's not necessarily guaranteed.

I think this is fine for v1 though and we can always change if it comes back to bite us. Just take a look at the comment I left about the extra `propTypes`.

danoc:
LGTM!

tscanlin:
Thanks @danoc! I agree that the button / label thing could potentially cause issues down the road especially if that will be a common pattern, but thanks for letting me merge this! We can always revisit our approach down the road if it is causing issues.

PR #390, opened 2016-04-28
switch away from using destructuring

tscanlin:
@danoc @daverau-optimizely @kwalker3690 

This switches us away from destructuring which can get kinda repetitive with lots of props. And also changes some `let`'s to `const`.


danoc:
Cool, thanks @tscanlin!

Did you happen to find a lint rule that can enforce this? Feel free to merge regardless.

tscanlin:
I did not find a lint rule for this unfortunately. But hopefully when we start on more components they copy the structure of one that has already been built so that it will be written in the same style as the others. I'm gonna merge this.

Issue #361, opened 2016-04-07
Prefix all the classes


danoc:
There are a few benefits towards prefixing/namespacing all classes:
1. When looking at HTML (outside the DevTools) it is easier to find out where a style is defined. For example, an engineer trying to figure out how `.center--child` works might search through the codebase and be confused when the search turns up no SCSS results.
2. Find and replaces become _waaaay_ easier. OUI v10.0.0 has a breaking change that renames `.wrap` to `.wrap` to `.wrap-text`. The find and replace is incredibly hard to do because `wrap` is a very common term in our HTML files. ([This breaking change could have been avoided](https://github.com/optimizely/oui/issues/301) had it already been namespaced.)

The argument for making namespacing optional is that it is not needed if you are starting with a fresh project. I'd say that the two benefits still apply to projects built on top of OUI. This is especially true since we encourage developers to build SCSS partials within their codebase as well.

The downsides of prefixing everything is that it adds a few more characters. This adds a bit of extra weight to the page and clutters the HTML more.

I think the benefits far outweigh the downsides though.

---

@kwalker3690, @daverau-optimizely, @tscanlin – What do y'all think?

tscanlin:
+1 I like this idea!

daverau-optimizely:
Would better naming up front have avoided this breaking change as well? Including more context of what the wrap is doing seems helpful.

I like the idea of better/clear naming of things so if namespacing isn't too heavy of an approach I'm for it.

danoc:
@daverau-optimizely – Yes it would. The breaking change was needed because p13n already has a class called wrap that does something else.

danoc:
So one way to do this would be to modify the SCSS for all classes to temporarily add the prefix along with the current class name.

```css
.wrap,
.oui-wrap {
  foo: bar;
}

.#{$namespace}attention,
.oui-attention {
  bar: foo;
}
```

We can leave it like this for a few months while we tackle renaming the classes in the Optimizely app (maybe even limit this to `v2.css`). Then we'd deploy a custom version, do some testing, and eventually remove the non-prefixed classes.

daverau-optimizely:
This is done, so closing this and will consider a new issue when we get rid of lego and possibly remove the prefix option all together.

Issue #356, opened 2016-04-05
Speed up `npm install` for applications that use OUI


danoc:
@jordangarcia pointed out that OUI is installing slowly in NPM.

There are a few ways to improve this:
* Remove devDependencies before publishing. Notice how React has so few: https://github.com/facebook/react/blob/80bff5397bf854750dbe7c286f61654ea58938c5/packages/react/package.json
* Use more command line tools instead of `gulp-*` packages that just wrap commands.
* Add `npm install browsersync` to `npm start`, not as a devDependency.

danoc:
This is done. See `.travis.yml` and search for `devDependencies`.

PR #392, opened 2016-05-04
Prepare OUI to allow other repos to require components

danoc:
This changes some `package.json`, webpack, and Babel config to create a `dist/` folder on `postinstall` with our transpiled JS. The transpilation is needed for Babel to convert ES6 to ES5 and JSX to JS so that other repos can use it.

The main entry point to OUI becomes `dist/main.js`. That file will export all of the OUI React components.

We should later change this from a `postinstall` and make the `dist/` part of the published NPM module (without having to check the `dist/` into git). See issue #356 for context.

This diff also begins adding `oui-` prefixed classes for all of the classes that our React components are using. This will help us complete #361 and avoid the issues that different namespaces can bring.

*** 
cc: @tscanlin, @daverau-optimizely, @kwalker3690


danoc:
Sorry for all the followup commits. Should now be good to review.

package.json
```diff
@@ -3,9 +3,11 @@
   "version": "11.3.0",
   "description": "Optimizely's Component Library.",
   "private": false,
-  "dependencies": {},
+  "main": "dist/main.js",
```
tscanlin: Does this affect the `includePath` stuff we use to require scss?

package.json
```diff
@@ -3,9 +3,11 @@
...
+  "dependencies": {
+    "babel-cli": "^6.7.7"
```
tscanlin: why wouldn't this be a `devDependency`?

package.json
```diff
@@ -44,6 +46,7 @@
     "jest": "jest --config .jest.config.json",
     "eslint": "eslint --ext .jsx --ext .js . --ignore-path .gitignore --max-warnings 0",
     "scss-lint": "bundle exec scss-lint src/ --exclude src/oui/library/**/*",
+    "postinstall": "babel src/ --out-dir dist/ --ignore \"tests/*.js, src/js, src/index.jsx\"",
```
tscanlin: Remember that this might break for some people if they have old npm versions... :/

tscanlin:
Aren't there still a lot of classes without the `oui-` prefix? Is the plan to do all of them in this PR?

I mostly had some minor feedback but this is looking really good! Thanks for tackling this!

package.json
danoc: It's needed for the `postinstall` hook so it needs to be a dependency. Later on (before releasing the React branch) we'll make it a pre-release step and be able to move it back to `devDependencies`. We can't do that yet since we can't release this branch as a version of OUI.

package.json
danoc: Ahh, right. I'll fix this later before merging the OUI React branch. There are a few ways to do that.

danoc:
@tscanlin – I've only done the ones that we are using in the React components. That allows us to not worry about the app vs OUI prefix differences and keep the changes in these branches scoped to the main goal.

package.json
danoc: Nice catch! It does, but I've already updated a branch in the app to get around that.

tscanlin:
Thanks for addressing my points, LGTM

PR #394, opened 2016-05-06
Add onClick prop to buttons

danoc:
This allows parent components to pass in a function that gets executed when a button is clicked.

cc: @tscanlin, @daverau-optimizely, @kwalker3690


tscanlin:
LGTM!

daverau-optimizely:
👍

PR #402, opened 2016-05-25
Add `gulp watch` task for css+js

daverau-optimizely:
@danoc was there a reason we didn't include this before? Just checking with ya!


danoc:
I just realized we have a task for `gulp html-examples` that might do exactly what you want.

PR #407, opened 2016-05-31
Explicitly define the placeholder colors 

### LINKED ISSUE #405: Set a color on placeholder text for both inputs and textareas
@amy-chiu was adding a placeholder to a textarea and the color was appearing as black. I don't think it was being styled with CSS, so this may be a browser default. If not, then it's probably a CodeMirror issue.

screen shot 2016-05-27 at 2 01 37 pm

If it is the default browser color, then we should add CSS to style all placeholders to be the same color:
https://css-tricks.com/snippets/css/style-placeholder-text/

danoc:
This explicitly defines input placeholder colors (to what is essentially the browser default) and adds support to style CodeMirror's fake placeholders.

Fixes #405.

cc: @kwalker3690, @amy-chiu


kwalker3690:
LGTM!

PR #412, opened 2016-06-10
First react component - tabs

kwalker3690:
Hey @daverau-optimizely and @danoc, here's my first attempt at an oui react component, tabs! This PR doesn't have any tests in it yet, just wanted to get your guys feedback on this first pass! Let me know what you think


danoc:
I'll take a look at this when I get back next Monday, but check out these links I have bookmarked that might be helpful:

* http://frend.co/components/tabs/
* https://cloudflare.github.io/cf-ui/#cf-component-tabs

(I just skimmed those FWIW – don't know how good they are.)

I'd also reach out to @jordangarcia and @jessicalc because we've talked about how to build tabs before.

src/components/Tabs/index.js
```diff
@@ -0,0 +1,98 @@
+import React from 'react';
+import classNames from 'classnames';
+
+class Tabs extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
```
danoc: Let's chat with Jordan about this. I think the preferred approach is to create only stateless components and he has some opinions on the best way to create tab components.

danoc: He sent me this a while back FWIW:

```
(props) => {
  return (
    <TabContainer>
      <Tab isActive={props.activeTab === 'one'}>Tab one</Tab>
      <Tab isActive={props.activeTab === 'two'}>Tab two</Tab>
      <Tab isActive={props.activeTab === 'tree'}>Tab tree</Tab>
    </TabContainer>
  )
}
```

kwalker3690: Oh interesting! Ok that seems like an easy change to make

tommygiglio:
Just thought I'd creep on your work. Nice work, all. 👍 

-The Ghost of Tommy

kwalker3690:
Closed because this is not a PR to devel

PR #419, opened 2016-06-17
Better HTML examples

daverau-optimizely:
cc @kwalker3690 @danoc This technically is a breaking change with the classname fix, but i'm calling this a patch for simplicity. 

I'm adding tokens and badges to OUI now and will release a new version shortly with those two pull requests.

### DIFF
--- a/src/oui/partials/trumps/_type.scss
+++ b/src/oui/partials/trumps/_type.scss
@@ -43,29 +43,6 @@
 }
 
 
-/// Admin Color Helpers
-/// @description Use to label admin-gated features for internal use only.
-/// @example[html] .color--admin
-///   <h3 href="#" class="color--admin">Some Gated Feature Title</h3>
-/// @example[html] .color-admin--background
-///   <ul class="badge"><li class="color-admin--background">Admin</li></ul>
-/// @example[html] .color-admin--border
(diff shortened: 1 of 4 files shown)


daverau-optimizely:
Here's a preview once I update scribesass and deploy:
screenshot 2016-06-17 13 02 55

src/oui/partials/trumps/_type.scss
```diff
@@ -43,6 +43,29 @@
...
+///   <ul class="badge"><li class="color-admin--background">Admin</li></ul>
+/// @example[html] .color-admin--border
+///   <div class="attention color-admin--border">Some Gated Feature Box</div>
+/// @advanced
+
+.color--admin {
+  color: map-fetch($color, ui, admin) !important;
+}
+
+.color-admin--border {
```
kwalker3690: We should be consistent here, either `.color-admin`, `.color-admin--border` or `.color--admin`, `.color--admin--border`

kwalker3690:
Why is it simpler for this to be a patch instead of a breaking change?

kwalker3690:
LGTM, I guess just patching it is ok because its a quick follow up, but technically is a breaking change...but yeah let's keep it to a patch increase

daverau-optimizely:
Updated admin changelog entry to feature, for correctness.

Issue #378, opened 2016-04-18
Create OUI React components


danoc:
Optimizely will soon be introducing React to the front end. It would be great to have a collection of OUI React components ready to go for engineers to use on day one. This will greatly increase UI consistency and improve developer productivity.

Here's a brain-dump of the work required to create these components. The steps are ordered but they can mostly be done in parallel. Please leave feedback—especially if any of this sounds wonky!

### Step 1: Identify components and properties

We should identify the components that we'd like to create. Although we have a "Components" section in our documentation, some of the items would be split into smaller components. `_forms.scss`, for example, contains code for text inputs, textareas, radio buttons, and a number of other potential components. (The resources at the bottom of this post have great example components.)

I've started identifying our components in a [JSON file within the `daniel/patterns.json`](https://github.com/optimizely/oui/blob/daniel/patterns.json/patterns.json) branch. This `patterns.json` file  is will serve source of truth (at least for now) for the list of components and their properties.

![screenshot 2016-04-18 10 55 05](https://cloud.githubusercontent.com/assets/916038/14614016/093fa792-0554-11e6-8e6c-f7aea1051f82.png)

I'd like for the React components to eventually pull component data and "rules" from `patterns.json`. The rules would map to props and propTypes. This is a step towards making our components language and framework agnostic and has two main benefits:
 
1. The `patterns.json` file can be used to generate an OUI React documentation site, keep design.optimizely.com up to date, and power React component propTypes. It can also be used on non-React websites (even static websites) to enforce that patterns are being used appropriately and throw client-side console errors as React does with props.
2. It becomes easier for non-technical designers to view and potentially edit the patterns.

**Update (June 27, 2016):** I've tabled the idea of a `patterns.json` file but would like to revisit it in the future.

### Step 2: Set up React development environment within OUI

I've added React, webpack, babel, and some other tools in the [`daniel/react`](https://github.com/optimizely/oui/tree/daniel/react) branch. You can currently checkout the branch and run `npm run react:serve`, but it doesn't yet watch files for changes.

**Update (March 18, 2016):** I've updated the build tools to fix the development environment issues. `npm run react:serve` will now both watch the files, re-run webpack, and reload browser-sync.

### Step 3: Move towards a componentized directory structure

This is described in issue #368 and depends on identifying the components. We can merge this before completing the other steps.

### Step 4: Build the stateless React components

[...]

Issue #408, opened 2016-06-02
Add info to `CONTRIBUTING.md` about updating Optimizely's `package.json` after releasing


danoc:
Our Docker containers only re-run `npm install` when the app's `package.json` changes. We should add a note to the documentation that says to change the `package.json` whenever we want to invalidate that cache.

**Example:** I added a CodeMirror SCSS file to OUI and released a new version. I ran `npm install` within Optimizely to grab the new version of OUI and imported the SCSS file in our `v2.scss`. It worked locally, but broke the build because Docker did not reinstall the `node_modules/`. I fixed this by updating `^11.3.0` to `^11.4.0` in Optimizely's `package.json` to invalidate the cache.

***
cc: @kwalker3690, @daverau-optimizely

danoc:
Also, FWIW, the `node_modules/` only changes on deploy if the `package.json` changes.

danoc:
Done!

Issue #360, opened 2016-04-07
Implement a better release process to make upgrades less painful


danoc:
OUI has been releasing often and, due to the "everything is global" nature of CSS, many changes are not backwards compatible. This leads to _huge_ releases full of breaking changes. 

**Problems with the current system:**

1. Upgrading the Optimizely app to support a large release is painful and error-prone.
2. "Unreleased" breaking changes prevent us from being able to release smaller patch and feature (backwards compatible) changes. This is mostly a problem in the Optimizely app since other repositories have less incentive to upgrade (especially since the documentation is now versioned).

FWIW, I think it's fine that we release often because our audience (ourselves and third-party Optimizely Apps developers) are either willing to do the upgrade work or are content sticking to the version they are on.

**Proposal to fix this problems:**

We can fix these problems by adopting a new release policy: All developers that want to merge a breaking change must release a version after merging and update the Optimizely app to support it.

If we follow this:

1. Breaking changes should never sit in the "Unreleased" queue for more than a few minutes.
2. OUI should always be ready to release a backwards compatible release.
3. Optimizely should always be on the latest version of OUI to allow backwards compatible changes to be pulled in automatically.

**Potential pain points:** 

1. We currently only care that the Optimizely repo is on the latest version of OUI. This will be more painful if we have to always upgrade multiple repositories.
2. We might begin to second-guess if breaking changes are worth the hassle of upgrading. (I'd also argue that this can be a good thing.)
3. We'll have to communicate to ensure that we're not trying to merge two breaking changes at the same-ish time.

**Longer term solution:**

Tom and I spoke a lot about this, but we may want to investigate splitting the components off into small individual repos. I'm concerned about increased maintenance cost and a (potentially) weird dependency tree, but apparently the [Financial Times has been doing this successfully](http://origami.ft.com/) and it is similar to how [Intuit](https://github.com/inuitcss) is structured. I think this is worth investigating in a few months.

Regardless, the first step towards this would be grouping components within their own directories. I think that is something we should start doing to make our codebase more modular. (I'll create another issue about that later.)

***

@tscanlin, @kwalker3690, @daverau-optimizely – I'd love to hear thoughts and ideas!

kwalker3690:
[...]

Issue #359, opened 2016-04-07
The link to ScribeSass in the readme points to a 404 page


kaelig:
Looks like https://github.com/optimizely/scribesass is a private repository?

danoc:
Heh, yes it is. We should just remove the link for now or add a note that it is private. I'd like to open source it in the future but it's far from ready and I can't prioritize that work now.

danoc:
I've added a note that the repository is currently private. 

Hoping to make big improvements to it in the future once the dust settles on our React project and then open source it. Maybe I can use the new [scss-parser](https://github.com/salesforce-ux/scss-parser) your team created instead of relying on gonzales-pe.

PR #428, opened 2016-06-27
Add React components to OUI

danoc:
This pull request adds React to the OUI codebase, includes a few OUI React components, and makes a few other modifications. See https://github.com/optimizely/oui/issues/378 or go/react-doc (internal only) for more context.

##### New React components

We have dozens of components to create, but these are the first that we tackled.

- ArrowsInline
- Attention
- Button
- Code
- Input
- Label

Next steps should be to create a prioritized list of components that need to be created.

##### Improved Testing: ESLint and Jest

All components include tests that live within the `<ComponentName>/tests` folder and are run with Jest. The linter was also extended to add the ESLint React plugin. These are run in Travis CI and can be run manually with the `npm test` command.

##### Faster `npm install` for projects that use OUI

We have tons of `devDependencies` that get unnecessarily downloaded by repositories that use OUI. This PR includes a change to remove the `devDependencies` from `package.json` right before publishing to NPM so that projects using OUI only get the packages they need.

##### Auto generated documentation that gets republished on deploy

We use JSDoc and [react-docgen](https://github.com/reactjs/react-docgen) to automatically generate documentation that gets deployed by Travis CI on the release of a new version.

Run `npm run react-docs:serve` locally to view the documentation.

##### Adopting the `oui-` class prefix

I've written a bit about the [importance of namespacing all of our classes](https://github.com/optimizely/oui/issues/361) with `oui-` and am [using the move to React](https://github.com/optimizely/oui/issues/361#issuecomment-213478237) as an opportunity to kickstart the process.

In practice, this means that OUI React components will only use classes that begin with the `oui-` prefix. See a few of the components in this PR for examples.

##### How to run this branch locally

1. Checkout the branch
2. `npm install`
3. `npm run react:serve`

##### How the build + development environment works

OUI uses webpack and Babel to convert our React components and ES6 into plain vanilla JS. It performs this compilations when running `npm run react:serve` locally and right before publishing to NPM. The latter allows us to use these tools without requiring other repos to download them.

##### To-do in this PR:

- [x] Update markdown files to include information about React
- [x] Add @daverau-optimizely's automatic documentation generator

*** 
@daverau-optimizely, @jordangarcia, @kwalker3690 – Can you please take a look? When reviewing, feel free to skip over the `example/*.html` files. Those were just find-and-replace and not critical.


kwalker3690:
Are the only SCSS changes adding `oui-` prefixes to classes that have a namespace on them already? It's a bit hard to see if classes may have been changed to add an `oui-` prefix where there wasn't a namespace before, and would then break.

In the future, I'd say we should break down PR's like this into smaller chunks, like we could have made the scss changes first, then (depending on how interlocked these things are) the dev tools changes so that by the time we wanted to add the react components those fundamentals would already be there. It's pretty tough to diligently code review these many changes! 

That being said, I've checked out the new docs in both places and every thing seems to be compiling well. Awesome work! I'm stoked to start implementing more of these components and getting them out into the wild of Optimizely's product. But I still agree with what you said about labeling this release as beta so that we can accurately check out all of these updates.

kwalker3690:
Oh also update the changelog!

danoc:
@kwalker3690 – Yeah, +1 on breaking these into smaller diffs in the future.

The `oui-` prefix shouldn't break any existing things, I think, because it isn't replacing any of the old class names. It's simply another available class. Eventually there will be a breaking change where we remove the non-oui prefixed classes.

danoc:
This PR does the following:

> ### Added
> - [Release] Add React to the OUI codebase. (#378)
> - [Release] Add Jest and the ESLint React plugin. (#378)
> - [Release] Add automatically generated React component documentation that gets deployed when new versions are released.
> - [Release] Create six React components: ArrowsInline, Attention, Button, Code, Input, and Label. (#378)
> - [Feature] Begin prefixing classes that React uses with the `oui-` prefix. (#361)
> 
> ### Changed
> - [Patch] Update `CONTRIBUTING.md` to include information about updating Optimizely's `package.json` after a release. (#408)
> - [Patch] Remove `devDependencies` before publishing OUI to NPM to increase speed of `npm install` for applications using OUI. (#356)
> - [Patch] Change release process guidelines to make upgrades less painful. (#360)
> 
> ### Fixed
> - [Patch] Add note to `README.md` saying that ScribeSass is internal only. (#359)
> - [Patch] Take a pass through our markdown files and update outdated information.

Got a 👍 on Slack and going to merge it once the tests pass.

PR #432, opened 2016-07-05
Add syntax highlighting option to React Code component

danoc:
This adds an `isHighlighted` prop to the Code component using a library called highlight.js. It also adds two tests, loads the third-party CSS file using webpack, and modifies Jest, our testing framework, to ignore the CSS file in tests.

The benefit of loading the CSS with React is that it only gets added to the page if there is a Code component that is asking for syntax highlighting.

@daverau-optimizely, @kwalker3690: Can you review?


daverau-optimizely:
LGTM — great work on this Dan, very fast turnaround.

PR #437, opened 2016-07-07
Create a Table React component.

danoc:
Screenshot of four tables:

screenshot 2016-07-07 11 25 27

@daverau-optimizely, @kwalker3690 – Can either of you review please?


src/components/Table/index.js
```diff
@@ -0,0 +1,90 @@
...
+    </table>
+  );
+};
+
+Table.propTypes = {
+  /** Array of data that appears within the table */
+  data: React.PropTypes.array.isRequired,
+  /** Headings for the `th` */
+  headings: React.PropTypes.array,
+  /** Availble border and hover options */
```
kwalker3690: Super nit: Available

danoc: Haha, good catch. Thanks!

src/index.jsx
```diff
@@ -239,6 +240,102 @@ var bat = 'baz';` }
...
+            <Table
+              data={
+                [
+                  ['Header CTA', '12%', 'Running'],
+                  ['Shorter Contact Form', '4%', 'Draft'],
+                  ['Larger search bar', '6.7%', 'Paused'],
+                  ['Center aligned headline', '9.3%', 'Running'],
+                ]
+              }
+              headings={
```
kwalker3690: I like how you've structured this data input. At first glance I was thinking the input would have looked more like:
{'Experiment': ['Header CTA', 'Shorter Contact Form' ...],
 'Conversion Rate': ['12%', '4%'...]
...}
But I think I like your version better, makes more sense with how the data is most likely already formatter.

Could the headings section be changed to just headings=['Experiment', 'Conversion Rate', 'Status']? It would align more with the layout of the data that corresponds to each column. The extra object wrapping seems unnecessary at first glance?

src/index.jsx
```diff
@@ -239,6 +240,102 @@ var bat = 'baz';` }
...
+                [
+                  {
+                    'text': 'Experiment',
+                  },
+                  {
+                    'text': 'Conversion Rate',
+                    'isNumerical': true,
```
kwalker3690: ohhh this is why it is an object?

src/index.jsx
kwalker3690: Alright I see what cleverness you've done here, using the isNumerical value for the corresponding table cells as well as the header, looks great!

src/components/Table/index.js
```diff
@@ -0,0 +1,90 @@
...
+    </thead>
+  );
+};
+
+const TableBody = (rows) => {
+  return (
+    <tbody>
+      { rows.map((row, i) => {
+        return (
+          <tr key={ i }>
```
kwalker3690: Do we need to support `oui-table-row--active`?

src/components/Table/index.js
```diff
@@ -0,0 +1,90 @@
...
+  );
+};
+
+Table.propTypes = {
+  /** Array of data that appears within the table */
+  data: React.PropTypes.array.isRequired,
+  /** Headings for the `th` */
+  headings: React.PropTypes.array,
+  /** Available border and hover options */
+  style: React.PropTypes.oneOf(['wall', 'rule']),
```
kwalker3690: What about `hover`?

src/components/Table/tests/index.js
```diff
@@ -0,0 +1,195 @@
...
+
+      expect(tableHeadRow[0].className).toBe('oui-cell-collapse');
+    });
+
+    it('should add numerical class when provided to a heading', () => {
+      const component = testHelpers.renderIntoDocument(TestTableWithHeadings);
+      const componentNode = testHelpers.getNodeFromComponent(component);
+      const tableHeadRow = componentNode.children[0].children[0].children;
+
+      expect(tableHeadRow[1].className).toBe('oui-numerical');
```
kwalker3690: Might be good to add a test for your clever handling of adding numerical class to `td` as well!

src/components/Table/tests/index.js
```diff
@@ -0,0 +1,195 @@
...
+    });
```
kwalker3690: Another useful test would be one that makes sure the right header is associated with the right column items, that is a piece that is a little tricky to understand at first because they are in two separate data structures, and I could see that getting misaligned fairly easily.

kwalker3690:
Couple of thoughts on tests and additional classes we currently support, but overall looks great!

kwalker3690:
Oh also, do you have any thoughts on how sorting will play into this component?

src/components/Table/index.js
danoc: Thought I'd leave that one out for now since it's a bit more complicated and not used often, I think. I can reintroduce later if we need it.

src/components/Table/index.js
danoc: Thought I'd leave that one out for now since it's a bit more complicated and not used often, I think. I can reintroduce later if we need it.

src/components/Table/tests/index.js
danoc: Good call! Adding it now.

src/components/Table/index.js
kwalker3690: Fair enough, yeah looks like its not used at all in the new product, we could consider deprecating that css then too if we dont want to support it as a feature

danoc: Yeah, I agree!

src/components/Table/tests/index.js
danoc: Good call! Done.

danoc:
@kwalker3690 – Going to just ignore the sorting for now and revisit when we need to implement it. 

Since we're keeping these components stateless, I think the parent component would handle the sorting logic and pass in an `onClick` prop for the headings that would update the order on the parent level and pass it down into this component.

src/components/Table/index.js
```diff
@@ -0,0 +1,101 @@
+import React from 'react';
+import { indexOf } from 'lodash';
+
+const cellClasses = (index, collapsedColumns, numberedColumns) => {
+  let classes = [];
+
+  if (indexOf(collapsedColumns, index) !== -1) {
```
kwalker3690: Just curious, why use lodash's indexOf instead of array.indexOf? Is there a difference in how those two are implemented?

danoc: Oh yeah, good call. Forgot that existed. Removed the dependency.

src/components/Table/index.js
```diff
@@ -0,0 +1,100 @@
...
+};
+
+Table.propTypes = {
+  /** Array of the column numbers, starting at 0, that should be collapsed */
+  collapsedColumns: React.PropTypes.array,
+  /** Array of data that appears within the table */
+  data: React.PropTypes.array.isRequired,
+  /** Headings for the `th` */
+  headings: React.PropTypes.array,
+  /**
```
kwalker3690: Nit: make comment style all the same here, and maybe move collapsedColumns and numberedColumns closer together since they're very similar

danoc: Both of those are because of a linter thing, but we can change the rule later if we want.

kwalker3690:
Looks awesome, one small nit but other than that LGTM!

src/components/Table/index.js
kwalker3690: Wait what's the linter problem?

danoc: There's a rule for alphabetizing the props and one for max line length of a comment

kwalker3690: Oh got it alright then nvm proceed with merging!

danoc:
Sweet, thanks Kelly!

PR #439, opened 2016-07-11
Stateless React Tabs

kwalker3690:
This is a work in progress diff, wanted to get some feedback on the style of creating these tabs before adding more examples and tests. @danoc what do you think about this way of creating the tabs? @jordangarcia would love insight from you here too if you have a chance, @danoc said you had some ideas about how to create stateless tabs.


danoc:
I'm gonna defer to @jordangarcia give feedback since I don't have much experience with this.

kwalker3690:
@williamklancnik Can you take a look at this?

src/components/Tabs/index.js
```diff
@@ -0,0 +1,67 @@
...
+  testSection: React.PropTypes.string,
+  style: React.PropTypes.arrayOf(React.PropTypes.oneOf([
+    'small',
+    'center',
+    'sub',
+  ])),
+  tabActive: React.PropTypes.number,
+};
+
+Tabs.MenuItem = (props) => {
```
jordangarcia: This doesn't seem very proper, to put additional components as properties of a component.

perhaps something like:

```
export { TabWrapper, Tab }
```

However I'm not sure, given the current state of transpilers, what that looks like in practice when importing.

Since we don't use import in the Optimizely codebase it may not work as intended.

This would be ideal if everywhere used es6 imports and we had all the components exported at top level.

```
import { TabWrapper, Tab } from 'optimizely-oui'
```

This may not be a bad idea (if it works)

```
const { TabWrapper, Tab } = require('optimizely-oui').Tabs
```

src/index.jsx
```diff
@@ -318,6 +319,25 @@ var bat = 'baz';` }
             </Table>
           </ComponentRow>
         </Component>
+
+        <Component title="Tabs">
+          <ComponentRow isPadded>
+            <Tabs activeTab={0}>
+              <Tabs.MenuItem isActive={this.props.activeTab === 0}
```
jordangarcia: Manually setting `isActive` and `activeTab` prop seems redundant.

Either we can do something magical with the `activeTab` + a `tabId` field

```
<TabWrapper activeTab={this.state.activeTab}>
  <Tab tabId="home">...</Tab>
  <Tab tabId="results">...</Tab>
</TabWrapper>
```

or just use `isActive` since the `TabWrapper` component seems really dumb (presentational only)

```
<TabWrapper activeTab={this.state.activeTab}>
  <Tab isActive={this.state.activeTab === 'home'}>...</Tab>
  <Tab isActive={this.state.activeTab === 'results'}>...</Tab>
</TabWrapper>
```

jordangarcia:
I think the imports / exports needs to be rethought, I definitely dont like the pattern of tacking on additional subcomponents to a component.

src/components/Tabs/index.js
mikeng13: `const { TabWrapper, Tab } = require('optimizely-oui').Tabs` does indeed work in our Optimizely Frontend codebase. Since this is open source though, there might be people using other transpilers, tools. I'd still vote for exposing it this way and figure out something else if other people complain

src/components/Tabs/index.js
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
+import classNames from 'classnames';
+
+const TabsMenu = (props) => {
+  let menuItems =  React.Children.map(props.children, (tab, index) => {
+    let isActive = props.activeTab === index;
+    let dataTestSection = tab.props.testSection ? tab.props.testSection : null;
+    let classes = classNames('tabs-nav__item', tab.props.isDisabled && 'tab-disabled', isActive && 'is-active');
```
mikeng13: this seems like something we can just do manually or write our own util function. I am just wary of pulling in external libraries which could negatively impact our bundle size

src/index.jsx
```diff
@@ -0,0 +1,346 @@
+/* eslint-disable react/jsx-filename-extension, react/prefer-stateless-function */
```
danoc: The changes to this file should be undone since it doesn't exist anymore! 

bitmoji

docs/index.jsx
```diff
@@ -23,6 +23,8 @@ import Label from '../src/components/Label';
 import LabelExample from '../src/components/Label/example';
 import Table from '../src/components/Table';
 import TableExample from '../src/components/Table/example';
+import Tabs from './components/Tabs';
```
danoc: Nit: All of the other components are singular. Might be best to keep consistent.

src/components/Tabs/index.js
```diff
@@ -0,0 +1,68 @@
...
+      className="tabs-nav"
+      data-test-section="tabs-menu">
+      {menuItems}
+    </ul>
+  );
+}
+
+const TabWrapper = (props) => {
+  let dataTestSection = props.testSection ? props.testSection : null;
+  let tabStyleClasses = props.style ? props.style.map((style) => {
```
danoc: This can be done using `classNames` thanks to ES6. See this example: https://github.com/JedWatson/classnames#dynamic-class-names-with-es2015

kwalker3690:
We've synced on Slack and in person and have agreed that the namespaced route is a good way to go, so I will revert this PR to the namespaced version, as well as make the other updates mentioned here.

jordangarcia:
lgtm

package.json
```diff
@@ -20,6 +20,8 @@
     "babel-jest": "^11.0.2",
     "babel-preset-es2015": "^6.6.0",
     "browser-sync": "^2.9.8",
+    "classnames": "^2.2.5",
+    "css-loader": "^0.23.1",
```
danoc: This was probably added by accident.

src/components/TabNav/index.js
```diff
@@ -0,0 +1,64 @@
...
+    'center',
+    'sub',
+  ])),
+  testSection: React.PropTypes.string,
+};
+
+const Tab = (props) => {
+  const isActive = props.activeTab === props.tabId;
+  const dataTestSection = props.testSection ? props.testSection : null;
+  const classes = classNames('oui-tabs-nav__item', props.isDisabled && 'oui-tab-disabled', isActive && 'is-active');
```
danoc: This is a little easier to read IMO:

```js
const classes = classNames({
  'oui-tabs-nav__item': true,
  'oui-tab-disabled': props.isDisabled,
  'is-active': isActive,
});
```

src/components/TabNav/index.js
```diff
@@ -0,0 +1,64 @@
...
+    'small',
+    'center',
+    'sub',
+  ])),
+  testSection: React.PropTypes.string,
+};
+
+const Tab = (props) => {
+  const isActive = props.activeTab === props.tabId;
+  const dataTestSection = props.testSection ? props.testSection : null;
```
danoc: This line doesn't really do anything since you can just do `{ props.testSection }` in the JSX and it will have the same effect.

src/components/TabNav/index.js
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
+import classNames from 'classnames';
+
+const TabNav = (props) => {
+  const dataTestSection = props.testSection ? props.testSection : null;
```
danoc: This line doesn't really do anything since you can just do { props.testSection } in the JSX and it will have the same effect.

src/components/TabNav/index.js
```diff
@@ -0,0 +1,64 @@
...
+  const tabStyleClasses = props.style ? props.style.map((style) => {
+    return 'oui-tabs--' + style;
+  }) : '';
+  const classes = classNames(tabStyleClasses, 'oui-tabs');
+  const childrenWithProps = React.Children.map(props.children, (child) => React.cloneElement(child, {
```
danoc: Can you add a comment explaining why this is needed?

src/components/TabNav/index.js
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
+import classNames from 'classnames';
+
+const TabNav = (props) => {
```
danoc: Can you add JSDocs to the two components and sets of props?

Example: https://github.com/optimizely/oui/blob/devel/src/components/Attention/index.js

package.json
```diff
@@ -20,6 +20,7 @@
     "babel-jest": "^11.0.2",
     "babel-preset-es2015": "^6.6.0",
     "browser-sync": "^2.9.8",
+    "classnames": "^2.2.5",
```
danoc: Needs to go under `dependencies`.

danoc:
Oh, also you need to export the new component in `src/main.js`.

danoc:
LGTM!!!! Thanks so much for doing this!

PR #449, opened 2016-07-13
Fix paths to static assets in React documentation

### LINKED ISSUE #444: Documentation should not expect that it is hosted at the root
We host the documentation here:
http://design.optimizely.com/docs/oui/13.0.0/react/

But the static files expect to live here:
http://design.optimizely.com/css
http://design.optimizely.com/js

I'm going to log into the server and manually fix for 13.0.0, but we should properly fix for the next release.

danoc:
This fixes #444.

cc: @mjc1283


mjc1283:
👍

PR #480, opened 2016-07-29
fix for disabled toggle buttons

daverau-optimizely:
1. add CSS rule to remove background and border on disabled hover
2. add more example buttons for testing

@kwalker3690 @danoc does this look good to you?


danoc:
Looks good to me!

bitmoji

PR #481, opened 2016-08-01
Include an example of a Table React component with a Button within it.

danoc:
@daverau-optimizely – Can you review? I think it's nice to see that the table can accept more than just strings. This is how it looks:

screenshot 2016-08-01 13 43 35

cc: @spencerwilson-optimizely


daverau-optimizely:
Great idea Dan! Really like seeing these examples grow.

PR #496, opened 2016-08-12
Add webpack loaders into `require(...)` and as `dependencies`

### LINKED ISSUE #485: Repositories that use OUI React need webpack CSS loader
Got this error while trying to use it in a project.

```
ERROR in ./~/optimizely-oui/~/highlight.js/styles/docco.css
Module parse failed: /Users/doconnor/bunji/node_modules/optimizely-oui/node_modules/highlight.js/styles/docco.css Unexpected token (5:0)
You may need an appropriate loader to handle this file type.
```

danoc:
**Problem:** OUI's React components require SVG and CSS files. These can only be loaded and used in JavaScript with tools such as webpack loaders. We currently define these loaders in a `webpack.config.js` file, but the loaders are `devDependencies`, not `dependencies`, and applications that use OUI React aren't able to inherit our webpack config.

This means that all applications using OUI need to manually install webpack loaders for CSS, SVGs, and other filetypes that we may load in the future. They must also update their webpack config to support these file types.

([Example of this complicated setup](https://github.com/jampy/kendo-ui-webpack/tree/bbb1523d460821366eeeaa9ebdd9173ddebadfe7#installing))

**Solution:** Move the loader information into the `require(...)`'s and add the loaders as `dependencies` within the OUI `package.json`.

This has the downside of coupling applications that use OUI's React components to webpack. Although not ideal, I think it's acceptable considering OUI's target audience. We can revisit this in the future if that becomes a problem.

**How to review:** I split the commits into two. The first is useful info, the second is auto-generated updates to our icon components. The rest are attempts to fix lints and tests.

***
This fixes #485.

Related discussion: https://github.com/webpack/webpack/issues/378


daverau-optimizely:
@danoc seems like a very reasonable tradeoff given the benefits. Do we know if anyone is actually using OUI locally to build from source vs just using the compiled CSS? I would imagine nobody else is compiling, so we're probably safe going this route without additional support costs.

👍  💯

danoc:
This seems interesting:
https://github.com/istarkov/babel-plugin-webpack-loaders

mjc1283:
Makes sense, LGTM.

PR #499, opened 2016-08-17
fix oui v14 webpack errors

### LINKED ISSUE #498: OUI v14 breaks webpack in app
Just doing a simple `npm i` for OUI v14 throws webpack errors

[After](https://github.com/webpack/css-loader/issues/240#issuecomment-231507670) [much](http://webpack.github.io/docs/using-loaders.html#usage) [investigation](https://github.com/webpack/css-loader/issues/295#issuecomment-231510027) it appears our `require` statements should be prefixed with `!` to ignore local webpack config, so both the Optimizely app webpack config and OUI `require` can coexist and not compete.

@mjc1283 discovered we can exclude OUI icons from the svg-inline loader in the frontend webpack config:
```
      {
        test: /\.svg$/,
        loader: 'svg-inline',
        exclude: [
          path.resolve(__dirname, '../node_modules/oui-icons/src/'),
        ],
      },
```

daverau-optimizely:
Fixes #498 

@mjc1283 can you review and verify this works for you locally?

cc @danoc


mjc1283:
👍  Verified that this fixes errors in optimizely.

PR #508, opened 2016-08-25
Drau/input isfilter

daverau-optimizely:
@danoc this:

1. adds the isFilter prop
2. applies a newly namespaced . oui-text-input--search
3. moves that CSS out of the parent .search container since there's no positioning on it

Here's the example in the doc:
isfilter prop


danoc:
LGTM!!!!

PR #516, opened 2016-08-26
sync changelog with master

daverau-optimizely:
And some cleanup with all the messy commits on my part. I'll try to be more aware and careful in commits and changelog updates.

@danoc


danoc:
No longer needed.

PR #521, opened 2016-08-26
move svgo from inline to bash

daverau-optimizely:
@danoc 

I believe this is done and ready to go. We'd have to release this and then update in the app to see if that fixes the Jenkins issue.


danoc:
LGTM!

PR #523, opened 2016-08-27
Set user-select: none on oui-button

spencerwilson-optimizely:
I hope this makes it so `Button` child `span`s can't be highlighted, but I haven't tested it.


danoc:
This looks good to me, but needs an entry in the `CHANGELOG.md`.

Example:
```
### Changed
- Prevent `button` text from being selected with `user-select: none`.
```

daverau-optimizely:
Works locally for me — @spencerwilson-optimizely would you mind making the changelog commit? Then it's good to go!

danoc:
LGTM once this passes.

PR #524, opened 2016-08-30
Add a `Table` style called `rule-no-bottom-border`

danoc:
@daverau-optimizely – Can you review this? I decided to make a style called `rule-no-bottom-border` instead of adding a prop called `hasBottomBorder` because it would end up making the JS logic more complicated with little benefit to the user.

cc: @spencerwilson-optimizely @mikeng13


mikeng13:
LGTM

daverau-optimizely:
LGTM

PR #526, opened 2016-08-30
Add the ability to pass a `width` to `Table.TD` and `Table.TH`

danoc:
@daverau-optimizely and @spencerwilson-optimizely – Can you review please?


src/components/Table/TH/index.js
```diff
@@ -7,10 +7,15 @@ const TH = (props) => {
     'oui-cell-collapse': props.isCollapsed,
   });
 
+  const styles = {
+    width: props.width,
```
daverau-optimizely: I really like this pattern, we should probably switch everything over to this styles object.

daverau-optimizely:
LGTM!

PR #527, opened 2016-08-30
Make `Button` `children` prop more generic.

### LINKED ISSUE #507: Button.propTypes.children is too strict
Button currently only accepts string children. It would be convenient be able to pass it multiple children (sample use case: <span>Label</span> <ArrowsInline direction="down" />).

danoc:
Fixes #507.

@spencerwilson-optimizely – Can you review this one?


spencerwilson-optimizely:
LGTM!

PR #530, opened 2016-08-30
fix buttonrow key warnings

daverau-optimizely:
This squashes the ButtonRow warnings in the docs @danoc


danoc:
Looks great. Thanks!

PR #532, opened 2016-08-30
Drau/copy icon

daverau-optimizely:
@danoc this is failing — crashing Chrome, works in Firefox. [Line 22](https://github.com/optimizely/oui/blob/6b42116f8ab82b17b47ac5daa0bb8c7f29a8f5f8/src/components/Code/index.js#L22) in this file is the cause, is there a cleaner way to fire the copy event without duplicating the element?


danoc:
I ran into a bunch of the same problems when I took a stab at this.

This is where I left off:
https://github.com/optimizely/oui/compare/daniel/copy

I forget what state it's in though.

daverau-optimizely:
Abandoning this — not a priority right now.

PR #538, opened 2016-09-01
Drau/revert tabnav examples

daverau-optimizely:
Revert TabNav examples to original version, even though it's more code and scrolling in the docs, those props are required and reasoning about the component more clear.

cc @danoc


danoc:
LGTM!

PR #541, opened 2016-09-02
Add a `density` prop to `Table` with a `loose` option.

### LINKED ISSUE #536: Create a Table style with more padding
The tables in our product often have more spacing in the `<td>`'s. I think we should move the product overrides into OUI and provide a prop for `Table` and modifier for `_table.scss`.

cc: @mikeng13 who requested this.

danoc:
This fixes #536.

Here's a screenshot:
screenshot 2016-09-02 10 10 13


daverau-optimizely:
Looks `tight` to me.

PR #545, opened 2016-09-02
Remove `text-align: center` from `[colspan]`.

### LINKED ISSUE #544: Remove existing center text alignment from table cells that span multiple columns
Our SCSS currently center aligns all `td`'s that have a `colspan` that isn't equal to `1`.

This isn't always desired and has forced developers to use overrides, so I'd like to change the default behavior to be left-aligned like the other table cells.

*** 
cc: @daverau-optimizely, @mjc1283 (who stumbled upon this)

danoc:
This fixes #544.

It is a breaking change and requires adding `text--center` to table cells that use colspan and don't have an override. It also allows us to remove the overrides.


daverau-optimizely:
👍

PR #547, opened 2016-09-07
Add a `tableLayoutAlgorithm` prop to the `Table` component.

### LINKED ISSUE #546: Table layout should be fixed by default
See [here](https://css-tricks.com/fixing-tables-long-strings/) and [here](https://developer.mozilla.org/en-US/docs/Web/CSS/table-layout).

danoc:
This PR adds a `tableLayoutAlgorithm` prop to the `Table` component and sets it to `fixed` by default. Existing uses of the component should be updated to either ensure they look correct with the new default or add `tableLayoutAlgorithm={ 'auto' }` to the instance of `Table`.

The prop is needed because text within cells can't wrap when `table-layout` is set to `auto`. This becomes a breaking change because the default is now different than the expected browser default, but it is [a lot more sensible for a web app](https://css-tricks.com/fixing-tables-long-strings/) IMO.

Fixes #546.


daverau-optimizely:
👍

spencerwilson-optimizely:
👍

danoc:
Merging because it passed locally.

Issue #550, opened 2016-09-08
Convert icons bash script to Node.js


danoc:
I went down a bash rabbit hole while trying to make an update to the icons bash script and decided to convert it to Node.js for ease of updating.

danoc:
Done in #551.

PR #551, opened 2016-09-08
Convert `sync-oui-icons` bash script to Node.js.

danoc:
A few things in this PR:

- [Patch] Convert `sync-oui-icons` bash script to Node.js. (#550)
- [Patch] Add a default case to each icon component so that a non-standard size can be passed.
- [Patch] Remove `svgo` since it was only minifying the SVGs locally and not in the published version of OUI.


daverau-optimizely:
👍 

Thanks for cleaning this up, Dan, all very good changes.

PR #564, opened 2016-09-09
Empty code component renders nothing

### LINKED ISSUE #557: Code component should handle children with null/empty state values

daverau-optimizely:
1. Loosen children props to be optional
2. Add test to check code element is null when no children are supplied to Code component.

Fixes #557


daverau-optimizely:
cc @danoc

danoc:
Ran tests locally.

Issue #515, opened 2016-08-25
Clarify OUI release process in contributing.md


daverau-optimizely:
1. Add any extra details to the instructions
2. Update script to only push current tag
3. Investigate Travis success release, URL

danoc:
Going to close this because we already did a decent amount of it.

PR #567, opened 2016-09-12
Use `git push --follow-tags` when releasing a new version

danoc:
Related to #515.

See this StackOverflow answer for more information:
http://stackoverflow.com/a/3745250/316602


daverau-optimizely:
👍

Issue #555, opened 2016-09-08
Input component doesn't use correct label class


haleybash-optimizely:
https://github.com/optimizely/oui/blob/devel/src/components/Input/index.js#L35

Should be `oui-label`

daverau-optimizely:
Form label styles are inconsistent — this change will make them more consistent for all new React-powered forms.

Issue #556, opened 2016-09-08
Add `isOptional` prop to input components


haleybash-optimizely:

daverau-optimizely:
This would add optional labels for text input form fields that use React.

daverau-optimizely:
This would add optional labels for text input form fields that use React.

PR #569, opened 2016-09-12
Final pass at Label isOptional/isRequired

daverau-optimizely:
@danoc What do you think about this fixup?
1. Should we include more robust tests for the `isOptional` prop?
2. Do we need code to prevent `isRequired` combined with `isOptional`?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ This file follows the format suggested by [Keep a CHANGELOG](https://github.com/
 
 ## [Unreleased][Unreleased]
 ### Added
-- [Feature] Add `isOptional` prop to Label React component to get properly colored optional/required labels. (#556)
+- [Feature] Add `isOptional` and `isRequired` prop to Label React component to get properly colored optional/required labels. (#556)
 
 ### Changed
 - [Patch] Use correct oui- prefix with label. (#555)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ This file follows the format suggested by [Keep a CHANGELOG](https://github.com/
 
 ## [Unreleased][Unreleased]
 ### Added
-- [Feature] Add `isOptional` prop to Input React component to get optional label for free. (#556)
+- [Feature] Add `isOptional` prop to Label React component to get properly colored optional/required labels. (#556)
 
 ### Changed
 - [Patch] Use correct oui- prefix with label. (#555)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This file follows the format suggested by [Keep a CHANGELOG](https://github.com/olivierlacan/keep-a-changelog).
 
 ## [Unreleased][Unreleased]
+### Added
+- [Feature] Add `isOptional` prop to Input React component to get optional label for free. (#556)
+
+### Changed
+- [Patch] Use correct oui- prefix with label. (#555)
 
 ## [17.1.0][17.1.0] - 2016-09-12
 ### Added
--- a/src/components/Code/CopyButton/index.js
+++ b/src/components/Code/CopyButton/index.js
@@ -27,7 +27,7 @@ class CopyButton extends React.Component {
           style="plain"
           ariaLabel="Copy code snippet"
           testSection={ buttonTestSection }>
-          <ClipboardIcon />
+          <ClipboardIcon size={ 16 } />
         </Button>
       </div>
       /* eslint-enable */
--- a/src/components/Label/index.js
+++ b/src/components/Label/index.js
@@ -9,10 +9,10 @@ const Label = (props) => {
   let classes = null;
   let fieldLabel = null;
 
-  if (props.isOptional) {
-    fieldLabel = <span className="label__optional">(Optional)</span>;
-  } else if (props.isRequired) {
+  if (props.isRequired) {
     fieldLabel = <span className="label--required"></span>;
+  } else if (props.isOptional) {
+    fieldLabel = <span className="label__optional">(Optional)</span>;
   }
 
   if (typeof props.children === 'string') {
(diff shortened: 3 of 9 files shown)


danoc:
Nice! Let's do some brainstorming though and see how we can make this work for all labels, not just `Input`.

A future `Textarea` component will have the same need, but we wouldn't want to duplicate this in two places.

daverau-optimizely:
I'll rework this

1. break out isOptional to the Label component
2. Input should reference Label (pass a component not a string)

daverau-optimizely:
@danoc this is ready for you again, thanks!

Here's the docs output:
react label docs

src/components/Label/tests/index.js
```diff
@@ -40,6 +40,24 @@ describe('components/Label', () => {
     });
   });
 
+  it('should render an optional label if isOptional is set', () => {
+    const component = render(
+      <Label
+        testSection="foo"
+        isOptional={ true }></Label>
+    );
+    expect(component.find('[data-test-section="foo"] span.label__optional').length).toBe(1);
```
danoc: I think it's better do just `.label__optional` because the `span` element is an implementation detail, and the test section isn't helping us since it is already the outermost element.

src/components/Label/tests/index.js
```diff
@@ -40,6 +40,24 @@ describe('components/Label', () => {
     });
   });
 
+  it('should render an optional label if isOptional is set', () => {
+    const component = render(
```
danoc: I'm pretty sure we can do `shallow` instead of render. A bit lighter weight and faster.
https://github.com/airbnb/enzyme/blob/master/docs/api/shallow.md

src/components/Label/tests/index.js
```diff
@@ -40,6 +40,24 @@ describe('components/Label', () => {
...
+      <Label
+        testSection="foo"
+        isOptional={ true }></Label>
+    );
+    expect(component.find('[data-test-section="foo"] span.label__optional').length).toBe(1);
+  });
+
+  it('should render an asterisk label if isRequired is set', () => {
+    const component = render(
```
danoc: Same here about `render` vs `shallow`.

src/components/Label/index.js
```diff
@@ -7,6 +7,14 @@ import React from 'react';
  */
 const Label = (props) => {
   let classes = null;
+  let fieldLabel = null;
+
+  if (props.isOptional) {
```
danoc: We should have a test that ensures required is not rendered if both `isRequired` and `isOptional` are passed.

CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This file follows the format suggested by [Keep a CHANGELOG](https://github.com/olivierlacan/keep-a-changelog).
 
 ## [Unreleased][Unreleased]
+### Added
+- [Feature] Add `isOptional` prop to Label React component to get properly colored optional/required labels. (#556)
```
danoc: Should mention that it also adds the `isRequired` prop.

src/components/Label/tests/index.js
daverau-optimizely: Good call, switched both these to `shallow`

src/components/Label/tests/index.js
daverau-optimizely: Nice. I dropped the `test-section` and opted for a simpler setup like you suggested, good call Dan!

src/components/Label/tests/index.js
daverau-optimizely: Done

src/components/Label/index.js
daverau-optimizely: Added this and flipped the order of the `if` statements so isRequired is checked first and given preference.

CHANGELOG.md
daverau-optimizely: Keen eyes! Sorry for being so sloppy here.

daverau-optimizely:
Thanks @danoc this is ready for a final pass. Very good nits on your part, thanks so much for always taking the time to do a thoughtful review.

src/components/Code/CopyButton/index.js
```diff
@@ -27,7 +27,7 @@ class CopyButton extends React.Component {
           style="plain"
           ariaLabel="Copy code snippet"
           testSection={ buttonTestSection }>
-          <ClipboardIcon />
+          <ClipboardIcon size={ 16 } />
```
daverau-optimizely: Added this to squash jest test warnings.

danoc: Nice, thanks!

daverau-optimizely:
@danoc Ok just updated the extra span, but I'm not sure I've done it any cleaner than before:
```
const Label = (props) => {
  let classes = null;
  let labelEl = null;

  // Optional/Required additional span label
  if (props.isRequired || props.isOptional) {
    let labelClass = null;
    let labelText = null;

    if (props.isRequired) {
      labelClass = 'label--required';
    } else if (props.isOptional) {
      labelClass = 'label__optional';
      labelText = '(Optional)';
    }
    labelEl = <span className={ labelClass }>{ labelText }</span>;
  }

  if (typeof props.children === 'string') {
    classes = 'oui-label';
  }

  return (
    <label
      className={ classes }
      data-test-section={ props.testSection }>
      { props.children }
      { labelEl }
    </label>
  );
};
```

src/components/Label/index.js
```diff
@@ -7,6 +7,14 @@ import React from 'react';
  */
 const Label = (props) => {
   let classes = null;
+  let fieldLabel = null;
+
+  if (props.isRequired) {
+    fieldLabel = <span className="oui-label--required"></span>;
```
danoc: I think you also have to update the SCSS.

(Looks like the automated tests caught it 😄)

daverau-optimizely:
Fixed!

Maybe we should have a git pre push hook to run tests before pushing. I'll check that out.

daverau-optimizely:
Fixes #555 and #556

PR #573, opened 2016-09-15
Disable Copy Clipboard 

daverau-optimizely:
@danoc I'm tried to get this working on my own and got stuck on a few `@QUESTION`s I had. Maybe we can pair on this a little so I can better understand the function flow back and forth.

Also, is there a better way to code review without making a pull request? That implies it's ready to go, but this isn't...

### DIFF
--- a/src/components/Code/tests/index.js
+++ b/src/components/Code/tests/index.js
@@ -102,12 +102,12 @@ describe('components/Code', () => {
   });
 
   describe('with copy functionality', () => {
+    beforeEach(() => {
+      global.document.queryCommandSupported = () => true;
+    });
+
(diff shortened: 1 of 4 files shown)


src/components/Code/tests/index.js
```diff
@@ -102,13 +102,13 @@ describe('components/Code', () => {
   });
 
   describe('with copy functionality', () => {
+    beforeEach(() => {
+      global.document.queryCommandSupported = () => true;
```
mjc1283: Super nit picking here: you might want to undo this in `afterEach`, even though it's so unlikely to cause an issue in another test.

PR #576, opened 2016-09-16
Adding isActive prop to Button component

kwalker3690:
@danoc

### DIFF
--- a/src/components/Button/tests/index.js
+++ b/src/components/Button/tests/index.js
@@ -31,6 +31,6 @@ describe('components/Button', () => {
       <Button isActive={ true }>Hello!</Button>
     );
 
-    expect(component.find('.is-active').length).toBe(1);
+    expect(component.hasClass('is-active')).toBe(true);
   });
 });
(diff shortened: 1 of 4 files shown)


src/components/Button/tests/index.js
```diff
@@ -25,4 +25,12 @@ describe('components/Button', () => {
 
     expect(component.is('[data-test-section="foo"]')).toBe(true);
   });
+
+  it('should add the active class if isActive is true', () => {
+    const component = shallow(
+      <Button isActive={ true }>Hello!</Button>
+    );
+
+    expect(component.find('.is-active').length).toBe(1);
```
danoc: Do either of these work?

```
// https://github.com/airbnb/enzyme/blob/master/docs/api/ShallowWrapper/hasClass.md
component.hasClass('is-active')
// https://github.com/airbnb/enzyme/blob/master/docs/api/ShallowWrapper/isEmpty.md
component.is('.is-active')
```

Super nit-picky, but it's a little nicer than doing a `find` since it doesn't look at `{ children}`. Feel free to leave as-is if none of those work.

PR #589, opened 2016-09-21
Update order of `npm test` tasks so that tests most likely to fail run first

### LINKED ISSUE #588: Change order of `npm test` tasks so most likely to fail ones are run first
ESLint and Jest should probably run earlier since those are the most likely to fail and break builds. It will lead to quicker feedback when there are problems.

danoc:
This fixes #588.


daverau-optimizely:
Nice little improvement. 👍

PR #593, opened 2016-09-26
Adding optional prop to input

kwalker3690:


src/components/Input/index.js
```diff
@@ -53,6 +54,16 @@ Input.propTypes = {
   isDisabled: React.PropTypes.bool,
   /** Includes search icon if true */
   isFilter: React.PropTypes.bool,
+  /** Adds an optional label if there is a label provided
+  *   @param {Object} props Object of props
```
danoc: Nit: The `*` should have one more space and be aligned like this:

screenshot 2016-09-26 15 37 43

src/components/Input/index.js
```diff
@@ -53,6 +54,16 @@ Input.propTypes = {
...
+  *   @returns {Error} Error or null
+  */
+  isOptional: function verifyIsOptionalProp(props) {
+    if (props.isOptional && !props.label) {
```
danoc: If we add logic here I think we should add a test to ensure that the error is thrown when expected.

src/components/Input/tests/index.js
```diff
@@ -140,6 +140,17 @@ describe('components/Input', () => {
...
   });
 
+  it('should throw an error if isOptional is passed without a label', () => {
+    spyOn(console, 'error').and.stub();
+
+    const component = mount(
+      <Input type="text" testSection="foo" isOptional={ true } />
+    );
+
+    expect(console.error).toHaveBeenCalled(); // eslint-disable-line
```
danoc: It'd be nice if this checks the arguments of `console.error` to ensure that the relevant one is being thrown.

danoc:
Tests passed locally. Merging now.

PR #595, opened 2016-09-26
Add `blur` method to `Input` and allow it to be referenced with `ref`

### LINKED ISSUE #591: Make input component not a pure function so that it can be used with ref

danoc:
This involves converting the pure function to an ES2015 class.

This was needed because @kwalker3690 wanted to add an event listener in a parent component that would blur the input whenever <kbd>ESC</kbd> was pressed. From what I understand, it's impossible to do that with our current setup because you can't use `ref` on a component that is a pure function.

We fix that by converting `Input` into a ES2015 class, but there is still an additional problem: adding `ref` to a component returns the component's instance. This means there is no way to access the underlying `<input>` HTML element and call `blur` on it. From the React documentation: 

> When attaching a ref to a DOM component like `<div />`, you get the DOM node back; when attaching a ref to a composite component like `<TextInput />`, you'll get the React class instance. In the latter case, you can call methods on that component if any are exposed in its class definition.

We can fix _that_ by adding a `blur` method on `Input`.

***
Fixes #591


danoc:
Tests passed locally, so merging now.

PR #596, opened 2016-09-26
Adding checkbox react component

kwalker3690:
Also added nifty new jest watch command


src/components/Checkbox/tests/index.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import Checkbox from 'components/Checkbox';
+import { mount, render } from 'enzyme';
+
+describe('components/Input', () => {
```
danoc: `components/Checkbox`

src/components/Checkbox/tests/index.js
```diff
@@ -0,0 +1,40 @@
...
+  it('should output an error if created without an "onChange" handler', () => {
+    spyOn(console, 'error').and.stub();
+
+    render(
+      <Checkbox label="foo" isChecked={ true } />
+    );
+
+    expect(console.error).toHaveBeenCalled(); // eslint-disable-line
```
danoc: Can we check to make sure the arguments indicate that the error is related to `onChange`? Otherwise this can pass if we make another prop required but omit it in the example.

src/components/Checkbox/tests/index.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import Checkbox from 'components/Checkbox';
+import { mount, render } from 'enzyme';
+
+describe('components/Input', () => {
+  it('should output an error if created without an "onChange" handler', () => {
+    spyOn(console, 'error').and.stub();
+
+    render(
```
danoc: Does the error only appear with `render`? I thought these things worked with mount and shallow too.

src/components/Checkbox/index.js
```diff
@@ -0,0 +1,50 @@
...
+        </Label>
+      </li>
+    </ul>
+  );
+};
+
+Checkbox.propTypes = {
+  /** Whether the checkbox is checked or not */
+  isChecked: React.PropTypes.bool.isRequired,
+  /** Prevents input from being modified and appears disabled */
```
danoc: Does `isDisabled` make the label look disabled?

src/components/Checkbox/index.js
```diff
@@ -0,0 +1,50 @@
...
+  isDisabled: React.PropTypes.bool,
+  /** Prevents input from being modified but doesn't appear disabled */
+  isReadOnly: React.PropTypes.bool,
```
danoc: Does this work on checkboxes?

src/components/Checkbox/index.js
```diff
@@ -0,0 +1,50 @@
...
+  /** Prevents input from being submitted without value */
+  isRequired: React.PropTypes.bool,
```
danoc: I don't think this applies to checkboxes.

src/components/Checkbox/index.js
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import Label from 'components/Label';
+
+/**
+ * Generates a `checkbox` element (optionally wrapped in a label) and accepts
```
danoc: The description needs to be updated.

src/components/Checkbox/tests/index.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
```
danoc: We should have a test that ensures the checkbox is surrounded by a Label (example: click on label, expect checkbox to now be checked) and the text passed to label is displayed.

kwalker3690:
Changes look good to me, I think we can add a value prop later if we'd like to have the option to have the component be controlled, as it is uncontrolled right now, but I think for this first pass its not absolutely necessary

PR #604, opened 2016-09-30
Added checked prop to Checkbox react component

kwalker3690:


src/components/Checkbox/tests/index.js
```diff
@@ -18,6 +18,24 @@ describe('components/Checkbox', () => {
...
+
+  it('should render an unchecked checkbox when `checked` is false', () => {
+    const component = mount(<Checkbox checked={ false } onChange={ function() {} } />);
+    expect(component.find('input[type="checkbox"]').get(0).checked).toBe(false);
+  });
+
+  it('should throw an error if a checkbox is rendered without an onChange function', () => {
+    spyOn(console, 'error').and.stub();
+
+    mount(<Checkbox checked={ true } />);
```
danoc: This should work with `shallow`, I think. (It's a bit faster)

PR #611, opened 2016-10-04
Performance and positioning improvement to Popover.

danoc:
Includes two changes:
 - [Patch] Call `Tether.disable()` when the overlay in `OverlayWrapper` is not visible. This leads to performance increases because Tether will essentially ignore the element until it becomes enabled.
 - [Patch] Reposition the Tether overlay when it is opened. This is needed because Tether can't properly calculate the position of elements that are hidden with `display: none;`.


src/components/OverlayWrapper/index.js
```diff
@@ -28,21 +28,45 @@ class OverlayWrapper extends React.Component {
...
       return React.cloneElement(child, {
         onClick: (event) => {
+          const newIsOverlayOpen = !this.state.isOverlayOpen;
+
           this.setState({
-            'isOverlayOpen': !this.state.isOverlayOpen,
+            'isOverlayOpen': newIsOverlayOpen,
           });
 
+          // Disable Tether when not visible for performance reasons.
```
kwalker3690: Move this comment to line 65 and add a comment here about enabling it since you've started it as disabled

src/oui/library/tether/1.3.7/_index.scss
```diff
@@ -4,96 +4,120 @@ $margin: 10px;
   &.tether-element-attached-middle.tether-element-attached-right.tether-target-attached-middle.tether-target-attached-left {
     margin-left: -$margin;
 
-    .oui-pop--over__content {
-      @include arrow(right, center, map-fetch($color, ui, white), map-fetch($color, ui, base));
+    @include arrow(right, center, map-fetch($color, ui, white), map-fetch($color, ui, base));
+
+    .oui-pop--over {
+      margin-right: -1px;
```
kwalker3690: Could be nice to have this as a constant at the top of this file, not a big deal tho.

src/components/OverlayWrapper/tests/index.js
```diff
@@ -51,65 +54,133 @@ describe('components/OverlayWrapper', () => {
...
-    // Ensure that the default state is what this test expects
-    expect(component.state('isOverlayOpen')).toBe(false);
-    // Ensure that we are looking at the correct call to `React.cloneElement`
-    expect(fakeButtonClone.args[0].type).toBe(FakeButton);
+      fakeButtonClone = React.cloneElement.calls.mostRecent();
 
-    // Simulate a click on the button
-    fakeButtonClone.args[1].onClick();
+      // Ensure that we are looking at the correct call to `React.cloneElement`
+      expect(fakeButtonClone.args[0].type).toBe(FakeButton);
```
kwalker3690: Since this is a new pattern, to have an expect in the beforeEach, make sure this fails if the expect is wrong. If it does, lgtm

src/components/OverlayWrapper/tests/index.js
```diff
@@ -51,65 +54,133 @@ describe('components/OverlayWrapper', () => {
...
+      fakeButtonClone.args[1].onClick();
 
-    const fakeButtonClone = React.cloneElement.calls.mostRecent();
+      // Tether should now be disabled
+      expect(disableMock.calls.length).toBe(initialDisableMockCalls + 1);
+    });
+
+    it('should position Tether when clicking on the children', () => {
+      // Ensure that we are looking at the correct call to `React.cloneElement`
+      expect(fakeButtonClone.args[0].type).toBe(FakeButton);
```
kwalker3690: If you have this in the beforeEach and it works there, you don't need to repeat it here

src/components/OverlayWrapper/tests/index.js
```diff
@@ -51,65 +54,133 @@ describe('components/OverlayWrapper', () => {
...
+
+      // The `Tether.position` function should have been called once
+      expect(positionMock.calls.length).toBe(initialPositionMockCalls + 1);
+    });
+
+    it('should pass `onClick` function to `children` that changes state of `OverylayWrapper`', () => {
+      // Ensure that the default state is what this test expects
+      expect(component.state('isOverlayOpen')).toBe(false);
+      // Ensure that we are looking at the correct call to `React.cloneElement`
+      expect(fakeButtonClone.args[0].type).toBe(FakeButton);
```
kwalker3690: Ditto

PR #620, opened 2016-10-07
Button full size prop

### LINKED ISSUE #615: Button: need full-width option to be separate from full
Talked to @danoc about this - I need to create a Button that has full width with respect to its container, but is still 'small'-sized.

daverau-optimizely:
Fixes #615 

cc @danoc


src/components/Button/index.js
```diff
@@ -9,11 +9,12 @@ import React from 'react';
 const Button = (props) => {
   let buttonStyleClass = props.style ? 'oui-button--' + props.style : '';
   let buttonSizeClass = props.size ? 'oui-button--' + props.size : '';
+  let buttonWidthClass = !props.width ? props.width : 'oui-button--full';
```
danoc: This would be a good opportunity to refactor this stuff to use `classNames`:

https://github.com/JedWatson/classnames
https://github.com/optimizely/oui/blob/9493764b668b333fa201212e9383da4ad4518320/src/components/Table/index.js#L16-L20

src/components/Button/index.js
```diff
@@ -59,6 +59,11 @@ Button.propTypes = {
   ]),
   /** Hook for automated JavaScript tests */
   testSection: React.PropTypes.string,
+  /** Various height and width options */
+  width: React.PropTypes.oneOf([
```
danoc: Maybe add a `defaultProps` to this component?

```
Button.defaultProps = {
  width: 'default',
};

danoc:
Merging since tests passed locally.

PR #628, opened 2016-10-07
Remove TimePicki library SCSS.

### LINKED ISSUE #627: Remove TimePicki library code from SCSS
This is a breaking change but we don't need the code anymore because we're not using it anywhere.

danoc:
This fixes #627.

***
@asaschachar or @daverau-optimizely, can you review?


danoc:
Ran tests locally and they passed. Going to merge.

PR #630, opened 2016-10-08
Improve test coverage for `Button` React component.

### LINKED ISSUE #629: Add more tests to the `Button` component
We currently only have three tests but there is a lot more we can be testing since the component is so crucial and has a lot of functionality.

danoc:
Here's the new set of tests:

```
 PASS  src/components/Button/tests/index.js
  components/Button
    ✓ should render a `button` HTML element (29ms)
    ✓ should render contents that are passed in (5ms)
    ✓ should call function that is passed in as `onClick` after click (4ms)
    ✓ should add an `aria-label` when provided (3ms)
    ✓ should add the active class if `isActive` is true (3ms)
    ✓ should add the disabled attribute if `isDisabled` is true (1ms)
    ✓ should be of type `button` when `isSubmit` is false (3ms)
    ✓ should be of type `button` when `isSubmit` is not provided (2ms)
    ✓ should be of type `submit` when `isSubmit` is true (2ms)
    ✓ should have a properly set size class if `size` is provided (4ms)
    ✓ should have a properly set style class if `style` is provided (1ms)
    ✓ should have a properly set test section (2ms)
    ✓ should have a properly set width class if `width` is provided (2ms)
```

*** 
Fixes #629 
*** 
@daverau-optimizely – Can you review?


danoc:
Going to merge since the tests passed locally.

Issue #495, opened 2016-08-12
Make webpack builds faster for React documenation


danoc:
The solution is to do one of the three things listed at the top of this page:
https://webpack.github.io/docs/build-performance.html

PR #631, opened 2016-10-10
Speed up local React JavaScript compilation.

danoc:
This greatly improves the speed in which JavaScript recompiles when watching files in the OUI React documentation. 

I've also cleaned up the NPM scripts a bit and renamed `npm run react-docs:serve` to `npm run react:watch`.

The one caveat of this it doesn't automatically refresh the page when Sass files change (they get recompiled, but you have to refresh manually.) Going to look into ways to fix this.

Fixes: #495

***
**Update:** Sass files are now watched and recompile on changes.
***
@daverau-optimizely – Can you review?


daverau-optimizely:
From `4006ms` to `360ms` initial compile and `261ms` recompile! Very fast 💯 👍

Issue #647, opened 2016-10-24
Consider adding a `poptip` prop to `Button`, `Icon` and other components


danoc:
This would make it easier for developers to add poptips to commonly used components.

danoc:
Going to ignore this for now because the current solution works and this is more of a nice-to-have.

Issue #646, opened 2016-10-24
Refactor `Popover` component to use `OverlayWrapper`


danoc:
The `Popover` component currently does not handle positioning. It leaves that up to the developer that may choose to use `OverlayWrapper`. 

This is an example:

``` html
<OverlayWrapper
  behavior="click"
  horizontalAttachment="center"
  overlay={<Popover title="Lorem ipsum dolor sit amet"><p>Ipsa officiis bad-news minus earum a aperiam! Aperiam reiciendis vitae nihil libero et, hic!</p></Popover>}
  verticalAttachment="top"
>
  <Button width="default">
    Open Popover
  </Button>
</OverlayWrapper>
```

It'd be a lot easier for the user if the API looked like this:

``` html
<Popover 
  title="Lorem ipsum dolor sit amet"
  content=(
    <p>Ipsa officiis bad-news minus earum a aperiam! Aperiam reiciendis vitae nihil libero et, hic</p>
  )>
  <Button width="default">
    Open Popover
  </Button>
</Popover>
```

jordangarcia:
The pattern we've been using within the react_components directory is an `activator` prop
and the contents being the children

PR #648, opened 2016-10-24
Create a React Poptip component.

### LINKED ISSUE #519: Create OUI React Poptip component
Needed by @mikeng13 and depends on #462.

danoc:
This creates a `Poptip` component that takes in a string of `content` and an activator as the `children`.

Here's an example:

```html
<Poptip content="Ipsa officiis bad-news minus earum a aperiam! Aperiam reiciendis
          vitae nihil libero et, hic!">
  <Button>
    Basic Poptip
  </Button>
</Poptip>
```

It can also accept more advanced positioning props (that are passed to `OverlayWrapper`). Here's a screenshot of an example:

screenshot 2016-10-24 13 19 02

***
**Future work:**
- Refactor `Popover` API to better match this one. (https://github.com/optimizely/oui/issues/646)
- Consider adding `poptip` prop to `Button`, `Icon`, and other components. (https://github.com/optimizely/oui/issues/647)

***
This fixes #519.
***
@jordangarcia, @daverau-optimizely – Can you review this please?


src/components/OverlayWrapper/index.js
```diff
@@ -133,6 +135,26 @@ class OverlayWrapper extends React.Component {
     }
   }
 
+  onChildMouseOver(event, child) {
+    // Enable Tether when visible.
+    this.enableTether();
+
+    // Run the `children`'s `onMouseOver` if it exists.
+    if (child.props.onMouseOver) {
+      child.props.onMouseOver(event);
```
jordangarcia: Is there a chance that this could be not a single child?

Should `React.Children.only` be used?

src/components/OverlayWrapper/index.js
```diff
@@ -141,19 +163,33 @@ class OverlayWrapper extends React.Component {
...
       /* eslint-disable react/jsx-no-bind */
-      <div data-test-section={ this.props.testSection }>
+      <div
+        data-test-section={ this.props.testSection }
+        style={ { display: 'inline-block' } }>
         <div
           style={ { display: 'inline-block' } }
-          ref={ (ref) => { this._activatorEl = ref; } }>
-          { Children }
+          ref={ (ref) => { this._activatorEl = ref; } }
```
jordangarcia: ahhhhh is this how you bind the ref of a child to the current component???

I've needed this before.

src/components/Poptip/Poptip/index.js
```diff
@@ -0,0 +1,26 @@
...
+      style={ { display: 'inline-block', opacity: 1, position: 'initial' } }
+      data-test-section={ props.testSection }>
+      { props.children }
+    </div>
+  );
+};
+
+Poptip.propTypes = {
+  /** Content that appears within the popover body */
+  children: React.PropTypes.string.isRequired,
```
jordangarcia: Do you think we'd ever use `Poptip` by itself without overlaywrapper?

if so should we expose a `isVisible` prop that actually controls the display / opacity style props.

I'm fine either way since it's a non-apparent use case.

src/components/Poptip/index.js
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Displays help text when hovering on an element.
+ * @param {Object} props - Properties passed to component
+ * @returns {ReactElement}
+ */
+const PoptipWrapper = (props) => (
+  <OverlayWrapper
+    behavior="hover"
+    overlay={
+      <Poptip data-test-section={ props.testSection }>
```
jordangarcia: This should be `testSection`

src/components/OverlayWrapper/index.js
danoc: `ref="myRefName"` also works, but React removed that syntax from the documentation and is probably going to deprecate it: https://facebook.github.io/react/docs/refs-and-the-dom.html

src/components/Poptip/index.js
danoc: Nice catch, thanks!

src/components/OverlayWrapper/index.js
danoc: Nice catch. Added a check for that (and a test).

PR #652, opened 2016-10-24
Add a `centerGroup` prop to `ButtonRow`.

### LINKED ISSUE #651: ButtonRow to allow center buttons
This is needed for the share link popover in the results page.

danoc:
This fixes #651 and adds a `centerGroup` prop to `ButtonRow` so that it looks like this:
screenshot 2016-10-24 15 16 19

***
cc: @jordangarcia


danoc:
@jordangarcia – Here is the updated `ButtonRow` that will work when the child buttons are full width.

screenshot 2016-10-24 16 32 02

jordangarcia:
Lgtm

PR #658, opened 2016-10-25
Expose a `hideOverlay` function in `OverlayWrapper` as context

### LINKED ISSUE #657: Expose a `hide` function in `OverlayWrapper`
There is currently no way for a child of `OverlayWrapper` to close the overlay.

This was first requested by @jordangarcia in #654.

danoc:
@jordangarcia – I haven't manually tested this yet and am not sure what kind of automated test to write.

Let me know what you think though!

***
Fixes #657


jordangarcia:
LGTM,

I'd say the correct test to write would like something like:

```
const TestOverlay = (props, context) => {
  render() {
    return <div></div>
  }
};

TestOverlay.contextTypes = {
  hideOverlay: React.PropTypes.func.isRequired,
};

const component = mount(<OverlayWrapper overlay={ TestOverlay }><button /></OverlayWrapper>);

expect(component.find('TestOverlay').context().hideOverlay).to.be.a.function;
```

You could also invoke the function and make sure it sets the correct state and calls the correct component methods.

danoc:
Added a manual test in the examples and wrote an automated test (it was actually a bit tricker than expected because `.context()` can only be used on the root).

Going to merge now. Thanks for the help!

PR #662, opened 2016-10-25
Remove `react-addons-test-utils` since it is not being used.

### LINKED ISSUE #661: Remove `react-addons-test-utils` from dependencies
It's no longer being used anywhere (thanks to Enzyme) and I've manually verified that tests pass without it.

danoc:
This should be safe to move if the build still passes. (I've tested locally, too.)

Fixes #661.

***
@daverau-optimizely


danoc:
Well, looks like it is needed. ¯\_(ツ)_/¯

PR #669, opened 2016-10-31
Add `isDraggable`, `order`, and `description` props to `Token`.

### LINKED ISSUE #667: Update the `Token` component to support cases where it has an order, is rearrangeable, and has a description
@jordangarcia needs this for work that he is doing.

danoc:
This PR does the following:

* Adds a `isDraggable` prop to `Token` that shows an `EllipsisIcon` when true.
* Adds a `order` prop to `Token` that shows an a number when provided.
* Adds a `description` prop to `Token` that appears under the `name` prop when rendered.
* Refactors the `DismissButton` function to be a subcomponent that is tested separately.
* Only show the `DismissButton` when both `isDismissible` and `onDismiss` are provided.
* Adds tests for the new functionality.

***
screenshot 2016-10-31 14 40 38
***
This fixes #667.
***
@jordangarcia, can you review?


danoc:
@jordangarcia – I also need to add support for `onDrag` functions.

Do you know which of these are relevant? Does it depend on the library we choose to adopt?

PR #673, opened 2016-11-09
Add `button--underline` style to OUI

daverau-optimizely:
Hey @danoc here's my WIP diff for adding the `button--underline` style.

My codepen: http://codepen.io/optidave/pen/EyZbJA

I used is-open to avoid writing extra CSS overrides for the is-active class, but we may want to use that to be consistent. Thanks for taking this over for me! I really appreciate all your hard work, Dan. I wouldn't be able to go on leave without ya, so thanks for all your help all the time.


danoc:
I've cleaned the code up a bit and added a React example.

It looks like this:

screen shot 2016-11-11 at 3 59 55 pm

```html
<Button
  style="underline"
  width="default"
>
  Underline Button&nbsp;
  <ArrowsInline direction="down" />
</Button>
```

`onClick` the component implementing this should change the position of the arrow to up. I'd rather this part be implemented in the Optimizely codebase (at least for now) since it is a combination of a few components: `Button`, `ArrowsInline`, and `Dropdown` (which isn't in OUI yet).

@kwalker3690 – Can you please review the Sass?
@jordangarcia – Can you please review the JS?

danoc:
@jordangarcia – Do you still need this?

jordangarcia:
@danoc yeah looks like we will still need it for the upcoming metrics work in Q1

raykhenberg:
There doesn't appear to be much JS to check. This is a button, so I will need to write a wrapper to make this part of a dropdown. I can do that in my component, but I was hoping that we could have done this in OUI. Would it be possible to update the Dropdown component to include an option to use this button instead of the normal button?

Issue #679, opened 2016-11-30
Only load relevant languages in `Code` component syntax highlighting


danoc:
[It seems like it is possible](https://github.com/isagalaev/highlight.js/issues/1257#issuecomment-254504876) to only bundle a subset of languages when requiring highlight.js. We should investigate and make this change for performance improvements.

PR #680, opened 2016-11-30
Asynchronously require code syntax highlighting

danoc:
We initially had syntax highlighting baked into the `Code` component but removed it because the highlight.js package is huge and was potentially causing issues in Optimizely.

This brings back syntax highlighting but loads it using `require.ensure` so that a separate bundle is created and only fetched when needed.

***

**Edit:** The next step will be to only require the languages that are needed. That will be done separately and is tracked in #679.

***
@mjc1283 – Can you review this please?


danoc:
@mjc1283 – Can you take a look at this when you get a chance?

src/components/Code/index.js
```diff
@@ -7,46 +8,83 @@ import CopyButton from './CopyButton';
...
-  let copy = null;
+  componentWillMount() {
+    // Load highlight.js asynchronously
+    if (typeof require.ensure !== 'function') {
+      require.ensure = (d, c) => {
+        c(require);
+      };
+    }
+
+    require.ensure(['highlight.js'], (require) => {
```
mjc1283: As discussed offline, I feel like this isn't the right place for this, but I get that it solves our immediate issue. It would be cool in the future if importing this component imported highlight.js immediately, but importing other components would continue to exlcude highlight.js.

src/components/Code/index.js
```diff
@@ -7,46 +8,83 @@ import CopyButton from './CopyButton';
...
+    }
 
-  if (props.type === 'inline') {
     return (
+      /* eslint-disable react/no-danger */
       <code
-        className='oui-code'
-        data-test-section={ props.testSection }>
-        { props.children }
+        className={ this.props.type === 'inline' && 'oui-code' }
```
mjc1283: Nit: maybe `this.prop.type === 'inline' ? 'oui-code' : ''` is clearer. I assume it works as-is, but didn't know offhand what the effect would be of assigning `false` as the `className` prop.

src/components/Code/index.js
```diff
@@ -7,46 +8,83 @@ import CopyButton from './CopyButton';
...
+class Code extends React.Component {
+  constructor() {
+    super();
+    this.renderCode = this.renderCode.bind(this);
   }
 
-  let copy = null;
+  componentWillMount() {
+    // Load highlight.js asynchronously
+    if (typeof require.ensure !== 'function') {
```
mjc1283: Nit: comment that this is done for tests. Also, might be good to hoist this to the top of the file - it doesn't need to be in `componentWillMount`, right?

src/components/Code/index.js
danoc: Yeah, good catch. It actually adds the class "false", so I'll change this.

src/components/Code/index.js
danoc: Fixed. Thanks!

src/components/Code/index.js
danoc: Yeah, good catch. I've changed this. It was actually adding the class "false" to the element.

PR #689, opened 2017-01-07
Clean up stray DOM nodes on`OverlayWrapper` unmount

### LINKED ISSUE #670: Looks like OverlayWrapper isn't cleaning up stray tether DOM after unmount

danoc:
This fixes #670. I've opened an issue with Tether to make sure that this is the desired approach on their end: https://github.com/HubSpot/tether/issues/236


src/components/OverlayWrapper/index.js
```diff
@@ -121,6 +121,7 @@ class OverlayWrapper extends React.Component {
   componentWillUnmount() {
     this.removeBodyEventListner();
     this._tether.destroy();
+    this._tether.element.remove();
```
jordangarcia: It may be better to do

```
if (this._tether.element && this._tether.element.remove) {
  this._tether.element.remove()
}
```

Just as a precaution since I assume this isn't 100% api stable

danoc: Good call. Will do!

danoc:
Tests are passing locally. Seems to be an issue with Travis CI on this PR so I'm going to go ahead and merge.

Issue #686, opened 2016-12-10
Revamp the OUI documentation


danoc:
I've started to work on a revamped version of the OUI documentation to address the feedback UIE have received from engineers and designers. This issue exists to point out the problems with [the current documentation](http://design.optimizely.com/docs/oui/18.2.0/) and share the proposed solutions.

## Problems with the current documentation

@daverau-optimizely and I have found the following problems in the OUI documentation from talking with engineers, designers, and using it ourselves.

### Navigation Issues
- [Sass](http://design.optimizely.com/docs/oui/18.2.0/sass) and [React](http://design.optimizely.com/docs/oui/18.2.0/react) documentation are separate. This leads to a frustrating user experience since we haven’t reached parity between Sass and React components. More importantly though, OUI provides a number of helpful utility (“trump”) classes that are only documented in Sass and are helpful even to developers working with React. There is also no link between the two documentation pages.
- Sass documentation is slow because there is too much in the DOM. This happens since we display all of the documentation on one large page.
- Sass documentation has confusing whitespace because it displays information from all files, even those that don’t have meaningful content.
- Component names are not always intuitive making it hard for the user to find what they are looking for.
- React documentation becomes harder to navigate as more components are added. This happens because the documentation is a single page, the navigation is not sticky, component code examples can get long, and component names are not always obvious.

### Missing Content
- Icons, colors, and font sizes aren’t documented in either documentation site. This is a huge pain point for designers that want to reference the styles and engineers that want to find the classes to implement them.
- Sass documentation isn’t quickly deployed when new versions are released. This is because, unlike the React documentation, the Sass documentation infrastructure lives outside of the OUI repository and is not part of an automated build system.
- There are missing examples in Sass documentation. This likely happens because it is very hard to view the examples in the OUI codebase.
- React OUI sub-components are not documented, making components like `Table` hard to use.
- There are many components that don’t exist in React and there is no link to the Sass documentation for those components.

### Incorrect Content
- Examples in Sass documentation don’t always work if they require JavaScript. We once wrote JavaScript to power a handful of our components, but the JavaScript depends on jQuery and hasn’t been adopted outside of the OUI documentation examples.
- Examples using JavaScript don’t work in the Optimizely codebase:
  - The components that require JavaScript use Vue.js in Optimizely but jQuery in OUI. This makes it hard to write examples that work for both.
[...]

PR #692, opened 2017-01-18
Create a V2 of the documentation that combines both Sass and React

danoc:
This addresses years of feedback received about the OUI documentation. The feedback has been described in #686.

***

screenshot 2017-01-18 10 01 10

***
### To run locally:
- Checkout this branch.
- `nvm use` (It may ask you to install a new version of Node. If so, run `nvm install v6.9.1`.)
- `npm install`
- `npm run docs:watch`

### Critical tasks left to do:
- [x] Go through Utils and see what examples need to be added
- [x] Fix the lint
- [x] Figure out how deploying this thing will work
- [x] Do a diff of the outputted CSS before and after this change to ensure that nothing is breaking
- [x] Read through all the changed code

***
@kwalker3690 – Going to add you to this as a reviewer since we're going to chat about it later.


danoc:
I've confirmed that the resulting CSS files are the same on `devel` and `daniel/docs`.

https://www.diffchecker.com/KhdmulZe

PR #696, opened 2017-01-27
Vertically align checkbox and radio with the top of the label.

### LINKED ISSUE #695: Vertically align checkbox to be at the top of a multi-line label

danoc:
See discussion in Slack for context:
https://optimizely.slack.com/archives/design/p1485468366000018

This fixes #695.

***

**After Screenshot:**
screenshot 2017-01-27 10 28 02


src/components/Checkbox/index.js
```diff
@@ -18,14 +18,15 @@ const Checkbox = (props) => {
...
-      <div className="flex flex-align--center">
+      <div className="flex">
         <input
           type="checkbox"
           defaultChecked={ props.defaultChecked }
           checked={ props.checked }
           disabled={ props.isDisabled }
           onChange={ props.onChange }
           data-test-section={ props.testSection }
+          style={ { 'marginTop': '0.35em' } }
```
kwalker3690: jw, why these values?

kwalker3690: And have you cross browser tested this? I feel like this has potential to be funky in different browsers

danoc: Just did a quick check in Safari and Firefox and it works well. Good call on checking though!

And this value makes it to line up well. Here's a screenshot without the margin:
screenshot 2017-01-27 13 00 07

We previously did not need this because we were doing `flex-align: center`, but we can't do that anymore because we don't want the checkbox to be center aligned if the label spans multiple lines.

PR #699, opened 2017-02-02
Remove `text-shadow` from buttons since the shadows are inconsistent.

danoc:


quintondang:
lgtm--does this mean all buttons will now inherit their text shadow from the parent?

danoc:
@quintondang – Sort of. They will have the initial value (`none`) unless something else on the page gives them a text shadow (which shouldn't happen).

PR #705, opened 2017-02-08
Add `Label` Sass examples to the documentation.

danoc:
This adds `Label` Sass examples and moves the file into a component folder.

Here's the diff of the CSS after moving the file in the source order:
https://www.diffchecker.com/3lHTbqoF

There's no difference between the block that was removed and the block that was added.

***
screenshot 2017-02-07 16 09 25


src/components/Checkbox/index.scss
```diff
@@ -0,0 +1,14 @@
+////
```
kwalker3690: Is there no scss needed, but the file is needed for examples? If so, is it possible to add a comment explaining this?

danoc: Good call! Thanks.

PR #710, opened 2017-02-15
Simplify the steps in `CONTRIBUTING.md`

danoc:
You can view the new version in a nicer format here:
https://github.com/optimizely/oui/blob/007ff9e8f5ddddcc1267a93d2e422e61082de6ba/CONTRIBUTING.md


danoc:
@spencerwilson-optimizely – The deprecated thing isn't useful nowadays because we're making fewer breaking changes.

And the beta stuff was a bit too confusing and I think that documentation had holes. I think removing it is better because it distracts from the rest of the document and doesn't add much value.

PR #712, opened 2017-02-15
Move Sass file for `Select` to add it to the documentation.

danoc:
screenshot 2017-02-15 11 31 38

This moves the file so that it appears in the documentation.


src/components/Select/index.scss
```diff
@@ -35,8 +37,4 @@
     font-size: map-fetch($button, size, large, font-size);
     height: map-fetch($button, size, large, height);
   }
-
-  &::-ms-expand {
```
kwalker3690: this is no longer needed?

danoc: Not so sure. Will leave it just in case though.

Seems like it might still be relevant:
https://developer.mozilla.org/en-US/docs/Web/CSS/::-ms-expand

PR #740, opened 2017-07-03
First pass wrapping input to show form-note

daverau-optimizely:
@inakiaranzadi-optimizely This is my quick attempt to add form-notes to the input react component, but I feel there is probably a more thoughtful approach similar to the way label is handled. I'm curious your thoughts here!

screenshot 2017-07-03 13 26 02

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This file is similar to the format suggested by [Keep a CHANGELOG](https://github.com/olivierlacan/keep-a-changelog).
 
 ## Unreleased
+### Added
+- [Feature] Add form-note as prop to Input component for more parity with SASS.
 
 ## 25.3.0 - 2017-05-30
 ### Added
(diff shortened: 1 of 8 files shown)


inakiaranzadi-optimizely:
How can we avoid repeating `.oui-form-bad-news` class when `hasError` is true and it have a `label` prop happening?

```html
<div class="oui-form-bad-news">
  <label>
    <div class="oui-label"><!-- react-text: 7 -->adfadf<!-- /react-text --></div>
    <input type="text" min="10" max="50" class="oui-text-input oui-form-bad-news" value="some default value" placeholder="just a placeholder">
  </label>
</div>
```

inakiaranzadi-optimizely:
Also we are styling label with a wrapper. I wonder if passing props to `<Label />` is a better more scalable alternative. What do you think @daverau-optimizely?

daverau-optimizely:
This latest push updates Label to handle displayError along with some minor CSS changes. Updating tests now.

screenshot 2017-07-06 15 54 47

inakiaranzadi-optimizely:
@daverau-optimizely  I think the build is failing because os some `lint` warning and errors.

PR #782, opened 2017-09-07
Update react-oui-icons to solve bug on build.

inakiaranzadi-optimizely:
Update *react-oui-icons* to solve bug on build.

@daverau-optimizely


daverau-optimizely:
This is currently failing on `FAIL  src/components/Token/tests/index.js (14.349s)
  ● components/Token › should show ellipsis icon if the token is draggable`

Is this still a Glamor issue or just a test we need to update?

PR #786, opened 2017-09-18
Inaki/dropdown component

inakiaranzadi-optimizely:
Beta version of the Dropdown components for OUI based on Optimizely existing implementation.

@daverau-optimizely


src/components/Dropdown/index.js
```diff
@@ -0,0 +1,74 @@
...
+import classNames from 'classnames';
+import PropTypes from 'prop-types';
+
+class Dropdown extends React.Component {
+  render() {
+    const {
+      children,
+      handleClick,
+      icon = 'chevron',
+      isDisabled,
```
jackschlotthauer-optimizely: you might want to default isDisabled to false, isOpen = false (when will a drop down start open?) and width to something like 200px (which is what the storybook uses.

src/components/Dropdown/index.js
```diff
@@ -0,0 +1,74 @@
...
+      isOpen,
+      fullWidth,
+      style,
```
jackschlotthauer-optimizely: is this a style object? or a parameter for a classname?

inakiaranzadi-optimizely: good call, I'll add a classname prop good call.

src/components/Dropdown/index.js
inakiaranzadi-optimizely: Good suggestion.

PR #802, opened 2017-09-25
Drau/empty state react component

daverau-optimizely:
BLOCKED: I'm still fixing image sizes for this component and will update the CSS width/height based on the final dimensions we set.


daverau-optimizely:
I messed up this commit with `yarn.lock` and a `package.json` — I'll try to revert this before squads & merge. FYI @inakiaranzadi-optimizely

PR #849, opened 2017-11-21
Add padding prop to Popover component

circAssimilate:
Add padding prop to Popover component, allowing for three padding classes or a default of no padding trumps.

This is needed to accomplish the design for [select container](https://docs.google.com/document/d/1sGpL3egjnjF0uESvYHTTaSfITBl7MuhYKemX3uaG0N4/edit#), while using the OUI component. Open to prop naming and approach, just wanted to put this out there for a poc review. 

If you all are on board with this approach, let me know and I can follow the rest of the contributing guidelines.


inakiaranzadi-optimizely:
@circAssimilate all this errors should go away if you rebase devel or master.

circAssimilate:
I just added an issue for this in #862.

circAssimilate:
I've added tests for this and updated the docs as well.

updated-docs

I'm not 100% sure what to do for storybook, but let me know what you think about this PR!

daverau-optimizely:
@circAssimilate Can you drop in your working code from the select element/container work for the storybook example? I think that would work very nicely to show the padding and scrollbar stuff (which I just added to OUI and will release later today)

src/components/Popover/index.js
```diff
@@ -12,21 +17,39 @@ import PropTypes from 'prop-types';
...
+  const popOverClasses = classNames({
     'oui-pop--over': true,
     'highlight-react--oui': localStorage.getItem('show_ouireact') === 'true',
   });
 
+  let popOverContentClasses = classNames({
+    'oui-pop--over__content': true,
+  });
+
+  const shouldAddPaddingClass = padding !== DEFAULT_PADDING_PROP && [
```
daverau-optimizely: @inakiaranzadi-optimizely Do you think we should we use CONST values vs hard-coded values since they are unlikely to change?

PR #861, opened 2017-12-07
Ricky/attention

raykhenberg:
I don't know if I'm doing this right, but it seems like having anchor tags and such in attention components should be pretty common.


inakiaranzadi-optimizely:
@raykhenberg this looks right. Also I like the idea to accept nodes as children here and probably we can expand this to other components like Disclose, were we are limited to a label.

Issue #863, opened 2017-12-13
Add scrollbar visible helper


daverau-optimizely:

daverau-optimizely:
Added in latest release.

PR #864, opened 2017-12-13
Drau/scrollbars

daverau-optimizely:
Adding scrollbar helper class so they are always visible on overflow containers. I'm assuming most people will use this for vertical scrolling so that's the default style I added here.

cc @circAssimilate


daverau-optimizely:
Fixes #863

PR #882, opened 2018-01-05
Add max height overrides and fix

daverau-optimizely:


daverau-optimizely:
Cancelling this after discussion with @inakiaranzadi-optimizely since we don't use these styles anywhere else, haven't had a need until this point, and probably would be better served using an inline style so it's more clear what's happening in this unique component. cc @circAssimilate

PR #884, opened 2018-01-09
Jess/readonly nofocus

jessicalc:
this feels like my first OUI PR!


daverau-optimizely:
Just FYI we typically squash and merge on review, so we'll release this for ya!

PR #891, opened 2018-01-16
Add arrowIcon prop to OUI Dropdown component for UI arrow

circAssimilate:
Just an initial commit for technical feedback. Docs can be added as well as tests if this approach is agreed upon.


src/components/Dropdown/index.js
```diff
@@ -67,6 +83,20 @@ const Dropdown = withToggle(({
...
 
 Dropdown.propTypes = {
+  /** Arrow icon direction:
+    * - Defaults to 'none', which hides the arrow
+    * - passing a prop value of false also hides the arrow
+    * - passing a prop value of true uses down arrow
+    */
+  arrowIcon: PropTypes.oneOf([
+    'down',
+    false,
```
circAssimilate: I must have forgot to take these Boolean propTypes out. Should we remove?

PR #924, opened 2018-02-22
ensure accordian class continues to work alongside oui-accordian

jamesopti:
@jessicalc


jessicalc:
What an elegant solution!

PR #943, opened 2018-04-26
Add new waitForSelector util to OverlayWrapper and wait for body before creating Tether

circAssimilate:
When mounting the OverlayWrapper, which uses the `tether` library, we need to ensure the `body` element is present in the document, which does not happen in it's current state when using the OverlayWrapper on components mounted to the RootVm like the Global Nav. This is due to how we delay loading the rest of the DOM before loading the VM. This diff will:

- Add a new util file to OUI with waitUntil and waitForSelector 
- Delay the creation of a Tether within OverlayWrapper until the body is present
- Update tests to accommodate for change

The `waitUntil` util is only used in the OUI tests. I found with jest that is was pretty hard to a) stub a promise and b) not rely on something like `waitUntil` to do the final assertions. Open to suggestions about better ways to accomplish this.

For posterity, this change was done after exhausting options laid out in Tether's documentation and after creating an [issue](https://github.com/HubSpot/tether/issues/285) with Tether to see if the body being required was expected behavior - and it sounds like it is.

I also removed an unneeded `.isRequired` from the `isConstrainedToScreen` to screen prop, since it has a default value provided already.

I don't have a great understanding of what the `package-lock.json` changes were, so would love more information on what is expected with regards to that.

I will complete the OUI Bug Bash tracker when I'm ready to release.

https://docs.google.com/document/d/1TTfdhCSH7mPBeUzVme99qHR-QsFg7PTKP2lGqB9Dk3Y/edit#heading=h.ogbetpi1tm4k


src/components/OverlayWrapper/tests/index.js
```diff
@@ -36,15 +37,18 @@ describe('components/OverlayWrapper when componentDidMount', () => {
...
 
     const instance = component.instance();
+
     spyOn(instance, 'disableTether');
     const initialCallCount = instance.disableTether.calls.count();
 
-    instance.componentDidMount();
-
-    expect(instance.disableTether.calls.count()).toBe(initialCallCount + 1);
+    waitUntil(() => instance.createTether.calls.count() !== 0, { shouldResolveOnTimeout: true })
```
jamesopti: I dont want to introduce this util just for testing, esp considering we dont need it.

I'd rather just use a global `beforeEach` to ensure our body exists.

circAssimilate: Yea, I felt really gross about it. I didn't even consider waiting for the body in a before each, which is really silly of me. Thanks!

daverau-optimizely:
Thanks for working on this Derek. Iñaki had reworked Dropdown and a few other components to use Popper.js as a replacement to overlay wrapper but didn't get a chance to finish everything before he left. I have a copy of his work that I could put up in a branch if that's at all helpful for interesting for you.

circAssimilate:
@daverau-optimizely, I'd love to at least see the branch to see how it could help us here - but I think that would expand the scope of this quite a bit (at least regarding all the QA that should be done) so may keep the changes separate if that's ok with you.

circAssimilate:
@jamesopti, this is ready for review again. I had quite the time working out the test timing issues - in 3 or so situations I ended up having to wait for the body again within a `waitForSelector` `.then()`. I'm pretty sure this is just passing due to getting put back on the stack, so it doesn't feel done yet - just wanted to get up what I had. Another thing that stuck out as odd to me was adding this to the `beforeEach()` would cause the tests to fail:

```
    waitForSelector('body').then(done);
```

But adding this wouldn't:

```
    waitForSelector('body').then(() => {
      done();
    });
```

I'm not aware of any reasons these would execute differently, but let me know what you think. I do not have the bandwidth to take on the Popper implementation or I would just do that here, but I'm tempted to use the Dropdown component (which uses popper) if we can't find a great solution for this.

circAssimilate:
Thanks for the IRL help @jamesopti! The stuff we paired on works like a charm and simplifies this greatly.

src/components/OverlayWrapper/index.js
```diff
@@ -40,10 +41,12 @@ class OverlayWrapper extends React.Component {
       options.targetAttachment = `${this.props.verticalTargetAttachment} ${this.props.horizontalTargetAttachment}`;
     }
 
-    this._tether = this.createTether(options);
-    // Disable Tether after creation for performance improvements. This is okay
-    // since it is hidden by default.
-    this.disableTether();
+    waitForSelector('body').then(() => {
```
jamesopti: Please add a comment as to why this nonsense is necessary :/

src/components/OverlayWrapper/tests/index.js
```diff
@@ -18,6 +19,19 @@ TestButton.propTypes = {
...
 };
 
+
+// Set in mountComponentAndSubscribeToBodyReady, this can be used
+// to ensure the body exists
+let bodyReadyPromise;
+
+// Create helper function to mount the component and return
+// a promise that will resolve when the body is ready
+const mountComponentAndSubscribeToBodyReady = (componentConfig) => {
```
jamesopti: `... because < OverlayWrapper> will do the same before it initializes tether`

src/components/OverlayWrapper/tests/index.js
```diff
@@ -18,6 +19,19 @@ TestButton.propTypes = {
   text: PropTypes.string,
 };
 
+
+// Set in mountComponentAndSubscribeToBodyReady, this can be used
+// to ensure the body exists
+let bodyReadyPromise;
```
jamesopti: Nit: As someone glancing at a single test below, this might be slightly more clear as `tetherReadyPromise.then...`

PR #949, opened 2018-05-08
Drau/link newwindow

daverau-optimizely:


src/components/Link/index.js
```diff
@@ -43,6 +45,8 @@ Link.propTypes = {
   href: PropTypes.string,
   /** Disable link and interactivity */
   isDisabled: PropTypes.bool,
+  /** new */
```
circAssimilate: Maybe we can make this "Open link in new window" instead?

src/components/Link/index.js
```diff
@@ -30,6 +31,7 @@ const Link = ({
       data-test-section={ testSection }
       data-track-id={ testSection }
       onClick={ onClick }
+      {...(newWindow ? { target: '_blank', rel: 'noopener' } : {})}
```
circAssimilate: nit: just add spaces on both sides of these brackets I'd say:

`{ ...(newWindow ? { target: '_blank', rel: 'noopener' } : {}) }`

daverau-optimizely: 👍

src/components/Link/index.js
daverau-optimizely: 👍

PR #951, opened 2018-05-09
Update Poptip to wait for the body before rendering Tooltip

circAssimilate:
Similar to the recent updates to `OverlayWrapper` (#943), OUI should not assume the body will be available at all times. This uses the `waitForSelector` utilities to set state and fully render the `Poptip` component when the body is available. If the body is not available, `null` will be returned. Additionally this diff will:

* Fix a bug with the `disabled` prop, and allow for the Poptip to be conditionally disabled.
* Expose `isAnimated` prop to disable animation in `Tooltip`. Under the hood, this just sets the animation duration to `0`

**TEST PLAN**
Updated snapshot and unit tests


src/components/Poptip/index.js
```diff
@@ -8,54 +9,85 @@ import { Tooltip } from 'react-tippy';
...
 
-const Poptip = ({
-  children,
-  content,
-  disable,
-  position,
-  trigger,
-  theme,
-}) => (
-  <Tooltip
```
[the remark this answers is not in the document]
circAssimilate: @jamesopti, I just tried using `document.body &&` as you suggested and it did as I feared, since `Tooltip`, which is wrapped in OUI `Poptip` contains the actual trigger element, it just never renders if it's falsy first. I verified it wasn't just the implementation by noticing it work with `true &&` and then switching back to `document.body &&`.

src/components/Poptip/index.js
```diff
@@ -8,54 +9,85 @@ import { Tooltip } from 'react-tippy';
...
+  // trigger prop left out as Trigger default, 'mouseenter focus', will be used
+  static defaultProps = {
+    disable: false,
+    isAnimated: true,
+    position: 'top',
+    theme: 'dark',
+  };
+
+  state = {
+    bodyIsReady: false,
```
jamesopti: This initial state should be `!!document.body`, but this can be handled easily with the `withState` decorator.

src/components/Poptip/index.js
```diff
@@ -8,54 +9,85 @@ import { Tooltip } from 'react-tippy';
...
+        html={ <div>{content}</div> }
+        position={ position }
+        theme={ theme }
+        trigger={ trigger }>
+        <div>{ children }</div>
+      </Tooltip>
+    );
+  };
+
+  componentWillMount() {
```
jamesopti: Can you move this to the constructor?

Then it can also just be `waitForSelector('body').then(this.setBodyDefined)`

src/components/Poptip/index.js
```diff
@@ -8,54 +9,85 @@ import { Tooltip } from 'react-tippy';
  * @param {Object} props - Properties passed to component
  * @returns {ReactElement}
  */
+export default class Poptip extends React.Component {
```
jamesopti: For really basic state like this, I prefer the `@withState` decorator from recompose:

`@withState('bodyDefined', 'setBodyDefined', !!document.body)`

src/components/Poptip/tests/__snapshots__/index.js.snap
```diff
@@ -1,44 +1,15 @@
...
-  interactiveBorder={2}
-  multiple={false}
-  offset={0}
-  onHidden={[Function]}
-  onHide={[Function]}
-  onRequestClose={[Function]}
-  onShow={[Function]}
-  onShown={[Function]}
-  popperOptions={Object {}}
+<Poptip
```
circAssimilate: @jamesopti @daverau-optimizely, because this was changed to a React component (or potentially because of the `withState` decorator), I had to update this snapshot to only be the component that was mounted. I couldn't easily find a way to preserve the `Tooltip` prop assertion. Any thoughts?

src/components/Poptip/Poptip.story.js
```diff
@@ -58,4 +58,17 @@ stories
         <a>Click Poptip, light theme, on the right</a>
       </Button>
     </Poptip>
+  )))
+  .add('Customizable with knobs', withInfo()(() => (
```
circAssimilate: Added a story for all the props too.

src/components/Poptip/tests/__snapshots__/index.js.snap
jamesopti: I actually think this is more correct!

circAssimilate: Sweet, I was thinking it made more sense than the inner component being checked but wasn't sure. Thanks!

PR #954, opened 2018-05-24
Textfield percent

daverau-optimizely:
# Outline of changes:

screen shot 2018-05-24 at 2 08 13 pm

1. Add another type to TextField called `percent` that's a number type and will append the % sign outside of the form field, and correctly vertically centered the % sign
2. Add text-align right class for all number and percent types
3. Update Jest snapshot tests for TextField

# Kinds of issues this change should fix:

Percent % symbol has bad vertical alignment:
screen shot 2018-05-24 at 10 48 43 am

Numbers not right-aligned:
screen shot 2018-05-24 at 1 48 57 pm

Input doesn't use the correct oui-input-text class so is missing focus state:
screen shot 2018-05-24 at 1 49 32 pm


src/components/TextField/index.js
```diff
@@ -38,9 +38,16 @@ const TextField = ({
   );
 
   let classes = classnames(
-    'oui-text-input', {'oui-text-input--search': isFilter}
+    'oui-text-input',
+    {'oui-text-input--search': isFilter},
+    {'text--right': (type === 'number' || type === 'percent')}
```
circAssimilate: This can be simplified like this! (using just one object for the conditionals)

```
const inputClassNames = classnames('oui-text-input', {
  'oui-text-input--search': isFilter,
  'text--right': (type === 'number' || type === 'percent')
})
```

Note: You can use `const` instead of `let` since it won't be reassigned (unless I missed where that's happening)

Also, supernit, I think the name `inputClassNames` would be nice and more descriptive!

PR #959, opened 2018-06-07
Reorder plugins in babelrc according to recommentations from owners

kwalker3690:
It is recommended that `transform-decorators-legacy` is placed before `transform-class-properties` in the README of the former: https://github.com/loganfsmyth/babel-plugin-transform-decorators-legacy


circAssimilate:
@jamesopti, looking at this file's history I do not see it, but I had thought you previously ordered these this way to fix an issue with how the `@connect` decorator transpiled decorators then classes in babel. Is that not right?

circAssimilate:
Scratch that! I was out of the loop and thinking a) this was `optimizely.git` and b) not understanding the core issue with `@withState()`. These updates look in line with https://github.com/optimizely/optimizely/blob/devel/src/www/frontend/.babelrc (which I know has similarly saved us in the past), so LGTM!

circAssimilate:
For posterity though, I wonder why this wasn't causing a problem for us in `optimizely.git` @jamesopti. 

@kwalker3690, I doubt if you'll be bumping `optimizely.git` with this change (although it would be awesome if you did), but please at least add a note pointing to this commit in the [OUI Bash Tracker](https://docs.google.com/document/d/1TTfdhCSH7mPBeUzVme99qHR-QsFg7PTKP2lGqB9Dk3Y/edit#heading=h.ogbetpi1tm4k) so the next person to bump knows to check an implementation of the OUI Poptip when bashing.

kwalker3690:
To elaborate on the problem, with the incorrect order of plugins, according to React debug tools, MGMT was rendering Poptips as
```
<Poptip>
  <Poptip>
     <Tooltip />
  </Poptip>
</Poptip>
```
After this change, according to React debug tools, MGMT renders Poptips as 
```
<withState(Poptip)>
  <Poptip>
     <Tooltip />
  </Poptip>
</withState(Poptip)>
```
Still a bit odd, but seems closer to being correct. Would love to know if there is a way to setup our babel compilation to not have withState(Poptip) become a React element itself

daverau-optimizely:
Hey @kwalker3690 I'm going to make a new OUI release tomorrow so let me know if we should squash & merge to include these or not.

PR #960, opened 2018-06-08
Adding Dropdown list components

kwalker3690:
In Optimizely's frontend we've created a Dropdown component that has subcomponents to help format the contents of the Dropdown. This diff ports those components from https://github.com/optimizely/optimizely/blob/master/src/www/frontend/src/js/react_components/dropdown/index.js to be subcomponents of the OUI dropdown component.

One difference between the two implementations is that the Optimizely implementation makes use of context to allow children components to toggle the dropdown open/closed and to hide the dropdown. This context is not necessary for the OUI implementation, so it has been removed.

### DIFF
--- a/src/components/Button/tests/index.js
+++ b/src/components/Button/tests/index.js
@@ -26,6 +26,19 @@ describe('components/Button', () => {
     expect(obj.func).toHaveBeenCalled();
   });
 
+  it('should call function that is passed in as `onBlur` after losing focus', () => {
+    const obj = { func: () => {} };
+    spyOn(obj, 'func').and.stub();
+
+    const component = shallow(
+      <Button onBlur={ obj.func }>Hello!</Button>
+    );
+
+    component.simulate('blur');
+
+    expect(obj.func).toHaveBeenCalled();
+  });
+
   it('should add an `aria-label` when provided', () => {
     const component = shallow(
       <Button ariaLabel="a11y">Hello!</Button>
--- a/src/components/Dropdown/DropdownBlockLink/index.js
+++ b/src/components/Dropdown/DropdownBlockLink/index.js
@@ -1,0 +1,53 @@
+import React from 'react';
+import classNames from 'classnames';
+import PropTypes from 'prop-types';
+
+export default class DropdownBlockLink extends React.Component {
+  static defaultProps = {
+    isLink: true,
+  };
+
+  static propTypes = {
+    /** Content to be shown in the menu option */
+    children: PropTypes.node.isRequired,
+    /** should this be a link or nonclickable text */
+    isLink: PropTypes.bool.isRequired,
+    /** Minimum width of the list item, useful if
+      * you need to have a block of description text */
+    minWidth: PropTypes.oneOfType([
+      PropTypes.string,
+      PropTypes.number,
+    ]),
+    /** Click handler for the menu option */
+    onClick: PropTypes.func.isRequired,
+    /** Used for data-test-section attribute on the link */
+    testSection: PropTypes.string,
+    /** Used for data-track-id attribute on the link */
+    trackId: PropTypes.string,
+  };
+
+  onClick = () => {
+    this.props.onClick();
+  };
+
+  render() {
+    const styleProps = {};
+    if (this.props.minWidth) {
+      styleProps.minWidth = this.props.minWidth;
+    }
+    return (
+      <div
+        className={ classNames({
+          'link': this.props.isLink,
+          'muted': !this.props.isLink,
+          'oui-dropdown__block-link': this.props.isLink,
+        }) }
+        style={ styleProps }
+        data-test-section={ this.props.testSection }
+        data-track-id={ this.props.trackId }
+        onClick={ this.onClick }>
+        { this.props.children }
+      </div>
+    );
+  }
+}
--- a/src/components/Dropdown/index.js
+++ b/src/components/Dropdown/index.js
@@ -5,11 +5,11 @@ import { Manager, Popper, Target } from 'react-popper';
 import { withState } from 'recompose';
 import { withToggle } from '../../utils/recompose-utils';
 
-import DropdownContents from './DropdownContents.js';
-import DropdownListItem from './DropdownListItem.js';
-import DropdownBlockLink from './DropdownBlockLink.js';
-import DropdownBlockLinkText from './DropdownBlockLinkText.js';
-import DropdownBlockLinkSecondaryText from './DropdownBlockLinkSecondaryText.js';
+import DropdownContents from './DropdownContents';
+import DropdownListItem from './DropdownListItem';
+import DropdownBlockLink from './DropdownBlockLink';
+import DropdownBlockLinkText from './DropdownBlockLinkText';
+import DropdownBlockLinkSecondaryText from './DropdownBlockLinkSecondaryText';
 
 class DropdownActivator extends React.Component {
   static propTypes = {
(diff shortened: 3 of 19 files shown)


jamesopti:
+ @jessicalc 

The goal is to replace all app dot usage with this ported one, correct?

kwalker3690:
Yep exactly! That would be the follow up to this

src/components/Dropdown/index.js
```diff
@@ -66,21 +98,34 @@ class Dropdown extends React.Component {
...
-              <div className='flex--1 truncate'>{ buttonContent }</div>
-              {
-                !!arrowIcon && arrowIcon !== 'none' && (
-                  <div className='text--right'><span className={ iconClass }/></div>
-                )
-              }
-            </div>
-          </button>
+          {
+            buttonContent && (
```
circAssimilate: Will there ever be a time a consumer would want both `buttonContent` and `activator`? Should we either ensure the other is not present before rendering? At least, I'd suggest explicitly mentioning this in the docstrings.

src/components/Dropdown/index.scss
```diff
@@ -99,7 +99,8 @@
   vertical-align: top;
 
   &.is-active {
-    .#{$namespace}dropdown {
+    .#{$namespace}dropdown,
+    .oui-dropdown {
```
circAssimilate: @daverau-optimizely @kwalker3690, should we be adding `display--block` to the dropdown or combining these two selectors?

kwalker3690: Yeah curious to hear @daverau-optimizely's opinion on this change :)

kwalker3690:
> I'm very excited to start migrating these to OUI. I want to be sure I'm understanding the benefit over using these sub components. Is it that migration will be easier and/or the hope to guide more consistent styling of these dropdown elements? I pulled your branch and ran storybook and saw the example but don't immediately see the difference in behavior between this and the others.

IMO the necessity of these components is for guiding consistency. Similar to how we have the Table component with TH, TD, etc - you can definitely create a table in React without them, but we use those components for consistent styling. I'm definitely open to being challenged on the necessity of the subcomponents, but if we choose not to use the subcomponents, the refactor in app dot will be much more complicated, potentially unnecessarily so. I think Dropdown should have been created with opinionated subcomponents like this from the getgo.

src/components/Dropdown/index.js
kwalker3690: Good call I will make it clearer in the docstrings. I also wish there was a way to enforce that using proptypes...

src/components/Dropdown/DropdownBlockLink/index.js
```diff
@@ -0,0 +1,53 @@
...
+    return (
+      <div
+        className={ classNames({
+          'link': this.props.isLink,
+          'muted': !this.props.isLink,
+          'oui-dropdown__block-link': this.props.isLink,
+        }) }
+        style={ styleProps }
+        data-test-section={ this.props.testSection }
+        data-track-id={ this.props.trackId }
```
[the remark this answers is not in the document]
kwalker3690: Oh really? TBH I have no idea - I'm not even sure who owns that. It is still used in a number of places in our frontend, so I'd rather err on the side of keeping support for it now than breaking things. Maybe we can add a follow up task for Yeesheen/Magda to remove it from the codebase so we know it's not necessary to have support for here?

src/components/Button/tests/index.js
```diff
@@ -26,6 +26,19 @@ describe('components/Button', () => {
     expect(obj.func).toHaveBeenCalled();
   });
 
+  it('should call function that is passed in as `onBlur` after losing focus', () => {
+    const obj = { func: () => {} };
```
jamesopti: I know this pattern was here before, but it feels clunky and unnecessary.

kwalker3690: Ha yeah that's true, I can simplify

daverau-optimizely:
I'll defer to others for the React parts and I just posted my thoughts on the CSS options above.

Thank you Kelly for working on this so quickly.

src/components/Dropdown/index.scss
kwalker3690: I normally veer towards the classname reuse in HTML approach as well, the only reason I went this direction is that I'm not really adding CSS here, just adding another name variation of the class. I'm actually not sure adding `oui-dropdown` is necessary here, along with all of the other places we have ```
.#{$namespace}<classname>
.oui-<classname>
```
Shouldn't we just set `oui` as our namespace? Is there a historical reason for this @daverau-optimizely

Comment (daverau-optimizely) [reply src/components/Dropdown/index.scss]: Hah, `.#{$namespace}`, `lego-`, `oui-`, ... what could go wrong and be confusing about that?! 🙃🙃🙃 

So the plan has been to move all OUI React components to use `oui-` classes even if they are duplicate rules of the `lego-` classes so at least going forward we could discern between classes used in React vs old classes to phase out at some point.

In theory once we're 100% react we can move away from `lego-` and probably kill the whole idea of the namespace, or at least fix it to `oui-`

Comment (kwalker3690) [reply src/components/Dropdown/index.scss]: So then isn't adding `oui-dropdown` here the 'correct' thing to do based on that plan?

Comment (daverau-optimizely) [reply src/components/Dropdown/index.scss]: Yep, that's correct!

File: src/components/Dropdown/index.js [hunk 5]
```diff
@@ -5,6 +5,36 @@ import { Manager, Popper, Target } from 'react-popper';
...
+    children: PropTypes.node,
+    handleOnBlur: PropTypes.func,
+    handleToggle: PropTypes.func,
+  };
+
+  render() {
+    // loop through children and add props to them
+    // we need this so that we can pass the props to the children
+    // passed into the activator
+    const childrenWithProps = React.Children.map(this.props.children,
```
jamesopti: From my understanding of how this component is used, there is no need for this to handle multiple children.

Am I missing a reason we cant just use `activator` in the main `<Dropdown>` like this?

```javascript
activator && (
  <activator handleToggle={ this.handleToggle } handleBlur={ this.handleBlur }>
)

```

src/components/Dropdown/index.js
```diff
@@ -171,4 +222,12 @@ Dropdown.defaultProps = {
   arrowIcon: 'none',
 };
 
-export default withToggle(Dropdown);
+const DropdownWithToggle = withToggle(Dropdown);
```
jamesopti: Curious why you chose this export interface (as properties of the class and not separate).

The consumption of these components in optimizely.git is often standalone.

Why not export them separately?

jamesopti: Oh also - You should be able to stack the `withToggle` as a decorator above `@withState` as long as you give the component a `displayName` (so the tests can still find it)

src/components/Dropdown/index.js
kwalker3690: Hmm... tbh this was a copy/pasta with just a change to not depend on context, but I think your solution will work better and be simpler! I will try it and see if there was a legit reason for the current version.

src/components/Dropdown/index.js
kwalker3690: I chose this approach because it matches [the pattern we use](https://github.com/optimizely/oui/blob/master/src/components/Table/index.js#L69) in Table, which I actually really like. It makes it super explicit that DropdownListItem etc are only meant to be used within Dropdown - if you want JSX for a list or a link outside of a dropdown, you should not use these components.

jamesopti: If thats the case, I completely agree. I thought I saw it being used outside that context, but I could be wrong.

This will require some manual updates in Optimizely.git but I think its worth it.

PR #961, opened 2018-06-14
Adding Select Dropdown to OUI

kwalker3690:
This diff moves react_components/select_dropdown to OUI as SelectDropdown. To enable the filtering functionality of SelectDropdown, Dropdown had to be refactored so that clicking on DropdownListItem toggles the Popper contents to be hidden. Previously, clicking on the Popper itself hid the contents - this meant that if a user tried to click into the filter input in a SelectDropdown, the dropdown contents would be hidden. 

This means that users of the Dropdown either must use Dropdown.ListItem, or must call this.props.handleToggle when the Dropdown should be closed. IMO, strongly encouraging use of Dropdown.ListItem in the contents of the Popper should be our approach, this will help us continue to have consistently styled Dropdowns.

### DIFF
--- a/src/components/SelectDropdown/index.js
+++ b/src/components/SelectDropdown/index.js
@@ -229,7 +229,7 @@ class SelectOption extends React.Component {
 
 SelectOption.defaultProps = {
   description: '',
-}
+};
 
 export default SelectDropdown;
 
--- a/src/components/SelectDropdown/index.js
+++ b/src/components/SelectDropdown/index.js
@@ -37,8 +37,8 @@ class SelectDropdown extends React.Component {
       description: PropTypes.string,
       label: PropTypes.string.isRequired,
       value: PropTypes.oneOfType([
-        PropTypes.string.isRequired,
-        PropTypes.number.isRequired,
+        PropTypes.string,
+        PropTypes.number,
         PropTypes.bool,
       ]).isRequired,
     })).isRequired,
@@ -85,6 +85,7 @@ class SelectDropdown extends React.Component {
 
   static defaultProps = {
     buttonStyle: 'outline',
+    inputPlaceholder: '',
     isFilterable: false,
     dropdownDirection: 'right',
     width: '100%',
@@ -103,7 +104,6 @@ class SelectDropdown extends React.Component {
   };
 
   filterTermValue = (value) => {
-
     return isFilterTermInItem(this.state.searchTerm, value);
   };
 
@@ -154,7 +154,7 @@ class SelectDropdown extends React.Component {
     return (
       <Dropdown
         width={ width }
-        zIndex={ zIndex }
+        { ...(zIndex ? { zIndex } : {}) }
         activator={ (
           <Button
             isDisabled={ this.props.isDisabled }
@@ -180,7 +180,7 @@ class SelectOption extends React.Component {
      */
     description: PropTypes.string,
     /** Toggle dropdown open/closed */
-    handleToggle: PropTypes.func,
+    handleToggle: PropTypes.func.isRequired,
     /**
      * Whether or not item has been selected or not.
      */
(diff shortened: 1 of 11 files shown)


daverau-optimizely:
`users of the Dropdown either must use Dropdown.ListItem, or must call this.props.handleToggle when the Dropdown should be closed` I am a fan of this default behavior Kelly, nice work.

daverau-optimizely:
Hey @kwalker3690 I'm going to make a new OUI release tomorrow so let me know if we should squash & merge to include these or not.

src/components/SelectDropdown/index.js
```diff
@@ -0,0 +1,231 @@
...
+     * Dropdown items that can be selected from the select dropdown.
+     */
+    items: PropTypes.arrayOf(PropTypes.shape({
+      description: PropTypes.string,
+      label: PropTypes.string.isRequired,
+      value: PropTypes.oneOfType([
+        PropTypes.string.isRequired,
+        PropTypes.number.isRequired,
+        PropTypes.bool,
+      ]).isRequired,
```
circAssimilate: Any idea why this isn't `PropTypes.bool.isRequired`? Seems like there maybe shouldn't be an `isRequired` for any of the value PropTypes, but I'm not sure.

src/components/SelectDropdown/index.js
```diff
@@ -0,0 +1,231 @@
...
+    this.props.items.forEach(item => {
+      if (item.value === value) {
+        selectedItem = item;
+      }
+    });
+
+    return (
+      <Dropdown
+        width={ width }
+        zIndex={ zIndex }
```
circAssimilate: Should `zIndex` be conditionally set here, like `...(zIndex ? {} : { zIndex })`?

src/components/SelectDropdown/index.js
```diff
@@ -0,0 +1,231 @@
...
+     * Width of the activator container.
+     */
+    width: PropTypes.oneOfType([
+      PropTypes.string,
+      PropTypes.number,
+    ]),
+    /**
+     * zIndex of dropdown group
+     */
+    zIndex: PropTypes.number,
```
circAssimilate: See comment below

src/components/SelectDropdown/index.js
```diff
@@ -0,0 +1,231 @@
...
+     */
+    buttonStyle: PropTypes.string,
+    /**
+     * Dropdown direction.
+     */
+    dropdownDirection: PropTypes.oneOf(['right', 'left']),
+    /**
+     * Placeholder text for the filter input.
+     */
+    inputPlaceholder: PropTypes.string,
```
circAssimilate: Can we make this an empty string in the defaults?

src/components/SelectDropdown/index.js
```diff
@@ -0,0 +1,231 @@
...
+  };
+
+  search = (event) => {
+    this.setState({
+      searchTerm: event.target.value,
+    });
+  };
+
+  filterTermValue = (value) => {
+
```
circAssimilate: nit: can we remove this space?

src/components/SelectDropdown/index.js
```diff
@@ -0,0 +1,231 @@
...
+}
+
+class SelectOption extends React.Component {
+  static propTypes = {
+    /**
+     * Description of select item.
+     */
+    description: PropTypes.string,
+    /** Toggle dropdown open/closed */
+    handleToggle: PropTypes.func,
```
circAssimilate: Can we make this required or provide it with a default, empty function? Ditto on the above.

src/components/SelectDropdown/index.js
kwalker3690: Yeah good call, agreed none of the values should be required, I think that's probably just ignored

kwalker3690:
Arrrg this should have been marked as a breaking change - it changes the innards of how dropdowns operate. My bad - @daverau-optimizely fyi

PR #963, opened 2018-06-18
Drau/index js export

daverau-optimizely:
That fix worked @alansun-optimizely! Tried it locally and all was well, so adding this back and will make a new release to include this change.


alansun-optimizely:
Part of me feels weird adding back in a file that's changed, but I think it's harmless enough

PR #967, opened 2018-06-21
Add min/max to percent textField and update tests.

daverau-optimizely:
This fixes design debt: https://optimizely.atlassian.net/browse/DESIGN-65

@alansun-optimizely Can you forsee any issues forcing min=0 max=100 on percent TextFields? I believe this is a reasonable change but wasn't sure if we allow greater than 100% anywhere; are you aware?


daverau-optimizely:
Thanks @jessicalc, I'm going to ship this and added an issue to discuss the right approach to testing this vs just relying on the jest snapshot output.

PR #972, opened 2018-06-26
Add note to Input, and note and label to Textarea

kwalker3690:
I'd like our form states to be more consistent, so I'm adding an optional note and label and note to input and textarea. This allows us to have consistent styling of these notes and of their error states. (Also it looks like we used to have a form note on input, based on the storybook options, but it may have gotten removed during the big refactor)

Input with label and note:
image
Input with label and note and error state:
image

Textarea with label and note:
image
Textarea with label and note and error state:
image

### DIFF
--- a/src/components/Input/index.js
+++ b/src/components/Input/index.js
@@ -79,7 +79,7 @@ class Input extends React.Component {
   }
 
   renderNote({note, testSection}) {
-    return (<div className='oui-form-note' data-test-section={ testSection && testSection + '-note' }>{note}</div>)
+    return (<div className='oui-form-note' data-test-section={ testSection && testSection + '-note' }>{note}</div>);
   }
 
   render() {
@@ -88,7 +88,7 @@ class Input extends React.Component {
       displayError,
       testSection,
       isOptional,
-      note
+      note,
     } = this.props;
 
     if (label) {
@@ -114,10 +114,10 @@ class Input extends React.Component {
           { this.renderInput(this.props) }
           { this.renderNote(this.props) }
         </div>
-      )
+      );
     }
 
-    return this.renderInput(this.props)
+    return this.renderInput(this.props);
 
   }
 }
--- a/src/components/Input/index.js
+++ b/src/components/Input/index.js
@@ -109,12 +109,10 @@ class Input extends React.Component {
     }
 
     if (note) {
-      return (
-        <div>
-          { this.renderInput(this.props) }
-          { this.renderNote(this.props) }
-        </div>
-      );
+      return [
+        this.renderInput(this.props),
+        this.renderNote(this.props),
+      ];
     }
 
     return this.renderInput(this.props);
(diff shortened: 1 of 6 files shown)


src/components/Input/index.js
```diff
@@ -98,11 +103,22 @@ class Input extends React.Component {
...
+    }
+
+    if (note) {
+      return (
+        <div>
+          { this.renderInput(this.props) }
+          { this.renderNote(this.props) }
         </div>
       );
     }
```
circAssimilate: Just a thought, you probably could be able to do one of these:

```
return [
  this.renderInput(this.props),
  this.renderNote(this.props),
]
```

or:

```
return (
<div>
{ [
  this.renderInput(this.props),
  this.renderNote(this.props),
] }
</div>
)
```

Such a small difference so just a suggestion. I would generally prefer the first approach but have witnessed some of our components like `OverlayWrapper` break when multiple elements are returned, even though PropTypes.node specified an array of nodes is valid. ¯\_(ツ)_/¯

src/components/Input/index.js
```diff
@@ -139,6 +155,8 @@ Input.propTypes = {
    * Min value for the `input`. Should be used only when `type` is `number`.
    */
   min: PropTypes.number,
+  /** Form note for the input */
+  note: PropTypes.string,
```
circAssimilate: Sorry, `defaultProp` police here :sob:. Can you add an empty string default or specify in the prop note undefined is expected?

src/components/Textarea/index.js
```diff
@@ -55,17 +57,75 @@ class Textarea extends React.Component {
...
+            </div>
+            { this.renderTextarea(this.props) }
+          </Label>
+          { note && this.renderNote(this.props)}
+        </div>
+      );
+    }
+
+    if (note) {
+      return (
```
circAssimilate: Ditto on the light suggestion of returning an array.

src/components/Textarea/index.js
```diff
@@ -55,17 +57,75 @@ class Textarea extends React.Component {
...
+    return null;
+  },
   /** Prevents textarea from being modified but doesn't appear disabled */
   isReadOnly: PropTypes.bool,
   /** Prevents textarea from being submitted without value */
   isRequired: PropTypes.bool,
+  /** Text that describes the textarea */
+  label: PropTypes.string,
+  /** Form note for the input */
+  note: PropTypes.string,
```
circAssimilate: samsies re: defaultProp

src/components/Input/index.js
kwalker3690: Cool I didn't know you can return an array!

src/components/Textarea/index.js
```diff
@@ -90,4 +148,11 @@ Textarea.propTypes = {
   value: PropTypes.string,
 };
 
+Textarea.defaultProps = {
+  displayError: false,
+  label: null,
+  note: null,
```
circAssimilate: Won't this throw a prop warning? Since an empty string is falsy I thought that would work.

PR #985, opened 2018-07-07
Fixes #984 - Add HelpPopover component to OUI

### LINKED ISSUE #984: Migrate react_components/help_popover to OUI
This replaces #982 since there is already an existing help tooltip component in optimizely.git; it just needs to be migrated into OUI for use in program management and other places.

jessicalc:
This diff adds a HelpPopover to the OUI component library; the next step would be to [move it out of react_components](https://github.com/optimizely/optimizely/blob/da1463f6f8fff3847e8a457d909d3a647cadd8da/src/www/frontend/src/js/react_components/help_popover/index.js). @alansun-optimizely

### DIFF
--- a/src/components/HelpPopover/HelpPopover.story.js
+++ b/src/components/HelpPopover/HelpPopover.story.js
@@ -16,21 +16,19 @@ stories
   ));
 
 stories
-  .add('default', withInfo()(() => (<div>
-    <div className="position--relative height--100 text--center">
-      <HelpPopover popoverTitle="This is a popover title">
-        <p>
-          This is a popover description. 🤓
-        </p>
-      </HelpPopover>
-    </div>
-  </div>)))
+  .add('default', withInfo()(() => (
+    <HelpPopover popoverTitle="This is a popover title">
+      This is a popover description. 🤓
+    </HelpPopover>
+  )))
   .add('hover', withInfo()(() => (<div>
     <div className="position--relative height--100 text--center">
-      <HelpPopover popoverTitle="This is a popover title" behavior="hover">
-        <p>
-         You hovered! 🤓
-        </p>
+      <HelpPopover
+        popoverTitle="This is a popover title"
+        behavior="hover"
+        horizontalAttachment="left"
+        verticalAttachment="middle">
+       You hovered! 🤓
       </HelpPopover>
     </div>
   </div>)));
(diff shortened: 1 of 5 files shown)


jamesopti:
LGTM assuming this is just copy/pastad from optimizely/react_components

src/components/HelpPopover/HelpPopover.story.js
```diff
@@ -0,0 +1,36 @@
...
+stories
+  .addDecorator(withKnobs)
+  .addDecorator(story => (
+    <div id="root-preview">
+      {story()}
+    </div>
+  ));
+
+stories
+  .add('default', withInfo()(() => (<div>
```
daverau-optimizely: Could we drop the `<div>` wrapper here? In my tests this works without issue.

src/components/HelpPopover/HelpPopover.story.js
```diff
@@ -0,0 +1,36 @@
...
+    <div className="position--relative height--100 text--center">
```
daverau-optimizely: I see why you smartly included this in the example, I wonder if the first storybook example should be the purest/most minimal without presentation style when possible, so something without any wrappers and minimal html for children. (This is the type of stuff I debate by myself as a team of one!)
```
  .add('Left', withInfo()(() => (
    <HelpPopover
      behavior="hover"
      horizontalAttachment="left"
      verticalAttachment="middle"
      >
        This is a short description. 🤓
    </HelpPopover>
  )))
```

src/components/HelpPopover/index.js
```diff
@@ -0,0 +1,68 @@
...
+  popoverTitle: PropTypes.string,
+  /** Test section */
+  testSection: PropTypes.string,
+  /** Vertical edge of the `overlay` that should touch the `children` */
+  verticalAttachment: PropTypes.oneOf(['top', 'middle', 'bottom']),
+};
+
+HelpPopover.defaultProps = {
+  behavior: 'click',
+  horizontalAttachment: 'center',
```
daverau-optimizely: This is just my personal preference but I've long disliked this prop setup for Overlay wrapper. What about simplifying this to a `position` prop with 8 possible values: 
- `top-left`, `top-center`, `top-right`
- middle could be simply `left` or `right`
- `bottom-left`, `bottom-center`, `bottom-right`
- and pick a default either `right`, or `bottom-center`

src/components/HelpPopover/HelpPopover.story.js
jessicalc: Sounds good! I just copy pasta'ed this from elsewhere. (hides)

src/components/HelpPopover/HelpPopover.story.js
jessicalc: Done!

PR #997, opened 2018-07-18
Rotate NPM Key

alansun-optimizely:
Due to the eslint-scope security hole, our npm key was automatically revoked. Our Travis CI job was therefore not able to publish to NPM. I've created a new key in NPM and updated the travis.yml file to include the new key.


jamesopti:
@alansun-optimizely In the future, can you notify #frontend and #oui that youre doing this?

We should be storing these keys in secret server.

PR #1000, opened 2018-07-26
Added isDisabled prop to BlockList.Item to disable the Item

spham1412:
Before: There's no way to disable an individual item from a BlockList.

Changed: Added an optional `isDisabled` prop to the BlockList.Item component.

After: If the `isDisabled` prop is passed in with `true`, it will only render the prop.children and ignore all other actions. The rendered item is also greyed out and un-clickable. The new prop is optional, so it shouldn't break any existing usage of the BlockList.Item component.

Tests: Added unit-tests and ensure existing tests passed.


src/components/BlockList/Item/tests/index.js
```diff
@@ -4,6 +4,26 @@ import { shallow, mount } from 'enzyme';
 
 describe('components/BlockList/Item', () => {
   describe('#render', () => {
+    describe('`isDisabled` is provided', function() {
+      let component;
+      const text = 'Goose';
+      const bool = true;
```
quintondang: could you add a quick test to check for the negative assertion? bool = false and can't find the associated classes you provided?

spham1412: Done

PR #1003, opened 2018-07-26
Add maxLength prop to Input and Textarea

mjc1283:
Input and Textarea components get new maxLength props (optional, of number type). These are passed directly through to the child input/textarea element.


src/components/Input/index.js
```diff
@@ -149,6 +151,11 @@ Input.propTypes = {
    * Max value for the `input`. Should be used only when `type` is `number`.
    */
   max: PropTypes.number,
+  /**
+   * Max length of the input value. Should be used only when type is 'text',
+   * 'email', 'search', 'password', 'tel', or 'url'.
+   */
+  maxLength: PropTypes.number,
```
circAssimilate: > Should be used only when type is 'text', 'email', 'search', 'password', 'tel', or 'url'.
Just confirming, nothing fatal happens when it is added to a type that isn't one of these - right? It's just a no-op basically.

src/components/Textarea/index.js
```diff
@@ -47,6 +48,7 @@ class Textarea extends React.Component {
         required={ isRequired }
         readOnly={ isReadOnly }
         disabled={ isDisabled }
+        maxLength={ maxLength }
```
circAssimilate: >Feedback copy/pasta

How will we treat the default of this? I think we should expect undefined and not add this value if it is undefined (since undefined props will still be added)

```
<textarea
	data-oui-component={ true }
    className={ classes }
	...(typeof maxLength === "undefined" ? { maxLength } : {})
/>
```

I've used and seen that pattern a few times throughout OUI at least. Open to the solution!

src/components/Input/index.js
```diff
@@ -71,6 +72,7 @@ class Input extends React.Component {
         onFocus={ onFocus }
         min={ min }
         max={ max }
+        maxLength={ maxLength }
```
circAssimilate: How will we treat the default of this? I think we should expect undefined and not add this value if it is undefined (since undefined props will still be added)

```
<input
	data-oui-component={ true }
    className={ classes }
	...(typeof maxLength === "undefined" ? { maxLength } : {})
/>
```

I've used and seen that pattern a few times throughout OUI at least. Open to the solution!

mjc1283: When `maxLength=undefined` is passed, it seems React renders an `input` element with no `maxlength` attribute. But I like your suggestion and will make that change.

src/components/Input/index.js
mjc1283: Right - I tested this with a few other types, and passing `maxLength` as a prop had no effect.

PR #1010, opened 2018-07-30
Tweaks to SelectDropdown and Poptip

kwalker3690:
- Allow users to provide a width for the activator of SelectDropdown (allows us to keep static button size as content of button changes)
- Force break content of Poptip in case of really long string:
image


package.json
```diff
@@ -91,7 +91,7 @@
     "babel": "babel src/ --out-dir dist/js --ignore \"tests/*.js, src/js\"",
     "codecov": "codecov",
     "jest": "jest --config jest.config.js --no-cache",
-    "jest:watch": "jest --config .jest.config.json --watch",
+    "jest:watch": "jest --config jest.config.js --watch",
```
alansun-optimizely: my bad. sorry about that.

Issue #1012, opened 2018-08-03
Update Checkbox to take in optional "description" string or node, and to take "label" as a string or a node


jessicalc:
Posted on behalf of zeeshan:

I have these checkboxes in the profile section which are in vue. while reactifying them, i wanted to use `<Checkbox>` OUI component. I am facing two problems with `<Checkbox>`
1. I dont see an option to make the label bold.
2. This also has a one liner description for each checkbox. I dont see an option to provide html to `<Checkbox>` label. If i keep description outside the `<Checkbox>` component, it messes up the alignment.

PR #1013, opened 2018-08-03
Update Checkbox's label prop to accept node #1012

jessicalc:
I updated the Checkbox to take a node in addition a string for the `label` prop.


src/components/Checkbox/index.js
```diff
@@ -42,7 +43,16 @@ const Checkbox = ({
           style={{ 'marginTop': '0.35em' }}
         />
         <div className={ labelClassNames }>
-          { label }
+          { !description && label }
+          { description &&
+            (
+              <div>
+                <strong>{ label }</strong>
```
kwalker3690: Can we use the label component here?

jessicalc: That's an EXCELLENT idea, let me fix

jessicalc: The entire checkbox input field is _already_ wrapped inside a `<Label>`! (Which I didn't even realize until you brought this up).

The reason for this seems to be so that the entire label is clickable, and not just the checkbox. For this reason, I don't think we need to add a `<Label>` component back in here. 

BUT! Zeeshan gave some good feedback in the #oui channel: 

> @jess .. while this is good enough for my use case.  just a suggestion for general use, instead of assuming label to be bold whenever description is available, isn't it better to take "making label bold" as a separate option irrespective of description being there or not? or even simpler, if label also accepts `<node>` optionally, then we might not even need a separate field for description or a bold property, We can format it the way we want. 

So I've updated the PR to take that into account.

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This file is similar to the format suggested by [Keep a CHANGELOG](https://github.com/olivierlacan/keep-a-changelog).
 
 ## Unreleased
+### Added
+- [Feature] Updated Checkbox component to take in description, and accept node for label prop. #1012
```
kwalker3690: This log is now wrong!

jessicalc: LOL okay thanks

PR #1014, opened 2018-08-08
EditableInput component

pauloptimizely:
**Summary**

Adding an `EditableTextField` component, that has similar behavior to the html5 `contenteditable` attribute. 

editabletextfield-demo

**Test Plan**

1. Run the test suite `npm test`
2. Verify the snapshots are correct
3. Run storybook and click around

**Jira**
no ticket


src/components/EditableTextField/EditableTextField.js
```diff
@@ -0,0 +1,129 @@
...
+                autoFocus={ true }
+                defaultValue={ defaultValue }
+                displayError={ displayError }
+                label={ label }
+                note={ note }
+                onFocus={ this.focus }
+                onBlur={ this.blur }
+                onChange={ onChange }
+                placehoder={ placeholder }
+                type={ type }
```
pauloptimizely: This might be too verbose of an implementation. I considered just passing all of the props to `TextField` and picking out props that need to be overridden.

For example:
```js
<TextField
  {...this.props}
  autoFocus={ true }
  onFocus={ this.focus }
```

Any opinions on verbose props definitions vs "loose" props defintions?

src/components/EditableTextField/tests/index.js
```diff
@@ -0,0 +1,106 @@
...
+
+    it('should render TextField when isEditing is true', function() {
+      const component = shallow(
+        <EditableTextField
+          isEditing={ true }
+          type='text'
+        />
+      );
+
+      expect(shallowToJson(component)).toMatchSnapshot();
```
kwalker3690: Is there a reason you are asserting on the snapshot for these tests? I personally find assertions like this really confusing as a dev looking at the test - I'd rather see an assertion like `expect(component.find('TextField').exists()).toBe(true);` - which I see you actually have in a test below this! 

This may completely reveal my dislike/possible lack of understanding of the usefulness of snapshot testing!

src/components/EditableTextField/EditableTextField.js
```diff
@@ -0,0 +1,128 @@
...
+      onChange,
+      placeholder,
+      type,
+      value,
+    } = this.props;
+
+    const { isEditing } = this.state;
+
+    return (
+      <div data-oui-component={ true } className={ classNames('oui-editable-text-field', className) }>
```
kwalker3690: are the classnames in this file just used for test hooks? If so, use `data-test-section` instead! We like to keep class names used only for styling, so that we can reconfigure elements/change classes on elements without the possibilty of breaking their connections to tests

package.json
```diff
@@ -8,6 +8,7 @@
   "dependencies": {
     "clipboard": "^1.5.12",
     "highlight.js": "^9.5.0",
+    "lodash.noop": "^3.0.1",
```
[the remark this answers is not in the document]
pauloptimizely: I think we have to keep this in dependencies, since it's being used in the component itself, since npm ignores devDependencies when publishing.

But if it's included in the build when we run `npm run build` than we should move it. I'll have to do some investigating.

src/components/EditableTextField/EditableTextField.js
```diff
@@ -0,0 +1,128 @@
...
+                onChange={ onChange }
+                placehoder={ placeholder }
+                type={ type }
+              />
+            )
+            : (
+              <div
+                className="oui-editable-text-field__label editable"
+                onClick={ this.focus }>
+                { value || defaultValue || placeholder }
```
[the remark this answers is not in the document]
pauloptimizely: I can't think of good use case. Only using value makes sense

src/components/EditableTextField/tests/index.js
```diff
@@ -0,0 +1,106 @@
+import React from 'react';
+import { shallow, mount } from 'enzyme';
+import { shallowToJson } from 'enzyme-to-json';
+
+import EditableTextField from '../EditableTextField';
+
+describe('components/EditableTextField', function() {
+
+  describe('basic rendering', function() {
+    it('should render text placehodler if the value is empty', function() {
```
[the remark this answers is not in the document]
pauloptimizely: I googled "HODL" lol

src/components/EditableTextField/tests/index.js
pauloptimizely: I was conforming to using snapshot testing. I agree,  mounting and asserting the existence is more stable!

src/components/EditableTextField/EditableTextField.js
pauloptimizely: It's not being used. I was coding too defensively here; thinking what if I want to manually override a class at some point..

src/components/EditableInput/tests/index.js
```diff
@@ -0,0 +1,95 @@
...
+
+    it('should be editable when label is clicked', function() {
+      const component = mount(
+        <EditableInput
+          type='text'
+        />
+      );
+
+      expect(component.state().isEditing).toBe(false);
+      component.find('.oui-editable-input__label').simulate('click');
```
kwalker3690: Can you hook into a data-test-section for all of these finds instead of looking for a class?

src/components/EditableInput/EditableInput.js
```diff
@@ -0,0 +1,113 @@
...
+    const { isEditing } = this.state;
+
+    return (
+      <div data-oui-component={ true }>
+        {
+          isEditing ?
+            (
+              <Input
+                focus={ true }
+                defaultValue={ value }
```
kwalker3690: I think this should be `value` instead of `defaultValue`

src/components/EditableInput/EditableInput.js
```diff
@@ -1,70 +1,62 @@
...
                 displayError={ displayError }
                 label={ label }
                 note={ note }
                 onFocus={ this.focus }
                 onBlur={ this.blur }
                 onChange={ onChange }
-                placehoder={ placeholder }
```
kwalker3690: We may want to still have placeholder as an option - ie in the case where the field starts off in 'isEditing' mode, with no value

src/components/EditableInput/index.scss
```diff
@@ -0,0 +1,5 @@
+.oui-editable-input {
+  &__label {
+    padding: 6px;
```
kwalker3690: user `spacer(0.5)` to get a value of 5px, less precise but allows us to have more consistency in spacing overall

kwalker3690: Oh or actually its a little more 'oui'-y to just use the class `soft-half` on the element to get 5px of padding on the element

src/components/TextField/index.js
```diff
@@ -1,3 +1,8 @@
+/**
```
kwalker3690: Thanks for adding!

src/components/EditableInput/tests/index.js
```diff
@@ -0,0 +1,95 @@
...
+        />
+      );
+
+      expect(component.find('.oui-editable-input__label').exists()).toBe(false);
+      expect(component.find('Input').exists()).toBe(true);
+    });
+  });
+
+  describe('editing', function() {
+    it('should have Input while editing', function() {
```
kwalker3690: This test is now covered by the previous test "should render Input when isEditing is true"

src/components/EditableInput/tests/index.js
```diff
@@ -0,0 +1,95 @@
...
+          value: newValue,
+        },
+      });
+
+      expect(onChangeSpy).toHaveBeenCalledTimes(1);
+      expect(onChangeSpy.mock.calls[0][0].target.value).toEqual(newValue);
+    });
+
+  });
+});
```
kwalker3690: Can you add a test to make sure that all the other miscellaneous props are passed along to Input?

src/components/EditableInput/EditableInput.js
pauloptimizely: Wait, so in a real world use-case the state of `value` will be managed by the `EditableInput`'s parent. When we pass that value to the `Input` component it's defaultValue should be `EditableInput`'s `value`.

example

```js
class SomeParent extends Component {
   state = {
       myValue: '',
    }

    onChange = event => this.setState({ myValue: event.target.value })

    render() {
         return <EditableInput
              type='text',
              onChange={this.onChange}
             value={ myValue } // When Input mounts the defaultValue is myValue
          />  
     }
}
```

In this scenario `EditableInput` and `Input` have independent states that are synced when `onChange` is called.

kwalker3690: Oh yeah exactly! I was saying that the prop key on Input should be `value`, _not_ `defaultValue`, because we're controlling the component externally. And that calling the prop for EditableInput `value` matches that! Nix defaultValue everywhere in this diff

src/components/EditableInput/EditableInput.js
```diff
@@ -0,0 +1,122 @@
...
+                onFocus={ this.focus }
+                onBlur={ this.blur }
+                onChange={ onChange }
+                placeholder={ placeholder }
+                type={ type }
+              />
+            )
+            : (
+              <div
+                className={ classNames('oui-editable-input__label', 'editable', 'soft-half', testSection) }
```
kwalker3690: We usually use actually a separate attribute, like `<div data-test-section={testSection]>`, then in tests you can grab it as `component.find('[data-test-section="prop-value"]')`

kwalker3690: Also I don't think you need `oui-editable-input__label` now - its no longer being used right?

pauloptimizely: I'll update `data-test-section`, that's much more flexible!!

We don't need `oui-editable-input__label` anymore, OUI internal styles take care of styling here.

src/components/EditableInput/tests/index.js
pauloptimizely: tests are updated

src/components/EditableInput/EditableInput.js
pauloptimizely: ah that's right! Passing `value` to `Input` is the right way, also it's also less confusing

src/components/EditableInput/EditableInput.js
```diff
@@ -0,0 +1,123 @@
...
+              </div>
+            )
+        }
+      </div>
+    );
+  }
+}
+
+EditableInput.defaultProps = {
+  defaultValue: '',
```
kwalker3690: This is no longer needed!

pauloptimizely: 👍

PR #1015, opened 2018-08-08
Dropdown should render children when isOpen is true

pauloptimizely:
**Summary**

Setting `isOpen` prop does not open the `Dropdown` component. The current test is a false positive.

This occurs because recompose maps state to props([source code ref](https://github.com/acdlite/recompose/blob/master/src/packages/recompose/withState.js#L31))

**Test**

Run the test suite with the original `withToggle` implementation; the tests should fail

**Jira**
no ticket


daverau-optimizely:
@pauloptimizely This has bugged me for months, thank you so much for fixing! 💯

Issue #1016, opened 2018-08-10
Add selectors for `.oui-{item}` where `.#{$namespace}item` is used 


jessicalc:
There are a couple of classes that don't have a `oui` selector that I suspect may need it, for example:

Horizontal rules: https://github.com/optimizely/oui/blob/090ed513300a6ffbcee7a5ba964f217ec31ce2bc/src/oui/partials/elements/_horizontal-rules.scss#L13-L14

Pagination:
https://github.com/optimizely/oui/blob/21e1e74d8acf2dd2f8b65ee6c106bbaf5c5d1190/src/components/Pagination/index.scss#L18 

Grid: 
https://github.com/optimizely/oui/blob/21e1e74d8acf2dd2f8b65ee6c106bbaf5c5d1190/src/oui/partials/layout/_grid.scss#L71-L72

Stats
https://github.com/optimizely/oui/blob/21e1e74d8acf2dd2f8b65ee6c106bbaf5c5d1190/src/components/Stats/index.scss#L22-L30

Nav:
https://github.com/optimizely/oui/blob/21e1e74d8acf2dd2f8b65ee6c106bbaf5c5d1190/src/components/Nav/index.scss#L34-L35 

Matrix:
https://github.com/optimizely/oui/blob/21e1e74d8acf2dd2f8b65ee6c106bbaf5c5d1190/src/oui/partials/layout/_matrix.scss#L29-L42

Array:
https://github.com/optimizely/oui/blob/21e1e74d8acf2dd2f8b65ee6c106bbaf5c5d1190/src/components/Array/index.scss#L18-L19 

... and so on 

@daverau-optimizely , is this intentional? Are some of these classes never to be exposed in the oui framework, to be forever locked inside the lego framework?

daverau-optimizely:
@jessicalc Yes this is intentional as we're only adding these `oui-` prefixes once a React component uses these classes. So all React components in OUI should only reference `oui-` prefixes and not `lego-` and we should only add the CSS `oui-` prefix when migrating HTML components to React. Does that make sense?

daverau-optimizely:
Intentional by design. Only add `oui-` prefix when moving `lego-` CSS classes from Vue to React.

PR #1017, opened 2018-08-10
Add oui-rule selector to horizontal rule css #1016

jessicalc:
This adds a `.oui-rule` selector in so we don't need to keep referring to lego classes!! 

See #1016


src/oui/partials/elements/_horizontal-rules.scss
```diff
@@ -11,7 +11,8 @@
 ////
 
 %#{$namespace}rule,
-.#{$namespace}rule {
+.#{$namespace}rule,
```
circAssimilate: Shouldn't `namespace` pick this up - or am I missing what that does?

alansun-optimizely:
Could you also add a HorizontalRule react component?

src/oui/partials/elements/_horizontal-rules.scss
jessicalc: It doesn't pick it up; anything without `oui-` as a prefix won't be available in `oui`. I'm not sure the historic reason for this, but I think @daverau-optimizely will know?

jessicalc:
> Could you also add a HorizontalRule react component?

Not within scope of ticket, but we should ticket it up in JIRA.

src/oui/partials/elements/_horizontal-rules.scss
jessicalc: https://github.com/optimizely/oui/blob/5d7bd8b67a6c0fdfbf4ef606dd406f0e940a142d/src/oui/_oui-variables.scss#L12

PR #1021, opened 2018-08-22
add labelWeight property to Checkbox component

quarklemotion:


src/components/Checkbox/index.js
```diff
@@ -14,13 +14,16 @@ const Checkbox = ({
...
   label,
+  labelWeight,
   onChange,
   testSection,
 }) => {
   const labelClassNames = classNames({
     'flush--bottom': true,
     'push--left': true,
-    'weight--normal': true,
+    'weight--light': labelWeight && labelWeight === 'light',
```
circAssimilate: Instead of doing `labelWeight === 'light'` here, is it possible to add it as a default "prop" through the function declaration? Like this may work:

```
const Checkbox = ({
  checked,
  defaultChecked,
  isDisabled,
  label,
  labelWeight: 'light',
  onChange,
  testSection,
}) => {	}) => {
```

But I was just thinking via `Checkbox.defaultProps` would be best if possible.

circAssimilate: I think you can remove this `&& labelWeight === 'light'` now, right?

src/components/Checkbox/Checkbox.story.js
```diff
@@ -24,6 +24,16 @@ stories
       />
     );
   }))
+  .add('Checkbox with labelWeight provided as string', withInfo()(() => {
+    return (
+      <Checkbox
+        defaultChecked={ true }
+        label={ text('label', 'checkbox component') }
+        labelWeight={ text('labelWeight', 'bold') }
```
circAssimilate: nit: I'm being such a pain, but I think you can also add all the options in here. As long as they're listed in the "show info" section of storybook, but I think polishing up this could help us create quicker.

https://github.com/storybooks/storybook/tree/master/addons/knobs

PR #1028, opened 2018-08-27
Add delay prop in poptip

mrasikh:
**Summary:**
There raised a need where we can have a delay property in Poptip to show a Poptip after a specific time delay. Since we are using react-tippy for Poptip which already supports a delay. So we have just exposed a new prop for our Poptip which it will pass to react-tippy. A story has also been added for this.


src/components/Poptip/index.js
```diff
@@ -28,6 +28,9 @@ export default class Poptip extends React.Component {
       PropTypes.node,
       PropTypes.string,
     ]).isRequired,
+    /** how long it takes after a trigger event is fired for a tooltip to show |
+     * default => 0 */
```
kwalker3690: Please add a defaultProp of 0 for this prop

src/components/Poptip/Poptip.story.js
```diff
@@ -71,4 +71,14 @@ stories
         Customizable Poptip
       </Button>
     </Poptip>
+  )))
+  .add('basic poptip with delay', withInfo()(() => (
```
kwalker3690: Thank you for adding a storybook entry, can you please also add some basic tests for this change?

kwalker3690:
@daverau-optimizely in MGMT, there is a need to have a tooltip appear over text that could or could not be truncated. Shane requested that instead of using the HTML 'title' attribute (which will show after a short delay), that we use a tooltip so that we can have custom styling in it. Since this tooltip will be added to many elements of text on a single page, it would be really jarring if it appeared without any delay, so we'd only like it to appear after a slight delay. Let me know if you have any other questions!

src/components/Poptip/Poptip.story.js
mrasikh: I have added the test case.

src/components/Poptip/index.js
mrasikh: should be good to go now.

mrasikh:
@kwalker3690 it should be good to go now.

PR #1029, opened 2018-08-31
Drau/select dropdown displayError prop

daverau-optimizely:
Here's the diff @mjc1283 

`displayError={ true }`

image


src/components/SelectDropdown/index.js
```diff
@@ -109,11 +113,17 @@ class SelectDropdown extends React.Component {
       }
     });
 
+    const outerClass = classNames(
+      {['oui-form-bad-news']: this.props.displayError}
```
mjc1283: You can get rid of the `[` `]` surrounding `'oui-form-bad-news'`.

PR #1034, opened 2018-09-18
Added auto-focus to the Textarea

mrasikh:
Textarea didn't have a focus prop for auto-focusing purpose. The Input component already has this. I have added the focus prop to Textarea, keeping the consistency with the Input component.


mrasikh:
@circAssimilate it's an interesting question. A little experiment I ran shows if multiple elements have auto-focus enabled, the last one in the DOM will get the priority over others.

And I have added the storybook entry and an example as well.

daverau-optimizely:
@circAssimilate Good question, I'll run `yarn build-storybook`, which I always do just before release, and that will build the static storybook that runs on `go/storybook`. But you can also include that in your commits which is helpful.

https://github.com/optimizely/oui/blob/devel/CONTRIBUTING.md#warning-pre-release

PR #1038, opened 2018-09-20
Drau/colors refined

daverau-optimizely:
1. This updates our SASS and CSS3 color variables with a the latest palette from Shane and d-team (includes badge colors)

screen shot 2018-09-20 at 4 04 30 pm

colors-9-19-18

2. I've simplified the color set so the SASS color definitions also auto generate the css3 variables so we don't have to create two sets of hex color sets manually.

### DIFF
--- a/src/oui/_oui-variables.scss
+++ b/src/oui/_oui-variables.scss
@@ -11,50 +11,28 @@
 
 $namespace                   : '' !default;
 
+/// Spacing Unit
+/// @description Defines spacing for paddings and margins using `spacer`
+/// function.
+
+$spacer-unit                 : 10px;
+
 /// Colors
 /// @description Default color for objects' borders etc.
 
 /// Do not use these values directly in your Sass files. Use the more
 /// specifically named variables below that use these root values.
 
-$base-black                  : #000;
-$base-white                  : #FFF;
-
-// CSS3 Variables
+$base-black                  : #000000;
+$base-white                  : #FFFFFF;
 
 :root {
-  /* brand */
-  --light-blue-100           : #0037FF;
-  --light-blue-75            : #4069FF;
-  --light-blue-50            : #809BFF;
-  --light-blue-25            : #BFCDFF;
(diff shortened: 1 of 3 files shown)


src/oui/_oui-variables.scss
```diff
@@ -11,50 +11,29 @@
 
 $namespace                   : '' !default;
 
+/// Spacing Unit
+/// @description Defines spacing for paddings and margins using `spacer`
+/// function.
+
+$spacer-unit                 : 10px;
```
kwalker3690: @daverau-optimizely was this taken out at some point? This seems pretty crucial?

src/oui/_oui-variables.scss
```diff
@@ -80,19 +61,38 @@ $brand-color: (
...
 );
 
+
+
+// CSS3 Variables
+// Create CSS3 variables for each $brand-color
+// Use `color: var(--light-blue-100);`
+:root {
+  @each $name, $hex in map-fetch($brand-color) {
+    --#{$name}: #{$hex};
```
kwalker3690: Curious why we want to have this var option, instead of just using the $brand-color?

src/oui/_oui-variables.scss
kwalker3690: Oh I see it below! Yeah this organization makes more sense

src/oui/_oui-variables.scss
daverau-optimizely: Ideally I'd like to move away from SASS and just rely on CSS3 features, but shade() and tint() SASS functions are very good and I'm haven't tested enough filter alternatives yet to fully commit to pure CSS3 vars only. We gain some benefits like being able to override the whole design system *after* importing OUI, something James and I discussed could be a nice way to handle importing OUI into the monolith. This is more of an exploratory step to that end.

daverau-optimizely: Open to your thoughts & ideas on this Kelly! 👂

PR #1040, opened 2018-09-26
Drau/switch disabled

daverau-optimizely:
Disaled state (`isDisabled=true`):
screen shot 2018-09-26 at 11 53 35 am

Default off state:
screen shot 2018-09-26 at 11 53 30 am

### DIFF
--- a/src/components/Switch/tests/index.js
+++ b/src/components/Switch/tests/index.js
@@ -1,27 +1,28 @@
 import React from 'react';
 import Switch from '../index';
-import { shallow } from 'enzyme';
+import { shallow, mount } from 'enzyme';
 import { shallowToJson } from 'enzyme-to-json';
 
 describe('components/Switch', () => {
-  let component;
-
-  beforeEach(() => {
-    component = shallow(
-      <Switch />
-    );
-  });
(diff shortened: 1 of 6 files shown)


src/components/Switch/tests/index.js
```diff
@@ -1,23 +1,27 @@
...
 
-  it('should render an `Toggle` with correct props', () => {
+  it('should render an `Switch` with correct props', () => {
     expect(component.length).toBe(1);
   });
 
+  it('should disable the switch', () => {
+    expect(component.hasClass('oui-switch--disabled')).toBe(false);
+  });
+
```
mjc1283: Can we add a test for when the `isDisabled` prop is true? It should assert that the `<input>` has the `'oui-switch--disabled'` class, and that the `<input>` has a `disabled=true` prop.

daverau-optimizely: Good call, can you peep my test update for a quick thumbs up? Just making sure I am covering the bases the way you'd expect.

mjc1283: Looks good, thanks!

PR #1052, opened 2018-10-09
Create Pagination component

pauloptimizely:
## Summary

This PR will focus around moving `PaginationControls` from the monolith to OUI

screen shot 2018-10-09 at 7 14 12 pm

`Pagination` will differ slightly from the original `PaginationControls`. I reduced the following props into a single `onChange` prop using quick maths:

- canGoBack
- canGoForward
- gotoNextPage
- gotoPrevPage
- gotoFirstPage
- gotoLastPage

The `onChange` prop will be called with a single argument; being the next page number.

## Test Plan

Run the test suite

```
npm test
```

Verify everything looks good in storybook

```
npm run storybook
```

## Jira

https://optimizely.atlassian.net/browse/MGMT-962


pauloptimizely:
Ignore this PR, for now. I'm going to rework this component to be compatible with the `Pager` component found in the monolith

https://github.com/optimizely/optimizely/blob/devel/src/www/frontend/src/js/react_components/pageable/index.js

src/components/Pagination/Pagination.js
```diff
@@ -0,0 +1,90 @@
...
+          </Button>,
+        ] }
+      />
+    );
+  }
+}
+
+Pagination.defaultProps = {
+  currentPage: 1,
+  onChange: noop,
```
circAssimilate: Whoa... I haven't seen lodash `noop` before? I typically tend to use `() => {}` for the default value, which does the same and is a bit easier to parse. What do you think?

src/components/Pagination/Pagination.js
```diff
@@ -0,0 +1,90 @@
...
+  totalPages: 10,
+};
+
+Pagination.propTypes = {
+  /** Sets the current page number */
+  currentPage: PropTypes.number.isRequired,
+  /** Function that is called when a pagination Link is clicked */
+  onChange: PropTypes.func,
+  /** Hook for automated JavaScript tests */
+  testSection: PropTypes.string,
```
circAssimilate: Can we add a default prop of an empty string for this?

src/components/Pagination/Pagination.js
```diff
@@ -0,0 +1,90 @@
...
+            Last
+          </Button>,
+        ] }
+      />
+    );
+  }
+}
+
+Pagination.defaultProps = {
+  currentPage: 1,
```
circAssimilate: Looks like currentPage is required below. Should it be required or not required with a default of 1?

src/components/Pagination/Pagination.story.js
```diff
@@ -0,0 +1,30 @@
...
+  .addDecorator(story => (
+    <div id="root-preview">
+      {story()}
+    </div>
+  ));
+
+stories
+  .add('default', withInfo()(() => {
+    return (
+      <Pagination
```
circAssimilate: I wonder if we could wrap this in a stateful component to simulate the next and last buttons a little more, but the logs you added are nice at least and it's probably better without.

pauloptimizely: I originally wanted to migrate the `pagable` hoc from the monolith, but I decided to hold off on creating a stateful wrapper, b/c I'm still not sure if that wrapper should be a connected component or stateful component. Once this component is integrated in to PM, i'll into furthering this abstracting.

src/components/Pagination/Pagination.js
pauloptimizely: I'll remove this default... i've never thought about this, but I suspect it's best if required props not have a default value

src/components/Pagination/Pagination.js
pauloptimizely: I can remember why I started using `noop`. I think was trying to be super strict on code semantics. Personally, I find `noop` easier to parse b/c i use it so often, but I'm open switching to `() =>  {}` if it's more consist w/ OUI coding style

pauloptimizely:
@circAssimilate I'm renaming this component from `Pagination` to `Pager`. The `Pagination` component found in MGMT diffs quite a bit from this design, so i'm going to create another PR for creating that `Pagination` component.

But since `Pager` can be used in the monolith, I propose we still add it to OUI.

kwalker3690:
@pauloptimizely could we chat with design about whether MGMT could use this version? It'd be really nice to visually consolidate here

pauloptimizely:
@kwalker3690 I'll ping design, I agree consolidating this would be ideal

pauloptimizely:
@circAssimilate @kwalker3690 Design says that the pagination component should visually match the monolith paginater, so i'm renaming this component back to `Pagination`

src/components/Pagination/Pagination.story.js
circAssimilate: Nice. Good call I think!

src/components/Pagination/Pagination.js
circAssimilate: Hmm. It's not a super strong opinion but I'm all for using vanilla JS when possible. I haven't seen `noop` before this, and am more used to `() => {}`. If you're indifferent, would prefer the latter but I'm ok with what you feel is best.

src/components/Pagination/Pagination.js
```diff
@@ -0,0 +1,88 @@
...
+            isDisabled={ currentPage === totalPages }>
+            Last
+          </Button>,
+        ] }
+      />
+    );
+  }
+}
+
+Pagination.defaultProps = {
```
circAssimilate: nit: one really small thing, but I kind of like props and defaultProps being defined as static entities on the `PureComponent` (or `Component`) class. Ok by me though if it's outside, but I think it's at the least easier to read through if the defaultProps come after the prop definitions.

pauloptimizely: I'll update `noop` to `() => {}` and good call on static props 👍

PR #1054, opened 2018-10-11
Migrating Toolbar component from optimizely.git

kwalker3690:
**Summary**
Optimizely.git has a [toolbar component](https://github.com/optimizely/optimizely/blob/master/src/www/frontend/src/js/react_components/toolbar/index.js) that we need to use in MGM. This diff migrates the toolbar over to OUI. I've also migrated the [scss from rebrand2018](https://github.com/optimizely/rebrand2018/blob/master/src/scss/v2/partials/objects/_toolbar.scss) that styles this toolbar. I had to make some small changes to the React code to make it work in the context of OUI, but no functionality change. I made no changes to the SCSS (besides removing commented out code)


kwalker3690:
@jamesopti sure thing.

Also want to note that I tested this in optimizely.git and the styling looked as expected for the toolbar

PR #1056, opened 2018-10-15
Add mousedown event handler to BlockList.Item

pauloptimizely:
## Summary

I'm creating a new component that requires the `BlockList.Item` component to have a mousedown handler.

The reason for using `onMouseDown` instead of `onClick` is there's a `blur` event (in my component) that triggers before a user clicks on a `BlockList.Item`. We want to handle the click event before our blur handler runs. We can achieve this ordering, by leveraging the browser's event precedence, which looks like this: 
`mousedown > blur > click`.

This PR is similar to https://github.com/optimizely/oui/pull/1026

## Test Plan

Run the test suite.
```
npm test
```


pauloptimizely:
@spham1412 good call on writing tests!! I ended up finding a bug

PR #1061, opened 2018-10-18
feat(FilterPicker): Add FilterPicker Render Component

circAssimilate:
**UPDATES:**
- Created FilterPicker Render Component, a Render Component uses a combination of existing OUI components
- Created story for FilterPicker
- Full set of unit tests for FilterPicker

**ABOUT:**
**_FilterPicker_** is meant to be a replacement for `optimizely.git`'s `react_components/entity_selector` and `react_components/filter_list`.

This component only aims to abstract commonly used display and filter functions, there should be no limit to how extensible it is due to it's [render props](https://reactjs.org/docs/render-props.html) design. While it may be tempting to prioritize lines of code over component composition, I think this pattern will prove to be much more helpful. FilterPicker is React _Render Component_ that uses [render props](https://reactjs.org/docs/render-props.html) to abstract away some of the implementation complexity while allowing for maximum extensibility. This pattern was used to avoid a "props explosion" that would come with all the variations we have of this pattern. This approach will give control to the implementer and keep this component lean.

I used StoryBook to prototype and clean this up, so in order to try it out do the following in the OUI repo (it’s really easy!):
* `git checkout derek/entity_selector_filter_list`
* `nvm use` (if you haven’t done this already)
* `yarn install`
* `yarn storybook`
* Once the Storybook Webstack starts up, go to the `FilterList` component

Some things to note:
* Using `react-immutable-proptypes`, FilterPicker can be used with Immutable or POJO datasets  
* Since Immutable isn't imported into OUI, a pretty interesting Immutable.List/Map _mock_ was used in the test file. This is the best way I can think to do this, but please eye it for any immutable discrepancies. I'll also manually QA in `optimizely.git` before releasing
* By default, this component filters `selectedEntityIds` and items that don't match the `filterQuery` to create and return `availableEntities` via the render props function  
* If a `customFilterFn` is used, it will be used in place of the `filterQuery` filter. If `selectedEntityIds` should be kept in, simply do not include that list as a component prop  
* Find out more in the propType docstrings or the "Show Info" in Storybook

**SCREENSHOT:**
image


src/components/FilterPicker/FilterPicker.js
```diff
@@ -0,0 +1,128 @@
...
+		for (const key in content) {
+			if (key) {
+				filterResults = filterResults.concat(isFilterTermInItem(string, content[key]));
+			}
+		}
+	}
+	// If any of the filter results were true, return true
+	return filterResults.some(item => item);
+}
+// END WIP TODO Consider moving the above to a util or module
```
pauloptimizely: optional request: You could create a `utils.js` file in this directory. It might make switching to a module/utils easier in the future, but also ok with leaving this as it for now, since it's not that much code

src/components/FilterPicker/FilterPicker.js
```diff
@@ -0,0 +1,128 @@
...
+  };
+
+  render() {
+    const { allEntities, children, selectedEntityIds } = this.props;
+    const { filterQuery } = this.state;
+
+    const availableEntities = allEntities.filter(entity => !selectedEntityIds.includes(entity.id)).filter((item) => {
+	    //TODO Refine Filter Fn
+	    //TODO Support Immutable JS
+	    const contentToSearch = `${item.name || ''} ${item.description || ''} ${item.category || ''}`;
```
pauloptimizely: I wonder if this should be configurable in some way. Something like:

```js
this.state.searchContent = []

const searchKeys = [].concat(this.state.searchContent, ['name', 'description', 'category']

const contentToSearch = searchKeys.reduce((results, searchKey) => {
    return item.hasOwnP
roperty(searchKey) ? `${results} ${item[searchKey]` : results
}, '')
```
Than we would provide a function for updating the `searchContent` state, wdyt?

CHANGELOG.md
```diff
@@ -11,6 +11,11 @@ This file is similar to the format suggested by [Keep a CHANGELOG](https://githu
 - [Patch] Adjust global border-radius from 8 to 5
 - [Patch] Update Travis.yml to send alerts to #oui-alerts in Slack
 - [Patch] Add hover and focus error states to input
+- [Patch] Added `raw-loader` package for Storybook's webpack config that loads an optional markdown file format (`.mkd`), as plain text until [#3458](https://github.com/storybooks/storybook/issues/3458) is resolved
+- [Feature] Create FilterPicker Render Component
+- [Feature] Add `react-docgen` to Storybook so prop types and descriptions are displayed
```
pauloptimizely: I can't wait to use these! 💯

src/components/ExampleComponent/ExampleComponent.js
```diff
@@ -3,19 +3,19 @@ import PropTypes from 'prop-types';
 
 class ExampleComponent extends Component {
   static propTypes = {
-    /** A string prop */
+    /* A string prop */
```
circAssimilate: Just wanted to clean these up!

package.json
```diff
@@ -13,6 +13,7 @@
     "lodash.debounce": "^4.0.8",
     "lodash.noop": "^3.0.1",
     "react-copy-to-clipboard": "^5.0.1",
+    "react-immutable-proptypes": "^2.1.0",
```
circAssimilate: Used via a recommendation form @quarklemotion!

src/components/ExampleComponent/tests/index.js
```diff
@@ -5,7 +5,7 @@ import { mountToJson } from 'enzyme-to-json';
 import Input from '../../Input';
 import Button from '../../Button';
 
-describe('components/ExampleComponent', () => {
+describe('components/ExampleComponent', function() {
```
circAssimilate: More cleanup, not sure why these were using arrow functions :/

src/components/FilterPicker/FilterPicker.js
circAssimilate: Good call! I ended up updating an existing, unused util I found!

src/components/FilterPicker/FilterPicker.js
circAssimilate: Great call, I think the updated version accounts for this via the `keysToSearch` prop!

src/components/ExampleComponent/ExampleComponent.js
circAssimilate: Whoops. I'll revert these as they're supposed to be this way.

package.json
circAssimilate: [Here's the link](https://docs.google.com/a/optimizely.com/document/d/1-d4o47oPnOt8gbTp04jJ8pka0pGvGFrtxv-jMNufNVs/edit?disco=AAAACSDkDsE) to the Google Doc comment for more context.

src/utils/filter.js
```diff
@@ -1,14 +1,39 @@
-export const isFilterTermInItem = function isFilterTermInItem(stringToFind, contentToSearch) {
-  if (!contentToSearch) {
```
circAssimilate: fwiw, this filter function was not being used - so I commandeered it and made it better.

src/components/FilterPicker/FilterPicker.js
```diff
@@ -0,0 +1,186 @@
...
+      keysToSearch,
+      selectedEntityIds,
+      testSection,
+    } = this.props;
+
+    let availableEntitiesFilter;
+    if (customFilterFn() !== null) {
+      availableEntitiesFilter = customFilterSetup(keysToSearch, filterQuery, customFilterFn);
+    } else {
+      availableEntitiesFilter = defaultFilterSetup(keysToSearch, filterQuery);
```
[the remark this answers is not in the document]
circAssimilate: We chatted about this on Slack and decided to take this route with some slight modifications. Check out the most recent version to see!

PR #1068, opened 2018-10-24
Migrate Dialog component to OUI

kwalker3690:
Copied [Dialog](https://github.com/optimizely/optimizely/blob/master/src/www/frontend/src/js/react_components/dialog) component from `react_components` over to OUI. The only change I made to the component was to use the react-oui-icons version of icon for the help popover.

I've added a storybook entry:
image


src/components/Dialog/Fieldset/index.js
```diff
@@ -0,0 +1,58 @@
...
+const Fieldset = props => (
+  <fieldset data-test-section={ props.testSection }>
+    { props.title && (
+      <div className="gamma push--bottom">
+        { props.title }
+        { props.isOptional && (
+          <span className="oui-label__optional">(Optional)</span>
+        ) }
+        { props.helpIcon && (
+          <OverlayWrapper
```
daverau-optimizely: `<HelpPopover` might give us the same output with a lot simpler JSX. I only say that because OverlayWrapper always gives me codesmell 👃 

Not a blocker. Also I'm happy to take a stab at this later since this version works as is and is ready to go! 😎

kwalker3690: Oh good call! I'll make that update here

PR #1070, opened 2018-10-25
Add test section to blocklist item

kwalker3690:
**Summary**
Simple diff to add a test section to an element in BlockList.Item. This allows us to test the mousedown handler on BlockList.Item.


src/components/BlockList/Item/index.js
```diff
@@ -56,7 +56,8 @@ class Item extends React.Component {
       );
     } else {
       item = (
-        <div className={ commonClasses } onMouseDown={ this.props.onMouseDown }>
+        <div className={ commonClasses } onMouseDown={ this.props.onMouseDown }
+          { ...this.props.testSection && { 'data-test-section': this.props.testSection + '-item' } }>
```
pauloptimizely: eslint wants this to look something like:

```js
let testSection = `${this.props.testSection || ''}-item`;
<div className={ commonClasses } onMouseDown={ this.props.onMouseDown } testSection={ testSection }>
```

PR #1074, opened 2018-10-27
Drau/listgroup component

daverau-optimizely:
This PR adds a new component called ListGroup, an indented list that we'll be using on the Settings page.

I'd like to expand tests later, but wanted to get this component in the app for @fayyazarshad to use ASAP.

screen shot 2018-10-26 at 5 06 50 pm

cc @fayyazarshad This will be ready for you to use on next OUI release, or you can git pull this branch using `npm link` to test.


src/components/ListGroup/ListGroup.story.js
```diff
@@ -0,0 +1,71 @@
...
+    </div>
+  ));
+
+stories
+  .add('ListGroup default', withInfo()(() => {
+    return (
+      <ListGroup
+        title="ListGroup Title"
+        subtitle="ListGroup Subtitle">
+        <ListGroupItem>
```
circAssimilate: Can you actually have an array containing these and map through them here? Like this (example includes `object` knob!):

```
const listItems = object('listItems', [
	{ name: 'ListGroupItem Text 1', value: 'Value 1' },
	{ name: 'ListGroupItem Text 2', value: 'Value 2' },
]);

...
listItems.map(item => (
  <ListGroup
    title={ text("ListGroup Title") }
    subtitle={ text("ListGroup Subtitle") }>
	<ListGroupItem>
        { item.name }
        <p className="muted">{ item.value }</p>
    </ListGroupItem>
  </ListGroup>
));
```

src/components/ListGroup/tests/index.js
```diff
@@ -0,0 +1,28 @@
...
+import {
+  ListGroup,
+  ListGroupItem,
+} from '../index';
+import { shallow } from 'enzyme';
+import { shallowToJson } from 'enzyme-to-json';
+
+describe('components/ListGroup', () => {
+
+  it('render listgroup with correct classes', () => {
```
circAssimilate: While generally I would ask for more tests, I'm really not too sure how much one could actually test in this pattern besides OUI classes being added - and I'm not really sold on the value of that alone. I'd say try and think of a few more tests if you can, but not blocking IMO.

src/components/Input/example/index.js
```diff
@@ -18,7 +18,7 @@ export default [
       <Input
         type="text"
         label="Test"
-        hasError={ true }
+        displayError={ true }
```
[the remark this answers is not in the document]
daverau-optimizely: Ah good catch, this was a stray commit by accident. These examples are from the old docs. I'll leave this since it's now correct but unused at the moment.

src/components/Input/Input.story.js
```diff
@@ -67,6 +67,13 @@ stories
             type="text"
           />
         </fieldset>
+        <fieldset>
+          <Input
+            placeholder="Has an error"
```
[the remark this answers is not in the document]
daverau-optimizely: Derek I'm curious what you think of this: so my approach with Storybook knobs has been to only touch the first example, since they are heavy to add and can clutter up your JSX fast. I've tried to only add knobs for the first example and leave the other examples as more of a set of JSX variations to show other cases or specific configurations.

Mainly this is a time/effort thing. A long time ago I tried to get this working a few times but ran into a few walls, maybe we could pair on this? Then we get all the knobs for free! https://github.com/storybooks/addon-smart-knobs

src/components/ListGroup/ListGroup.story.js
daverau-optimizely: Thanks Derek, I got this working and decided to keep it in since I find it useful for demonstration purposes, but I made a note the knob isn't a prop (called it demoArray and made a note it's just for demo knob purposes only and not a real prop)

daverau-optimizely:
Thanks for the great feedback @circAssimilate — I think this is ready to go.

src/components/Input/Input.story.js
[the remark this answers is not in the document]
circAssimilate: Amazing! Let's definitely pair on that - would be such a win! (and the storysource addon too!)

src/components/ListGroup/index.js
```diff
@@ -0,0 +1,56 @@
...
+  title: PropTypes.string,
+};
+ListGroup.defaultProps = {
+  subtitle: '',
+  testSection: '',
+  title: '',
+};
+
+export const ListGroupItem = ({
+  children,
```
circAssimilate: Can we use `testSection` here and on line #42 or remove it as a default prop?

daverau-optimizely: Whoops, good catch. Fixed by adding `data-test-section={ testSection }` in the latest push.

PR #1077, opened 2018-10-31
Add autoComplete and hidden attributes to Input component

ryn-optimizely:
Adds `autoCompleteOption` and `isHidden` props to Input component, which are used on the `input` element as `autoComplete` and `hidden` attributes. Addresses these warnings:

autocomplete attribute error

The `hidden` attribute was added to provide some context for password managers. See [here](https://www.chromium.org/developers/design-documents/create-amazing-password-forms) for more details.


ryn-optimizely:
I'm going to address this in an updated PR.

PR #1086, opened 2018-11-10
Drau/listgroup padding

daverau-optimizely:
### Summary
This is a fix for @fayyazarshad and will add default padding to the inner `<ListGroup.Item>`s and give you the option to override individual list group items like so: `<ListGroup.Item density="tight">`

### With text
screen shot 2018-11-09 at 5 06 09 pm

### With plain `<Input>s`
screen shot 2018-11-09 at 5 06 15 pm

### DIFF
--- a/src/components/ListGroup/index.js
+++ b/src/components/ListGroup/index.js
@@ -39,13 +39,14 @@ ListGroup.defaultProps = {
 export const ListGroupItem = ({
   children,
   testSection,
-  density='loose',
+  density = 'loose',
 }) => {
+
   let classes = classNames({
     'listgroup__item': true,
     'border--bottom': true,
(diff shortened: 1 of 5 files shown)


src/components/ListGroup/index.js
```diff
@@ -38,16 +39,26 @@ ListGroup.defaultProps = {
 export const ListGroupItem = ({
   children,
   testSection,
+  density = 'loose',
```
quarklemotion: consider moving this default value to `ListGroupItem.defaultProps = { density: 'loose' }` - I think doing so will make the default value show up more prominently in the storybook generated docs

daverau-optimizely: Thanks @quarklemotion, good catch. Fixed!

PR #1087, opened 2018-11-13
Separate SASS colors & variables files

daverau-optimizely:
No code added/removed, only moved around:
- Keep variables file strictly internal (no output)
- Move color output helpers to separate partial file
- By separating concerns we allow for easy exclusion in Preview UI and other use cases when we don't want global color variables for free


daverau-optimizely:
Thanks @kwalker3690!

PR #1089, opened 2018-11-15
Prep for release v40.13.1

mjc1283:


mjc1283:
OK, I was gonna follow up with another commit with no PR that does that, but that's fine.

PR #1091, opened 2018-11-21
Add ConfirmButton component

pauloptimizely:
## Summary

This PR creates a `ConfirmButton` component that renders a cancel/confirm button group.

screen shot 2018-11-19 at 9 38 48 am

## Test Plan

1. Run the test suite `yarn test`
2. Verify it renders correctly in storybook `yarn storybook`

### DIFF
--- a/src/components/ConfirmButton/tests/index.js
+++ b/src/components/ConfirmButton/tests/index.js
@@ -1,0 +1,72 @@
+
+import React from 'react';
+import { mount } from 'enzyme';
+
+import ConfirmButton from '../index';
+
+describe('components/ConfirmButton', function() {
+  let component;
+  let onCancelSpy;
+  let onConfirmSpy;
+  beforeEach(function() {
+    onCancelSpy = jest.fn();
+    onConfirmSpy = jest.fn();
+    component = mount(
+      <ConfirmButton
+        onCancel={ onCancelSpy }
+        onConfirm={ onConfirmSpy }
+      />
+    );
+  });
+
+  afterEach(function() {
+    component.unmount();
+  });
+
+  describe('basic rendering', function() {
+    it('should render ButtonRow', function() {
+      const buttonRowComponent = component.find('ButtonRow');
+      expect(buttonRowComponent.exists()).toBe(true);
+      expect(buttonRowComponent.prop('rightGroup').length).toEqual(2);
+    });
+
+    it('should render cancel button', function() {
+      const cancelButton = component.find('[data-test-section="confirm-button-cancel-button"]');
+      expect(cancelButton.exists()).toBe(true);
+    });
+
+    it('should render confirm button', function() {
+      const confirmButton = component.find('[data-test-section="confirm-button-cancel-button"]');
+      expect(confirmButton.exists()).toBe(true);
+    });
+  });
+
+  describe('editing', function() {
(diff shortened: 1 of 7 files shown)


src/components/ConfirmButton/tests/index.js
```diff
@@ -0,0 +1,72 @@
...
+    });
+
+    it('should not call onConfirm when confirm button is disabled and clicked', function() {
+      component.setProps({ disableConfirm: true });
+      const confirmButton = component.find('[data-test-section="confirm-button-confirm-button"]');
+      confirmButton.simulate('mousedown');
+      expect(onConfirmSpy).toHaveBeenCalledTimes(0);
+    });
+  });
+
```
ryn-optimizely: To be consistent, I think this space should be removed.

src/components/ConfirmButton/tests/index.js
```diff
@@ -0,0 +1,72 @@
...
+import React from 'react';
+import { mount } from 'enzyme';
+
+import ConfirmButton from '../index';
+
+describe('components/ConfirmButton', function() {
+  let component;
+  let onCancelSpy;
+  let onConfirmSpy;
```
ryn-optimizely: For consistency, a space should be added before the `beforeEach` block.

src/components/ConfirmButton/tests/index.js
```diff
@@ -0,0 +1,72 @@
+
+import React from 'react';
+import { mount } from 'enzyme';
+
+import ConfirmButton from '../index';
+
+describe('components/ConfirmButton', function() {
+  let component;
+  let onCancelSpy;
```
ryn-optimizely: Line 8 & 9, should be `const`, since they won't be reassigned.

daverau-optimizely:
[`<ButtonRow>`](http://optimizely.github.io/oui/storybook/?selectedKind=ButtonRow&selectedStory=button%20row%20skip%20step&full=0&addons=1&stories=1&panelRight=1&addonPanel=storybook%2Factions%2Factions-panel) is very similar to this, would ConfirmButton be used in a different way?

src/components/ConfirmButton/tests/index.js
pauloptimizely: The value of `component` and `onCancelSpy` is reassigned each time `beforeEach` is called, this guarantees each test block has a "fresh" context. This style of integration testing is commonly used in the program-management React components. It's really just a simple way to provided things to each `it` block in a deterministic way

pauloptimizely:
@ryn-optimizely there's some quirky things that `ConfirmButton` abstracts.

One major quirk is `mousedown` handlers instead of `click` handlers should be used for button clicks for "editable" components. see https://github.com/optimizely/oui/pull/1026

daverau-optimizely:
Would we prefer to update PM to use the ButtonRow pattern so the HTML is consistent? I'm not seeing the benefit of creating a new but similar pattern. 

Perhaps ButtonRow should be improved to handle click handlers better or is there a better way to use the existing ButtonRow component and add new click handler functionality in a new component, without adding new HTML output that's potentially new tech/design debt?

pauloptimizely:
@daverau-optimizely I'm going to keep this in PM for now!

PR #1092, opened 2018-11-27
Remove npm-run-all dependency that was pulling in malicious flatmap-stream module

quarklemotion:
Currently blocking OUI builds as this malicious module has been removed from the yarn registry.


yarn.lock
```diff
@@ -4439,11 +4417,6 @@ flat-cache@^1.2.1:
     graceful-fs "^4.1.2"
     write "^0.2.1"
 
-flatmap-stream@^0.1.0:
-  version "0.1.1"
```
quarklemotion: here is the removal of the malicious module

PR #1093, opened 2018-11-27
Make Dropdown label a storybook knob

quarklemotion:
Make the Dropdown label a storybook knob so that troubleshooting UI issues with short or long text is easier:

screen shot 2018-11-27 at 3 50 31 pm


storybook/iframe.html
```diff
@@ -54,6 +54,6 @@
 <body>
     <div id="root"></div>
     <div id="error-display"></div>
-<script type="text/javascript" src="static/preview.bea85e2884a4d4b33389.bundle.js"></script></body>
+<script type="text/javascript" src="static/preview.720eca6762426f352d9e.bundle.js"></script></body>
 
 </html>
```
spham1412: I know this file is generated automatically, but anyway we and add an EOF line for it so git won't bark anymore?

PR #1094, opened 2018-12-04
patch(SelectDropdown): Update props correctly to avoid production errors

circAssimilate:
This PR will update `SelectDropdown`'s props to reduce errors in production apps. Fully backwards compatible and no visible changes.


src/components/SelectDropdown/index.js
```diff
@@ -27,7 +27,7 @@ class SelectDropdown extends React.Component {
      * Dropdown items that can be selected from the select dropdown.
      */
     items: PropTypes.arrayOf(PropTypes.shape({
-      activatorLabel: PropTypes.string,
+      activatorLabel: PropTypes.node,
```
circAssimilate: There were a few cases where optimizely.git was passing in a JSX object, so this should allow for that without warning.

PR #1095, opened 2018-12-04
feat (BlockList) Add ability to remove padding from blocklist items

kwalker3690:
**Summary**
I want to display a blocklist of tokens, using the Blocklist component. However, there is padding on Blocklist.Item that makes the tokens indented, which is not ideal acc to design. This diff provides a prop to remove that side padding from the items.
Example:
Before:
image

After:
image

**Test plan**
Added unit test


src/components/BlockList/BlockList.story.js
```diff
@@ -139,5 +140,24 @@ stories
         <BlockList.Item>Lorem ipsum dolor sit amet.</BlockList.Item>
       </BlockList.Category>
     </BlockList>
+  </div>)))
```
circAssimilate: Can you also add knobs to at least one of these stories too, pleaaseeee?

PR #1097, opened 2019-01-03
feat (TagsInput Component): Add Component to edit groups of Tags

jamesopti:
## Summary

Adds a `<TagsInput>` component as a wrapper around https://github.com/olahol/react-tagsinput

tagsinput


src/components/Token/index.js
```diff
@@ -1,10 +1,17 @@
...
 import Icon from 'react-oui-icons';
 import DismissButton from './DismissButton';
 import classNames from 'classnames';
 
+const DARK_BACKGROUND_STYLES = [
+  'tertiary',
+];
+
+const getStylingInfo = style => DARK_BACKGROUND_STYLES.includes(style) ?
+  { font: 'oui-token--font-dark', fill: '#080738' } :
```
kwalker3690: Where does this hex code come from?

src/components/TagsInput/index.js
```diff
@@ -0,0 +1,106 @@
+import React from 'react';
```
kwalker3690: Im curious your thoughts on calling this TagsInput vs something a little more generic like TokensInput - is there ever a case where Tags would not actually be an appropriate name for the use case?

src/components/TagsInput/index.js
```diff
@@ -0,0 +1,106 @@
...
+ * @param {Object} props - Properties passed to component
+ * @returns {ReactElement}
+ */
+export const TagsInput = ({tags, addKeys, placeholder, onChange}) => {
+  function renderTag(componentProps) {
+    const { tag, key, onRemove } = componentProps;
+    function onDismiss() {
+      onRemove(key);
+    }
+    const style = tag.style || undefined;
```
kwalker3690: Shouldn't this be `null` over `undefined`? I've always operated with `null` is what is actually set by an implementer, `undefined` is set if a property is not defined.
Also, would tag.style ever be some other falsy property, so that explicitly setting something is necessary? (Rather than just allowing style to be undefined if tag.style is undefined)

src/components/TagsInput/index.js
```diff
@@ -0,0 +1,106 @@
...
+        onDismiss={ onDismiss }
+        name={ tag.name }
+        style={ style }
+        padding={ false }
+        testSection="token"
+      />
+    );
+  }
+
+  function __onChange(data, changedIndexes) {
```
kwalker3690: Are we back on using this underscore syntax? I remember us talking sometime ago about how we didn't want to bother with this anymore, but I could be mistaken

src/components/TagsInput/index.js
```diff
@@ -0,0 +1,106 @@
...
+    const updatedTags = data.reduce((acc, tag, index) => {
+      if (typeof tag === 'string') {
+        tag = {
+          name: tag,
+          style: null,
+        };
+      }
+
+      const existing = changedIndexes.indexOf(index) < 0;
```
kwalker3690: Can you explain and/or comment on what this is doing? I'm a little confused at first glance without looking into the implementation of ReactTagsInput

src/components/TagsInput/index.js
```diff
@@ -0,0 +1,106 @@
...
+   *   }
+   */
+  tags: PropTypes.arrayOf(PropTypes.object).isRequired,
+};
+
+TagsInput.defaultProps = {
+  /**
+   * Default to space, enter, comma.
+   */
+  addKeys: [9, 13, 32],
```
kwalker3690: Why did you choose to expose this as a prop? I'd think we'd want to have consistency across instances of this

src/components/TagsInput/index.js
jamesopti: Sure!

`data` is a list of all tokens and `changedIndexes` is actually mislabeled.

The intent here is to allow the current iterable's tag if it is an existing tag OR if it is a new unique value.

I will update!

src/components/TagsInput/index.js
jamesopti: MGMT has different behavior than ADA for token entry (spaces are allowed in MGMT, where the spacebar should create a new token for ADA), so this is necessary to allow control over what values tags can have.

src/components/Token/index.js
jamesopti: `fill` is a property passed as an attribute to the underlying `<svg>` (https://github.com/optimizely/react-oui-icons/blob/master/src/Icon/index.js#L90)

As such, we can't use our "very very dark black" SCSS variable with a class here, so I copied it.

Alternative is to just use `black`.

wdyt?

src/components/Token/index.scss
```diff
@@ -20,11 +20,15 @@
 
 .#{$namespace}token-wrap,
 .oui-token-wrap {
```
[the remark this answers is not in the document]
jamesopti: Not used directly in Optimizely.git anywhere.

src/components/Token/index.scss
```diff
@@ -20,11 +20,15 @@
...
-  display: inline-flex;
-  border: 1px solid map-fetch($color, ui, base);
-  background: map-fetch($color, background, light);
+  display: inline-block;
   padding: map-fetch($token, padding);
+}
+
+.#{$namespace}token-padding,
+.oui-token-padding {
```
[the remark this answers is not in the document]
jamesopti: I struggled with the naming for this too.

Jess and I pondered some prop names like

- tokenPadding
- tokenBubble
- outline
- padding

I'll update to use BEM symtax as it is an extension/variant of wrap

src/components/TagsInput/index.js
jamesopti: Went with the latter solution.

src/components/TagsInput/index.js
jamesopti: Updated. Our initial implementation was overcomplicating it (its simpler now).

src/components/TagsInput/index.js
jamesopti: Moved to a single prop `spacesAllowed`

src/components/Token/index.js
jamesopti: Went with `black` for now. Dave may want to refine this later.

src/components/TagsInput/index.js
kwalker3690: Yeah imo the library we're wrapping is an implementation detail, we could swap it out for a different library that doesn't reflect the name Tag - I lean towards TokensInput so that we aren't unnecessarily adding new concepts to OUI without really needing them (ie what if we wanted to add a different tag like concept in the future, would we be bummed we'd used the name up in this instance?)

src/components/TagsInput/index.js
```diff
@@ -71,18 +96,20 @@ export const TagsInput = ({tags, addKeys, placeholder, onChange}) => {
 
 TagsInput.propTypes = {
   /**
-   * Keys which control when a new tag is created.
+   * Whether or not spaces are a valid character in a tag name.
```
kwalker3690: Description is on the wrong propType :)

src/components/Token/index.js
```diff
@@ -98,21 +100,18 @@ Token.propTypes = {
...
-   * Whether or not to display the background padding circle
-   * around the Token
-   */
+  /** Whether or not a well is presented around the token. */
   style: PropTypes.oneOf(['primary', 'secondary', 'tertiary', 'error']),
-  /** Hook for automated JavaScript tests */
+  /** Determines style of token depending on priority level or error */
   testSection: PropTypes.string,
+  /** Hook for automated JavaScript tests */
+  well: PropTypes.bool,
```
kwalker3690: nit: doesn't sound like a bool, would prefer `hasWell`

src/components/Token/index.scss
```diff
@@ -20,15 +20,14 @@
...
+  display: inline-flex;
   padding: map-fetch($token, padding);
-}
 
-.#{$namespace}token-padding,
-.oui-token-padding {
-  background: map-fetch($color, background, light);
-  border-radius: map-fetch($token, border-radius);
-  border: 1px solid map-fetch($color, ui, base);
+  &--well {
```
kwalker3690: This is a great name 👍

src/components/TagsInput/index.js
jessicalc: I like the idea of calling it TokensInput as well.

src/components/TagsInput/index.js
```diff
@@ -0,0 +1,130 @@
...
+import PropTypes from 'prop-types';
+import ReactTagsInput from 'react-tagsinput';
+
+import Token from '../Token';
+
+/**
+ * Key codes which should be interpreted as an
+ * indication to add the current text as a new tag.
+ */
+const ADD_KEYS = [9, 13];
```
jessicalc: are we not going to use commas as well?

jessicalc: and tabs? :)

src/components/TagsInput/index.js
```diff
@@ -0,0 +1,130 @@
...
+ */
+export const TagsInput = ({tags, spacesAllowed, placeholder, onChange}) => {
+  /**
+   * Wrap the layout in a flexed <div>
+   * https://github.com/olahol/react-tagsinput#renderlayout
+   * @param {Array<ReactElement>} tagComponents - Tag components
+   * @param {ReactElement} inputComponent - Input component
+   * @returns {ReactElement}
+   */
+  function defaultRenderLayout(tagComponents, inputComponent) {
```
jessicalc: nit: should this not be called `renderLayout`, since it's not actually the default? 

If we call this `defaultRenderLayout`, then we should call the `renderTag` below `defaultRenderTag`. But I think since it's not default we should just call them `renderTag`/`renderLayout`.

src/components/TagsInput/index.js
```diff
@@ -0,0 +1,130 @@
...
+    }, []);
+
+    onChange(updatedTags);
+  }
+
+  return (
+    <div className="oui-text-input text--left flush">
+      <ReactTagsInput
+        value={ tags }
+        addKeys={ ADD_KEYS.concat(spacesAllowed ? [] : /* Space Key */ 32) }
```
jessicalc: shouldn't this be 
```
addKeys={ ADD_KEYS.concat(spacesAllowed ? /* Space Key */ 32 : []) }
```

src/components/TagsInput/index.js
```diff
@@ -0,0 +1,130 @@
...
+        onlyUnique={ true }
+        addOnPaste={ true }
+        maxTags={ 12 }
+        onChange={ __onChange }
+        renderTag={ renderTag }
+        renderLayout={ defaultRenderLayout }
+        inputProps={{
+          className: 'no-border width--150',
```
jessicalc: could you add `soft-half--ends soft--sides` to give more padding to the input field?

src/components/TokensInput/index.js
```diff
@@ -0,0 +1,145 @@
...
+    }, []);
+
+    onChange(updatedTags);
+  }
+
+  return (
+    <div className="oui-text-input text--left flush">
+      <ReactTagsInput
+        value={ tags }
+        addKeys={ ADD_KEYS.concat(spacesAllowed ? [] : /** SPACE */ 32) }
```
jessicalc: Can you make this `spacesAllowedInToken`? I thought this meant `spacesAllowedAsDelimiter`

src/components/Token/index.scss
```diff
@@ -38,14 +41,26 @@
...
 
-.#{$namespace}token--primary,
-.oui-token--primary {
-  background: map-fetch($token, background-color, primary);
-}
-
-.#{$namespace}token--secondary,
-.oui-token--secondary {
-  background: map-fetch($token, background-color, secondary);
+.#{$namespace}token,
```
jessicalc: NOICE.

src/components/TokensInput/TokensInput.story.js
```diff
@@ -0,0 +1,56 @@
...
+
+stories.add('Default', withInfo()(() => {
+  return <TokensInput onChange={ action('tokens changed') } tags={ SAMPLE_DATA }/>;
+}));
+
+stories.add('spacesAllowed', withInfo()(() => {
+  return (
+    <TokensInput
+      onChange={ action('tokens changed') }
+      tags={ SAMPLE_DATA }
```
jessicalc: for the sample data, how about including an example with spaces? also, please update this to be `spacesAllowedInToken` or something

`tags={ SAMPLE_DATA.concat([{ name: 'this tag has spaces', style: 'secondary' }]) }`

src/components/TokensInput/index.js
```diff
@@ -0,0 +1,145 @@
...
+ * indication to add the current text as a new tag.
+ */
+const ADD_KEYS = [
+  /** TAB   */ 9,
+  /** ENTER */ 13,
+  /** COMMA */ 188,
+];
+
+/**
+ * Tags Input Component
```
jessicalc: Tokens Input Component

src/components/TokensInput/index.js
```diff
@@ -0,0 +1,145 @@
...
+ * @param {Object} props - Properties passed to component
+ * @returns {ReactElement}
+ */
+export const TokensInput = ({tags, spacesAllowed, placeholder, onChange}) => {
```
jessicalc: If we call this `TokensInput`, shouldn't we call the `tags` prop `tokens` instead? 

I'm starting to think we should go back to calling this `TagsInput` instead, since there are many props with `tag`gy names being bandied about.

src/components/TokensInput/tests/index.js
```diff
@@ -0,0 +1,115 @@
...
+
+const SAMPLE_DATA = [
+  { name: 'error', style: 'error' },
+  { name: 'no-style' },
+  { name: 'primary', style: 'primary' },
+  { name: 'secondary', style: 'secondary' },
+  { name: 'tertiary', style: 'tertiary' },
+];
+
+describe('components/TagsInput', () => {
```
jessicalc: If we keep the name as `TokensInput`, this needs to be updated

src/components/TokensInput/tests/index.js
```diff
@@ -0,0 +1,115 @@
+import React from 'react';
+import { cloneDeep } from 'lodash';
+import { mount } from 'enzyme';
+
+import TagsInput from '../index';
```
jessicalc: TagsInput or TokensInput?

PR #1100, opened 2019-01-09
patch(SelectDropdown): SelectDropdown gets disabled when the isDisabled prop is true

Hassan-Khalid-folio3:
### The Following Files include changes for the fix
1. `src/components/Dropdown/index.js`
2. `src/components/SelectDropdown/index.js`

### The Following Storybook file includes changes
1. `src/components/SelectDropdown/SelectDropdown.story.js`

### Test case added in the following file for SelectDropdown isDisabled prop
1. `src/components/SelectDropdown/tests/index.js`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ This file is similar to the format suggested by [Keep a CHANGELOG](https://githu
 
 ## Unreleased
 
+### Fixed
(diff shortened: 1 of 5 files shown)


Hassan-Khalid-folio3:
Hey @spham1412. I have added a test case as requested.

PR #1104, opened 2019-01-17
 docs(FilterPicker): Update Storybook and docs

circAssimilate:
**Summary:**
While at first I thought the `max-scroll--medium` class should go within the `FilterPicker` component itself, I don’t think it really makes sense with the render props pattern it uses. And I think it would be very awkward to allow it to scroll away from the input, which is the only possible way while respecting the render props. We just have to socialize this and ensure people know about the need for it - hence these docs updates.

**Screenshots:**
Without class example:
screen shot 2019-01-16 at 6 15 08 pm

With class example:
screen shot 2019-01-16 at 6 14 58 pm


daverau-optimizely:
Thank you for these changes @circAssimilate 🙌 
Nice quick fix here.

PR #1105, opened 2019-01-18
Update OUI to allow script tag usage

daverau-optimizely:
Update webpack config to export main.js to include UMD bundle for consumption via script tag.


webpack.config.js
```diff
@@ -60,6 +61,11 @@ const config = {
     extensions: ['.json', '.js'],
   },
   plugins: plugins,
+  externals: {
+    react:
```
quarklemotion: I think the linter will complain about this - will likely want you to change it to:
```
    react: 'React',
    'react-dom': 'ReactDOM',
```

PR #1114, opened 2019-01-29
Drau/dashboard filtertoggle

daverau-optimizely:
WIP — DO NOT MERGE, THX!

Hey @kwalker3690, How would you recommend setting the default `tintImage=true` here — what I tried isn't working the way I'm expecting, it's always true even when using using `tintImage=false` in a story.

### DIFF
--- a/src/components/EmptyDashboard/EmptyDashboard.story.js
+++ b/src/components/EmptyDashboard/EmptyDashboard.story.js
@@ -182,24 +182,4 @@ stories
         />
       </div>
     );
-  })
-  .add('No tint', () => {
-    return (
-      <EmptyDashboard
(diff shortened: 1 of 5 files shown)


src/components/EmptyDashboard/EmptyDashboard.story.js
```diff
@@ -182,4 +182,24 @@ stories
         />
       </div>
     );
+  })
+  .add('No tint', () => {
+    return (
+      <EmptyDashboard
+        headline={ text('headline', 'Get started with Optimizely X') }
+        description={ <div>This is a test description about <Link href='http://google.com'>X Feature</Link> with a Google.</div> }
+        imageTint={ false }
```
kwalker3690: This should be `tintImage`, I think that might be your issue

PR #1116, opened 2019-01-31
feat (TokensInput): Handle pasted content better and enable any addKeys

jamesopti:
## Summary

A couple fixes for the `<TokensInput>` component:

*1. Enhancement* - Implement our own `pasteSplit ` method. This is necessary because the native one does not split based on the `addKeys` which have been specified. Its a little silly that we have to do this, but it makes the behavior consistent between typing and pasting.

*2. Bug Fix* - Pass the `maxTags` prop to the `<ReactTagsInput>` component instead of hardcoding 12.

BREAKING CHANGE
*3. Enhancement* - Allow any additional `addKeys` to be passed by the consumer via the  prop `extraAddKeys`. This change removes the specific `allowSpacesInTokens` in favor of `extraAddKeys`, which I discovered can be keyCodes or strings.

## Test Plan
Unit test added for new paste and `extraAddKeys` behavior.

### DIFF
--- a/src/components/TokensInput/TokensInput.story.js
+++ b/src/components/TokensInput/TokensInput.story.js
@@ -31,7 +31,7 @@ class TokensInput extends React.Component {
   onChange = (tokens) => {
     this.setState({ tokens });
     this.props.onChange(tokens);
-  }
+  };
 
   render() {
     const { tokens, onChange, ...rest } = this.props; //eslint-disable-line
@@ -52,12 +52,12 @@ storiesOf('TokensInput', module)
   .add('Default', withInfo()(() => {
     return <TokensInput onChange={ action('tokens changed') } tokens={ SAMPLE_DATA }/>;
   }))
-  .add('spacesAllowedInToken', withInfo()(() => {
+  .add('with extraAddKeys', withInfo()(() => {
     return (
       <TokensInput
         onChange={ action('tokens changed') }
         tokens={ SAMPLE_DATA_WITH_SPACES }
-        spacesAllowedInToken={ true }
+        extraAddKeys={ [' ', '_', ';', '|', '.'] }
       />
     );
   }));
--- a/src/components/TokensInput/index.js
+++ b/src/components/TokensInput/index.js
@@ -64,6 +64,7 @@ export const TokensInput = ({
   /**
    * Render an OUI Token to display each token.
    * https://github.com/olahol/react-tagsinput#rendertag
+   *
    * @param {Object} renderOptions - Values to render token
    * @returns {ReactElement}
    */
@@ -74,6 +75,7 @@ export const TokensInput = ({
     const { name, style } = tag;
     return (
       <Token
+        key={ key }
         isDismissible={ true }
         onDismiss={ onDismiss }
         name={ name }
@@ -87,6 +89,8 @@ export const TokensInput = ({
   /**
    * Handler for when a string is pasted into the input.
    * Splits the string on all the addKeys and extraAddKeys.
+   * https://github.com/olahol/react-tagsinput#pastesplit
+   *
    * @param {string} str - The pasted string.
    * @returns {Array<string>}
    */
@@ -95,8 +99,8 @@ export const TokensInput = ({
       .map(k => k.match)
       .concat(extraAddKeys)
 
-      // Split the string by all our addKeys by joining with an
-      // \n which we will use as the final split operator.
+      // Split the string by all our addKeys by joining with a
+      // \n, which we will use as the final split operator.
       .reduce((acc, value) => acc.split(value).join('\n'), str)
       .split('\n')
       .filter(k => !!k);
@@ -154,7 +158,7 @@ TokensInput.propTypes = {
    * an intent to enter the current string as a new Token.
    * See ADD_KEYS above.
    */
-  extraAddKeys: PropTypes.arrayOf(PropTypes.number),
+  extraAddKeys: PropTypes.arrayOf([PropTypes.number, PropTypes.string]),
 
   maxTags: PropTypes.number,
 
--- a/src/components/TokensInput/index.js
+++ b/src/components/TokensInput/index.js
@@ -16,9 +16,21 @@ import Token from '../Token';
  * indication to add the current text as a new token.
  */
 const ADD_KEYS = [
-  /** TAB   */ 9,
-  /** ENTER */ 13,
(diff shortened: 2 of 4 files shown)


src/components/TokensInput/TokensInput.story.js
```diff
@@ -52,12 +52,12 @@ storiesOf('TokensInput', module)
...
     return <TokensInput onChange={ action('tokens changed') } tokens={ SAMPLE_DATA }/>;
   }))
-  .add('spacesAllowedInToken', withInfo()(() => {
+  .add('with extraAddKeys', withInfo()(() => {
     return (
       <TokensInput
         onChange={ action('tokens changed') }
         tokens={ SAMPLE_DATA_WITH_SPACES }
-        spacesAllowedInToken={ true }
+        extraAddKeys={ [' ', '_', ';', '|', '.'] }
```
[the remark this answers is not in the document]
circAssimilate: Also, how does this work with api_names that have `_`'s in them?

src/components/TokensInput/index.js
```diff
@@ -66,48 +86,80 @@ export const TokensInput = ({tokens, spacesAllowedInToken, placeholder, onChange
...
+   */
+  function pasteSplit(str) {
+    return ADD_KEYS
+      .map(k => k.match)
+      .concat(extraAddKeys)
+
+      // Split the string by all our addKeys by joining with a
+      // \n, which we will use as the final split operator.
+      .reduce((acc, value) => acc.split(value).join('\n'), str)
+      .split('\n')
```
circAssimilate: nit: I find it odd to add and split by `/n` since you never really need a true return character (or do you?).

src/components/TokensInput/index.js
```diff
@@ -66,48 +86,80 @@ export const TokensInput = ({tokens, spacesAllowedInToken, placeholder, onChange
...
+      .filter(k => !!k);
```
circAssimilate: This just filters out empty strings?

src/components/TokensInput/tests/index.js
```diff
@@ -78,35 +78,86 @@ describe('components/TokensInput', () => {
...
-    describe('with spaces allowed', function() {
+    describe('with extra addKeys allowed', function() {
       beforeEach(function() {
         mockOnChange = jest.fn();
         component = mount(
-          <TokensInput onChange={ mockOnChange } tokens={ SAMPLE_DATA } spacesAllowedInToken={ true } />
+          <TokensInput
+            onChange={ mockOnChange }
+            tokens={ SAMPLE_DATA }
+            extraAddKeys={ [' ', '.', '_', ','] }
```
pauloptimizely: In storybook, I can add tokens with semi-colons, is this intended or do we want to restrict add tokens to only this key set?

src/components/TokensInput/index.js
pauloptimizely: :+1: for using functional pipelines

src/components/TokensInput/index.js
```diff
@@ -66,48 +86,80 @@ export const TokensInput = ({tokens, spacesAllowedInToken, placeholder, onChange
...
-      }
-      return acc;
-    }, []);
+    const updatedTokens = allTokens
+      .reduce((acc, token) => {
+        // A newly typed token will be in string form
+        if (typeof token === 'string') {
+          token = { name: token };
+        }
+        if (!acc.find(item => item.name === token.name)) {
```
circAssimilate: Could do `acc.every(item => item.name !== token.name)`. I just found the `!acc.find(`... more difficult to parse.

src/components/TokensInput/tests/index.js
jamesopti: The expected behavior is the core/hardcoded `ADD_KEYS` (tab, enter, comma) plus any `extraAddKeys` provided in props.

The story has more `extraAddKeys` than this test file:

```js
<TokensInput
  onChange={ action('tokens changed') }
  tokens={ SAMPLE_DATA_WITH_SPACES }
  extraAddKeys={ [' ', '_', ';', '|', '.'] }
/>
```

src/components/TokensInput/TokensInput.story.js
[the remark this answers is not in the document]
jamesopti: So this name is following the `react-tagsinput` propName that it extends (`addKeys`).

re: "api_names", this component doesnt really care. It just has 3 mandatory addKeys that submit the input as a token (enter, tab, comma). Then, the consumer can append to that list via `extraAddKeys`

src/components/TokensInput/index.js
jamesopti: So this reduce is working around the limitation that we cant call `.split` by multiple strings.

As such, we use reduce to keep "massaging" the input string by splitting and joining and splitting and joining.

Since `\n` will always be a split character, we can use that as the temporary placeholder while we iterate through all the other split keys.

Does that make sense?

Im totally open to other solutions here. This was just the best way i could think of to do something like this:

`str.split('\t', '\n', ',', ...)` <= What i would like to do (doesnt actually work).

src/components/TokensInput/index.js
jamesopti: yup! if the input contained `one,,,,two`, the split would contain a lot of empty strings, so we filter them out here.

src/components/TokensInput/index.js
jamesopti: Hmmm, in this case, I simply want to know if a token exists with this name already. `find` will get me the first, which is all I care about, and doesnt require iterating through the entire list everytime.

src/components/TokensInput/TokensInput.story.js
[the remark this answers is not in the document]
circAssimilate: Gotcha!

src/components/TokensInput/index.js
circAssimilate: Yea, that makes sense. I guess maybe making it a constant named `SPLIT_CHARACTER` or something would be most preferable to me, but it's good.

src/components/TokensInput/index.js
circAssimilate: I see. Was thinking there had to be an easier / clearer way, but all good.

src/components/TokensInput/index.js
circAssimilate: Oh, good to know that `find()` bails when something is found. I thought it might look through everything anyways.

PR #1120, opened 2019-02-04
Drau/emptystate remove css filter

daverau-optimizely:
Removing this CSS filter to (coincide with updating the empty state SVG source colors)[https://github.com/optimizely/optimizely/pull/9040] to the updated rebrand color palette.

### DIFF
--- a/src/components/EmptyDashboard/index.js
+++ b/src/components/EmptyDashboard/index.js
@@ -29,7 +29,7 @@ const EmptyDashboard = ({
               <object
                 data={ imagePath }
                 className="svg--non-scaling-stroke display--inline"
-                style={{ maxWidth: '450px', maxHeight: '290px', 'filter': 'hue-rotate(33deg) saturate(130%)' }}
+                style={{ maxWidth: '450px', maxHeight: '290px' }}
                 data-test-section={ testSection && `${testSection}-image` }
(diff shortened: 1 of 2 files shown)


src/components/EmptyDashboard/index.js
```diff
@@ -29,7 +29,7 @@ const EmptyDashboard = ({
               <object
                 data={ imagePath }
                 className="svg--non-scaling-stroke display--inline"
-                style={{ maxWidth: '450px', maxHeight: '290px', 'filter': 'hue-rotate(33deg) saturate(130%)' }}
+                style={{ maxWidth: '450px', maxHeight: '290px' }}
```
circAssimilate: I believe these can just be Numbers and it will default to pixels.

```
style={{ maxWidth: 450, maxHeight: 290 }}
```

daverau-optimizely: Ah good point @circAssimilate. I guess my personal preference is to specify the value so it's more clear and readable. If these were 0 I'd definitely be inclined to remove the value since 0% vs 0px vs 0em is all the same.

Is this a code formatting style we utilize anywhere else, if so I'm fine with following that convention? Otherwise I'd be inclined to leave the value in for clarity.

PR #1125, opened 2019-02-20
fix(tokensInput): Tighten spacing around keyword tokens

ayeshakmaz:
Introduced a new prop for Tokens: `hasSnugWrap` which adds a tighter wrap around the Token. This new prop is then used for the tokens inside the TokensInput component

Before:

flat-before

stacked-before

After:

flatafter
stackedafter

Jira
[REC-758](https://optimizely.atlassian.net/browse/REC-758)


jamesopti:
Do we want these changes to apply to all `<Tokens>`? There are many consumers of `Token` that arent the `TokensInput`.

daverau-optimizely:
I feel bad for requesting changes here! Just a code style thing for us at Optimizely.

daverau-optimizely:
Good call James. Ayesha and I talked about making a mini component to address the font size and padding changes and TokenInput would use miniTokens. Also Tokens will be renamed to Pills soon, but that's coming late this quarter or early next.

ayeshakmaz:
@daverau-optimizely @jamesopti Made some updates - now TokensInput uses a smaller wrap on the Tokens. Let me know what you think

PR #1128, opened 2019-02-22
fix(editableInput): Flip pencil icon to point toward text

ayeshakmaz:
Flip the direction of the pencil icon in EditableInput to point toward the text

Before:

screen shot 2019-02-21 at 5 52 16 pm

After:

screen shot 2019-02-21 at 5 52 12 pm

Jira:
[DESIGN-326](https://optimizely.atlassian.net/browse/DESIGN-326)


pauloptimizely:
@spham1412 @ayeshakmaz I’m totally fine with flipping the icon. This component is no longer used by Program Management

daverau-optimizely:
> @spham1412 @ayeshakmaz I’m totally fine with flipping the icon. This component is no longer used by Program Management

Hey @pauloptimizely, ah thanks for the heads up here. What is PM using now? I'm confused by this comment, did you move to another pattern, use a different icon, or drop the pattern all together? I ask because I'm currently still seeing the edit pencil in use in PM, so just wondering what the final state should be.

Thanks!

screen shot 2019-02-22 at 12 06 36 pm

spham1412:
@daverau-optimizely from my convo with @pauloptimizely 

> Making the pencil icon only visible when hovering was a design choice for Program Management, but we’re currently not using the `EditableInput` from OUI. There’s a more abstract `Editable` component in our codebase. There’s two path we could go down:
> 
> 1. I can add the abstract `Editable` component that exists in PM to OUI; and add in a prop call `alwaysShowIcon`.
> 2. Or I add the `alwaysShowIcon` prop to the `EditableInput` in OUI
> 
> If you think you’ll need other components to be editable, outside of just `Input`. I’d be happy to make a PR for option 1

daverau-optimizely:
Ok, just making sure the icon is still being used. We'll proceed with this PR which is only to change the orientation of the icon. 

After this PR lands then we can decide between options #1 or #2. I don't have a strong opinion but I'm leading toward a generic editable component. 

Also I'd be in favor of keeping the PM design decision to only show pencil icon on hover.

PR #1132, opened 2019-03-01
Update arrowsinline and dropdown

daverau-optimizely:
- Set arrowsInline default direction to down
- Fix warning on invalid prop type for Dropdown (remove true option for arrowIcon)
- Update snapshots and stories

### DIFF
--- a/src/components/ArrowsInline/index.js
+++ b/src/components/ArrowsInline/index.js
@@ -26,6 +26,10 @@ const ArrowsInline = ({
   );
 };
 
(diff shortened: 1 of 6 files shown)


src/components/ArrowsInline/index.js
```diff
@@ -26,6 +26,10 @@ const ArrowsInline = ({
   );
 };
 
+ArrowsInline.defaultProps = {
```
circAssimilate: Please put these below the propType definitions.

PR #1133, opened 2019-03-08
patch(Select): Fixed Select export by exporting it in src/main.js

Hassan-Khalid-folio3:
# Summary
Select component exists in the OUI code base but was not being exported. This PR fixes this issue by exporting it inside `src/main.js` 

### The Following File include changes for the fix
1. `src/main.js`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ This file is similar to the format suggested by [Keep a CHANGELOG](https://githu
 ## Unreleased
 - [Patch] Update <Steps> styling to have an outline for the Active step and filled circles for Complete steps
 
+### Fixed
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ This file is similar to the format suggested by [Keep a CHANGELOG](https://githu
 ## Unreleased
 - [Patch] Update <Steps> styling to have an outline for the Active step and filled circles for Complete steps
 
+### Fixed
+- [Patch] Fixed <Select> export by exporting it in src/main.js
```
spham1412: Shouldn't this be put under `## Unreleased`? I don't recall we have a separate `### Fixed` category.

PR #1134, opened 2019-03-15
feat(dialog/sheet): Update Dialog to be small window, add Sheet component

ayeshakmaz:
- Added new component for Sheet
- Updated styling of Dialog to be small contained window

Screen Shot 2019-04-08 at 4 54 02 PM

Screen Shot 2019-04-01 at 4 51 31 PM

- Added CloseButton cmp for the shared close button b/w Dialog and Sheet

- Added variables for z-index for Dialog and Overlay, and footer color

TODO:
- [x] Write tests
- [x] Update changelog


src/components/Sheet/index.scss
```diff
@@ -0,0 +1,60 @@
+$sheet-width: map-fetch($dialog, width, wide);
```
daverau-optimizely: Should we move this to oui_variables.scss under $dialog?
```
$dialog: (
  width:(
    base                     : 700px,
    narrow                   : 500px,
    wide                     : 900px
  ),
  footer:(
    color                    : #f7f7f7,
  )
);
```

src/components/Dialog/index.js
```diff
@@ -1,13 +1,56 @@
...
-  Title,
-  Wrapper,
+const Dialog = props => (
+  <div className="oui-dialog__wrapper">
+    <div className="oui-dialog__overlay" />
+    <div
+      data-ui-component={ true }
+      className="oui-dialog overflow-y--auto"
+      data-test-section={ props.testSection }>
+      {props.hasCloseButton && <button className="oui-dialog__close" onClick={ props.onClose }>
```
daverau-optimizely: Consider a new CloseButton component

src/oui/partials/sass/_mixins.scss
```diff
@@ -322,3 +322,25 @@
     }
   }
 }
+
+@mixin dialog-close {
```
daverau-optimizely: I'll leave this up to you, but perhaps class vs mixin since we're kind of phasing out mixins as they can be hard to search for and reason about.

src/components/Sheet/index.scss
ayeshakmaz: Just realized I only added this locally in this file to keep consistent b/w the width of the sheet and the width of the footer, so we don't need it as an actual variable (it already points to dialog-width-wide)

src/oui/partials/sass/_mixins.scss
ayeshakmaz: Deleted in favor of the CloseButton cmp

src/components/Dialog/Dialog.story.js
```diff
@@ -1,66 +1,183 @@
...
+        <Dialog
+          title={ text('title', 'Welcome to this Feature') }
+          hasCloseButton={ boolean('hasCloseButton', true) }
+          onClose={ action('Dialog was closed') }
+          footerButtonContent={ [
+            <Button style="highlight" key={ 0 } onClick={ noop }>
+              Get Started
+            </Button>,
+          ] }>
+          <p>Let's get you up and running with our feature.</p>
```
circAssimilate: nit: I think it's lame github barfs on this but you can use `&apos;` here to make this look better.

src/components/Dialog/index.js
```diff
@@ -1,13 +1,70 @@
...
+
+Dialog.propTypes = {
+  /**
+   * The body of the dialog to request minimal information from the user.
+   */
+  children: PropTypes.node.isRequired,
+  /**
+   * Array of buttons used in the footer of the dialog.
+   */
+  footerButtonContent: PropTypes.array.isRequired,
```
circAssimilate: `footerButtonContent` leads me to believe this would be `PropTypes.node`. Maybe this should be `footerButtonGroup` or `footerButtonList`? Those are kinda bad suggestions, but hoping it may avoid some implementation errors.

src/components/Dialog/index.js
```diff
@@ -1,13 +1,70 @@
...
+    }
+  },
+  /**
+   * A subtitle for the dialog.
+   */
+  subtitle: PropTypes.string,
+  /**
+   * Identifier used to create data-test-section attributes for testing.
+   */
+  testSection: PropTypes.string,
```
circAssimilate: Can we add a default prop of `''` for this?

src/components/Dialog/index.js
```diff
@@ -1,13 +1,70 @@
...
+    if ((props['hasCloseButton'] !== false && (props[propName] === undefined || typeof (props[propName]) !== 'function'))) {
+      return new Error(
+        'Please provide an onClose function'
+      );
+    }
+  },
+  /**
+   * A subtitle for the dialog.
+   */
+  subtitle: PropTypes.string,
```
circAssimilate: Can we add a default prop of `''` for this?

src/components/Dialog/index.js
```diff
@@ -1,13 +1,70 @@
...
+  footerButtonContent: PropTypes.array.isRequired,
+  /**
+   *  Used to determine if the dialog should have a close button.
+   */
+  hasCloseButton: PropTypes.bool,
+  // eslint-disable-next-line valid-jsdoc
+  /**
+   *  Function to perform when the dialog is closed.
+   */
+  onClose: function(props, propName) {
```
circAssimilate: Nice custom prop! However, I'd recommend making this `PropTypes.function` and providing a default prop of `() => {}`. The anonymous function will be a noop and won't decrease performance!

src/components/Dialog/index.scss
```diff
@@ -1,45 +1,58 @@
...
-///     <div class="#{$namespace}dialog__close">
-///       <svg class="#{$namespace}icon">
-///         <use xlink:href="#close-16"></use>
-///       </svg>
-///     </div>
-///   </div>
-////
-
-.#{$namespace}dialog,
 .oui-dialog {
```
circAssimilate: Are we no longer doing the `.#{$namespace}dialog,` in scss definitions? I actually never got it to work, so am fine with that just curious.

src/components/Sheet/index.js
```diff
@@ -0,0 +1,70 @@
...
+        {props.children}
+      </div>
+      <footer className="oui-sheet__footer">
+        <ButtonRow rightGroup={ props.footerButtonContent }/>
+      </footer>
+    </div>
+  </div>
+);
+
+Sheet.propTypes = {
```
circAssimilate: See all the props comments on `Dialog/index.js`, but still think we could consider combining these components!

src/components/Sheet/index.js
```diff
@@ -0,0 +1,70 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import ButtonRow from '../ButtonRow';
+import CloseButton from '../CloseButton';
+
+const Sheet = props => (
```
circAssimilate: I know some classes are different, but this definitely feels too close to the Dialog component to be it's own thing. What if the Dialog component had a `type` prop that could be of type `default` or `sheet` (or some other enumerator combo)? If these will continue to diverge then I could see it, but given the similarities it may not hurt to keep these together to avoid divergence and the need to update both.

ayeshakmaz:
> The `<Dialog>` updates here are a breaking change. Do we know how much this component is used across experimentation and mgmt?

@jamesopti It's used in about 12 places total (10 in experimentation, 2 in mgmt). They would all be easy to convert, so I was planning to convert those and simultaneously merge that work when merging this PR. 

OR I can keep the old code in this PR, merge it, convert the others, then come back and remove the deprecated code. 

Let me know which you think would be better.

src/components/Dialog/index.js
ayeshakmaz: Oh totally, good call - I think `footerButtonList` works well

src/components/Dialog/index.js
ayeshakmaz: If I do that, the space for the subtitle will always be reserved, even when there's no content. Is there a reason for adding empty string default?

circAssimilate: The main goal is to avoid a proptype warning and always provide a value of the type specified. Since the code below is used, an empty string will be falsy and not render a subtitle I believe!

```
{props.subtitle && <p className="push--top flush--bottom">{props.subtitle}</p>}
```

circAssimilate:
>OR I can keep the old code in this PR, merge it, convert the others, then come back and remove the deprecated code.
I would recommend this, just seems like the safest approach if someone in the monolith were to bump OUI while you were working on removing the implementations.

src/components/Dialog/index.js
ayeshakmaz: Doh - don't know why I thought empty would be truthy, totally right

What PropType warning would get introduced without it though?

src/components/Dialog/index.scss
ayeshakmaz: Yeah chatted with @daverau-optimizely and we no longer need them for new components

src/components/Sheet/index.js
ayeshakmaz: So, I totally agree that code-wise, they are very similar, but we actually decided to make them separate to avoid engineers using the wrong one, since they have very different use cases. We figured splitting them out will help inform people when to use which one

ayeshakmaz:
@jamesopti @circAssimilate Just made updates to keep the old components and export the new Dialog with the temporary name of `DialogNew`, let me know how it looks

This way new Dialogs can be made with `DialogNew`, old ones will go from using Dialog.Wrapper, etc. to `DialogNew`, and then once everything is converted, `DialogNew` will just become `Dialog`

Also - fun fact: most Dialogs in frontend are using a copy in `react-components`, so they're not even using OUI

src/components/Dialog/index.js
circAssimilate: From [React's docs](https://reactjs.org/docs/typechecking-with-proptypes.html), it says:

>PropTypes exports a range of validators that can be used to make sure the data you receive is valid. In this example, we’re using PropTypes.string. When an invalid value is provided for a prop, a warning will be shown in the JavaScript console.

Here's an example of what that could look like and something I think it would be great to avoid if possible:

image

src/components/Sheet/index.js
circAssimilate: Interesting! I think going for _DRY_ can sometimes be a scam, but I think a default prop could achieve the same thing. But this is just a nit and obviously you've considered both. So nothing I think I think is worth blocking on.

src/components/Dialog/index.js
ayeshakmaz: But that only happens when it's marked as required right? testSection and subtitle aren't required

src/components/Sheet/index.js
```diff
@@ -34,15 +33,20 @@ Sheet.propTypes = {
   /**
    * Array of buttons used in the footer of the sheet.
    */
-  footerButtonContent: PropTypes.array.isRequired,
+  footerButtonList: PropTypes.array.isRequired,
   /**
    *  Used to determine if the sheet should have a close button.
    */
   hasCloseButton: PropTypes.bool,
+  // eslint-disable-next-line valid-jsdoc
```
circAssimilate: I think you can remove this now!

src/components/Dialog/index.js
circAssimilate: Oh, that may be true! I at least think our linter wants us to always define. I know the monolith props always does at least. Taking inspiration from the below bullet on https://github.com/airbnb/javascript/tree/master/react#props, which I believe we try to use for our linters.

>Always define explicit defaultProps for all non-required props

I guess we use `react/require-default-props` in the Monolith but don't in OUI (I think we should though!). Just helps us get closer to more strongly typing things and avoiding confusing bugs from small things (although probably unlikely still possible). Hope that helps clarify, thanks for pointing out the big about the warning being wrong!

ayeshakmaz: Ahhhh I see! Yeah, we should probably align our linters

src/main.js
```diff
@@ -11,6 +11,7 @@ export { default as Checkbox } from './components/Checkbox';
 export { default as Code } from './components/Code';
 export { default as CopyButton } from './components/CopyButton';
 export { default as Dialog } from './components/Dialog';
+export { DialogNew as DialogNew } from './components/Dialog';
```
jamesopti: Can you add a comment here about the transition period?

ayeshakmaz: Sure thing

ayeshakmaz:
@jamesopti Yeah, once this is merged in to both the experimentation and mgmt repos, I'll go in and first update the Dialogs actually using OUI's Dialog, then come back to OUI and make the conversion from `DialogNew` -> `Dialog`

PR #1135, opened 2019-03-19
feat(button): Add new button that disables and adds spinner

ayeshakmaz:
- Added a new button that disables and adds a spinner by passing in prop isLoading
  - Button also changes text to 'Processing' or props.loadingButtonText
Screen Shot 2019-03-22 at 3 31 14 PM
Screen Shot 2019-03-22 at 3 31 19 PM

- Added new Story examples

ToDo:

- [x] Add tests
- [x] Update current tests
- [x] Update changlog

Jira:
[APPX-130](https://optimizely.atlassian.net/browse/APPX-130)


src/components/Button/tests/index.js
```diff
@@ -1,9 +1,45 @@
...
+    expect(button).toMatchSnapshot();
+  });
+
+  test('should disable the button after click if disableAfterClick is true', () => {
+    const { container } = render(
+      <Button disableAfterClick={ true } >Create Campaign</Button>
+    );
+    const button = container.firstChild;
+    fireEvent.click(button);
+    expect(button).toMatchSnapshot();
```
pauloptimizely: Instead of checking the the snapshot, we should check that the `isDisabled` prop is false. We've encountered some issues in the past with snapshot testings.

The check would look something like:

```js
expect(button.prop('isDisabled')).toBe(true);
```

src/components/Button/Button.story.js
```diff
@@ -113,4 +117,9 @@ stories
         <Link>Say Hi</Link>
       </Button>
     );
-  }));
+  }))
+  .add('Button that disables on click', () => {
+    return (
+      <Button style="highlight" disableAfterClick={ true } onClick={ testButtonClick }>Create Campaign</Button>
```
pauloptimizely: For onClick events, use storybooks `actions` handler. This handle will log events to the storybook "knobs" tab

https://storybook.js.org/docs/addons/introduction/

Example:

```js
import { action } from '@storybook/addon-actions';

<Button onClick={action('clicked')}>Hello Button</Button>
```

src/components/Button/tests/index.js
```diff
@@ -1,23 +1,68 @@
...
+    );
+    expect(component.getDOMNode().attributes.getNamedItem('disabled')).toBe(null);
+  });
+
+  it('should disable the button after click if disableAfterClick is true', () => {
+    const component = mount(
+      <Button disableAfterClick={ true } >Create Campaign</Button>
+    );
+    component.find('button').simulate('click');
+    expect(component.getDOMNode().attributes.getNamedItem('disabled').value).not.toBe(null);
```
pauloptimizely: This totally works for me 👍

ayeshakmaz:
@pauloptimizely Can you take one last look? Added one last change, plus the changelog that I forgot to update. Thanks!

pauloptimizely:
@ayeshakmaz yep, i'll take a look

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This file is similar to the format suggested by [Keep a CHANGELOG](https://github.com/olivierlacan/keep-a-changelog).
 
 ## Unreleased
+- [Feature] Added a variation to <Button> component that disables the button after click using `disableAfterClick` prop
```
pauloptimizely: Good call! I always forget to update the changelog

yarn.lock
```diff
@@ -8987,6 +9046,16 @@ react-treebeard@^2.1.0:
     prop-types "^15.6.2"
     schedule "^0.5.0"
 
+react@^16.4.0:
```
jamesopti: This means we now have non-overlapping react version requirements, so 2 versions must be bundled.

Can you run `yarn-deduplicate yarn.lock` and see if its a yarn issue?

If not, we should figure out what is still requiring the old versions and upgrade those packages too.

ayeshakmaz:
@jamesopti Definitely, my plan is to get versions of both frontend repos running locally with this version of OUI and test to make sure thinks work properly/as expected. We can also discuss more in FeCoP

ayeshakmaz:
FYI @circAssimilate Here is the new version

src/components/Button/Button.story.js
```diff
@@ -37,6 +37,7 @@ stories
         }, 'highlight') }
         isDisabled={ boolean('isDisabled', false) }
         isActive={ boolean('isActive', false) }
+        isLoading={ boolean('isDisabled', false) }
```
circAssimilate: Can you update it to be isLoading for the knob?

src/components/Button/index.js
```diff
@@ -51,15 +55,17 @@ const Button = ({
...
       type={ type }
       onBlur={ onBlur }
       onClick={ onClick }
       onMouseDown={ onMouseDown }
       data-test-section={ testSection }
       aria-label={ ariaLabel }
+      aria-live="polite"
       ref={ buttonRef }>
-      { children }
+      {isLoading && <Spinner data-test-section="test-spinner" size="tiny"/>}
```
circAssimilate: nit: should it be “button-spinner”?

src/components/Button/index.js
```diff
@@ -119,9 +129,12 @@ Button.propTypes = {
 
 Button.defaultProps = {
   isLink: false,
+  isDisabled: false,
+  isSubmit: false,
```
circAssimilate: Can we add false for isLoading and an empty string for loadingButtonText.

src/components/Button/index.js
```diff
@@ -29,6 +32,7 @@ const Button = ({
     [`oui-button--${size}`]: size,
     [`oui-button--${width}`]: width,
     ['is-active']: isActive,
+    'oui-button--loading': isLoading,
```
circAssimilate: nit: could we remove the brackets from the others that don’t need it or add them here?

src/components/Button/index.js
```diff
@@ -80,8 +86,12 @@ Button.propTypes = {
   isDisabled: PropTypes.bool,
   /** Changes the button to a div for insertion within a Link component */
   isLink: PropTypes.bool,
+  /** When true, adds a spinner to the button and disables the button */
+  isLoading: PropTypes.bool,
   /** Make the button act as a submit button */
   isSubmit: PropTypes.bool,
+  /** When the button adds a spinner, it displays this text */
+  loadingButtonText: PropTypes.string,
```
circAssimilate: Also, nit, but I think it could just be loadingText since it’s happening within the button

yarn.lock
```diff
@@ -1090,6 +1090,15 @@
   resolved "https://registry.yarnpkg.com/@icons/material/-/material-0.2.4.tgz#e90c9f71768b3736e76d7dd6783fc6c2afa88bc8"
   integrity sha512-QPcGmICAPbGLGb6F/yNf/KzKqvFx8z5qx3D1yFqVAjoFmXK35EgyW+cJ57Te3CNsmzblwtzakLGFqHPqrfb4Tw==
 
+"@jest/types@^24.8.0":
```
circAssimilate: Following up on our FE CoP discussion today, I think trimming yarn clean, yarn install, and yard deduplicate should be used here. Also I saw a new import is added but is it needed?  Or seeing react-testing-library imported anywhere but may be missing it!

package.json
```diff
@@ -84,6 +84,7 @@
     "react-ga": "^2.1.2",
     "react-router": "^3.0.1",
     "react-test-renderer": "^16.0.0",
+    "react-testing-library": "^6.0.1",
```
circAssimilate: Not seeing this imported. Is this still needed?

src/components/Button/Button.story.js
ayeshakmaz: 🤦🏽‍♀️

src/components/Button/index.js
ayeshakmaz: Made a more sweeping style change here (to my personal preference), let me know if you prefer the bracket way

yarn.lock
ayeshakmaz: Good call, I think the yarn.lock is properly tidied now

src/components/Button/index.js
```diff
@@ -51,15 +55,17 @@ const Button = ({
     <button
       data-oui-component={ true }
       className={ buttonClassNames }
-      disabled={ isDisabled }
+      disabled={ isDisabled || isLoading }
```
circAssimilate: I have one more request on this that I think will make our buttons better and less able to be hacked around. Could we wrap `onClick` in the following to take extra steps to ensure the onClick won't be called?

```
const handleOnClick = () => {
  if (isDisabled || isLoading) {
    return;
  }
  onClick();
};
// ...
<button
      data-oui-component={ true }
      className={ buttonClassNames }
      disabled={ isDisabled || isLoading }
      onClick={ handleOnClick }
```

I know that's overkill, but I've definitely removed the "disabled" class from the dom before and got access to some features I shouldn't have. I believe this (and some much needed FE and BE permissions auditing) will go a long way to ensure customers and collaborators get the right experience.

ayeshakmaz:
@circAssimilate Made those updates and reset the yarn.lock, but now the FilterPicker tests are failing their `onClick` tests because of the event.stopPropagation call in FilterPickerListItem. If you don't mind, could you investigate and carry this over the finish line while I'm out? Thanks for all your awesome feedback!

src/components/Button/index.js
```diff
@@ -23,16 +26,25 @@ const Button = ({
...
+      [`oui-button--${style}`]: style,
+      [`oui-button--${size}`]: size,
+      [`oui-button--${width}`]: width,
+      'is-active': isActive,
+      'oui-button--loading': isLoading,
+    });
 
   const type = isSubmit ? 'submit' : 'button';
 
+  function handleOnClick() {
```
circAssimilate: Ah! I'm a fool! It should be:

```
function handleOnClick(event) {
    if (isDisabled || isLoading) {
      return;
    }
    onClick(event);
}
```

We of course need to pass the event through. That should fix your `FilterPickerListItem` issues!

circAssimilate:
I can handle the merging from here, Ayesha! Looping in @daverau-optimizely for the UI Eng approval.

src/components/Button/index.js
```diff
@@ -119,9 +137,14 @@ Button.propTypes = {
 
 Button.defaultProps = {
   isLink: false,
+  isLoading: false,
```
jamesopti: Would be sweet if I could pass a boolean OR a promise to `isLoading`, but we can explore that in another PR.

circAssimilate: Yea, @jamesopti, agreed! Ayesha and I were talking about that and also decided this could be a good addition if needed down the road.

PR #1136, opened 2019-03-22
patch(Input): Fixed the isRequired prop on <Input> component which wasn't working as expected.

zashraf1985:
# Summary
isRequired prop on Input does not add `*` to the label . Fixed it to add a *.

# Test Plan
1. Thoroughly tested manually.
2. Added a unit test to verify the isRequired prop functionality.

### DIFF
--- a/src/components/Input/Input.story.js
+++ b/src/components/Input/Input.story.js
@@ -33,7 +33,7 @@ stories
         onBlur={ action('on blur') }
         onKeyDown={ action('on key press') }
         placeholder={ text('placeholder', 'just a placeholder') }
-        required={ boolean('required', false) }
+        isRequired={ boolean('isRequired', false) }
         type={ select('type', ['text', 'password', 'number', 'date'], 'text') }
       />);
   })).add('Inputs', () => {
--- a/src/components/Input/Input.story.js
+++ b/src/components/Input/Input.story.js
@@ -49,6 +49,15 @@ stories
         </fieldset>
         <fieldset>
           <Input
+            label="Label for required field"
+            isRequired={ true }
+            note="A short description or note about this field."
+            placeholder="Just a placeholder"
+            type="text"
+          />
+        </fieldset>
+        <fieldset>
+          <Input
             label="Field label"
             placeholder="Just a placeholder"
             type="text"
--- a/src/components/Input/index.js
+++ b/src/components/Input/index.js
@@ -93,6 +93,7 @@ class Input extends React.Component {
       testSection,
       isOptional,
       note,
+      isRequired,
     } = this.props;
 
     if (label) {
@@ -103,6 +104,7 @@ class Input extends React.Component {
           <Label testSection={ testSection && testSection + '-label' }>
             <div className="oui-label">
               { label }
+              { isRequired && <span className="oui-label--required"></span> }
               { isOptional && <span className="oui-label__optional">(Optional)</span> }
             </div>
             { this.renderInput(this.props) }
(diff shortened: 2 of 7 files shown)


src/components/Input/index.js
```diff
@@ -103,6 +104,12 @@ class Input extends React.Component {
           <Label testSection={ testSection && testSection + '-label' }>
             <div className="oui-label">
               { label }
+              { isRequired &&
```
ayeshakmaz: Could we utilize the isRequired and isOptional props that already exist on `<Label>` instead of hard coding it here?

So essentially more like:
```
<Label testSection={ testSection && testSection + '-label' } isRequired={isRequired} isOptional={isOptional}>
{label}
</Label>
```

zashraf1985: Yeah. Totally makes sense. It didn't come to my mind. Makes me feel so dumb.

ayeshakmaz: No I totally get where you were coming from, especially since the `isOptional` piece was already hardcoded in there :)

zashraf1985: Haha! Ya, I just went with the flow instead of thoroughly thinking it through. Fixed it now

src/components/Input/Input.story.js
```diff
@@ -33,7 +33,7 @@ stories
         onBlur={ action('on blur') }
         onKeyDown={ action('on key press') }
         placeholder={ text('placeholder', 'just a placeholder') }
-        required={ boolean('required', false) }
+        isRequired={ boolean('isRequired', false) }
```
ayeshakmaz: Nice prop name update 👍

src/components/Input/Input.story.js
```diff
@@ -33,7 +33,7 @@ stories
...
         type={ select('type', ['text', 'password', 'number', 'date'], 'text') }
       />);
   })).add('Inputs', () => {
```
ayeshakmaz: It might be nice to add another Input example here where a label is defined and  `isRequired` is true, since the knobs example doesn't have a label by default

src/components/Input/index.js
```diff
@@ -93,20 +93,21 @@ class Input extends React.Component {
...
           className={ classNames({'oui-form-bad-news': displayError}) }>
-          <Label testSection={ testSection && testSection + '-label' }>
-            <div className="oui-label">
-              { label }
-              { isOptional && <span className="oui-label__optional">(Optional)</span> }
-            </div>
-            { this.renderInput(this.props) }
+          <Label
+            testSection={ testSection && testSection + '-label' }
+            isRequired={ isRequired }
```
ayeshakmaz: Awesome

src/components/Input/index.js
```diff
@@ -145,7 +146,7 @@ Input.propTypes = {
   },
   /** Prevents input from being modified but doesn't appear disabled */
   isReadOnly: PropTypes.bool,
-  /** Prevents input from being submitted without value */
+  /** Includes required asterisk label if true */
   isRequired: PropTypes.bool,
```
ayeshakmaz: Might be good to add the same function check here as is used for `isOptional` so that people can't mass in isRequired without passing in the `label` prop

zashraf1985: Done

src/components/Input/Input.story.js
zashraf1985: Done

src/components/Input/Input.story.js
```diff
@@ -47,6 +47,15 @@ stories
             type="text"
           />
         </fieldset>
+        <fieldset>
+          <Input
```
ayeshakmaz: Awesome 😄

PR #1137, opened 2019-03-26
Drau/tokens colors

daverau-optimizely:
# Design Tokens
1. This will help us greatly standardize colors, font sizes, spacing units, etc across Sass, JavaScript and beyond
2. Our first pass includes only color right now
3. Thumbtack publishes multiple packages from within the same repo (https://github.com/thumbtack/thumbprint/tree/master/packages) — I like this approach but would like to start using tokens now. If the reviewer things this is worth doing before we release, I'm open to the idea but would likely prefer to do that in another PR.
4. This first pass imports design tokens for colors into our SASS package. Once this lands in OUI I'll be making another PR to update Optimizely.git and OUI with a search/replace across the codebases.

```
import designTokens from "@optimizely-oui/src/tokens/forimport/index.es";
```

Screen Shot 2019-03-27 at 5 17 38 PM


src/tokens/changelog.md
```diff
@@ -0,0 +1,8 @@
+# OUI Design Tokens Change Log
```
circAssimilate: Have we started including changelog on the component / module level across all of OUI? Was only thinking we did in the main changelog. Mostly just curious!

.storybook/tokens.story.js
```diff
@@ -0,0 +1,65 @@
...
+import { CopyToClipboard } from "react-copy-to-clipboard";
+
+import Select from "../src/components/Select";
+import TabNav from "../src/components/TabNav";
+import Code from "../src/components/Code";
+import Icon from "react-oui-icons";
+
+import css from "/Users/drau/Sites/csswhat/css/oui.css.json";
+import designTokens from "../src/tokens/src/color.json";
+// import OUItokens from "@optimizely-oui/src/tokens/forimport/index.es";
```
circAssimilate: Should this be removed?

.storybook/config.js
```diff
@@ -1,28 +1,28 @@
...
+import { setOptions } from "@storybook/addon-options";
 
 require(`../dist/styles.js`);
 
 require(`../src/oui/oui.scss`);
 
 setOptions({
-  name: 'OUI Storybook',
-  url: 'https://github.com/optimizely/oui',
+  name: "OUI Storybook",
```
circAssimilate: nit: Was this an autofix? Was thinking our linter prefers single quotes actually

.storybook/tokens.story.js
```diff
@@ -0,0 +1,65 @@
...
+          products and services.
+        </p>
+
+        <Code
+          hasCopyButton
+          testSection="my-code-box"
+          type="block"
+          language="js"
+        >
+          import designTokens from
```
circAssimilate: Does this inline import not throw an error?!

.storybook/config.js
daverau-optimizely: Just got back from vacation — thanks for all your comments here @circAssimilate.

Hah, yeah I'm experimenting with using VSCode full-time and haven't gotten all my settings locked down yet. Fixing this now.

src/tokens/changelog.md
daverau-optimizely: No only at the global level right now but I thought maybe we'd want to keep better track of these changes since they represent design decisions given a context, vs just a dumping ground of variables. The former might benefit from a more thoughtful changelog doc. Thoughts?

.storybook/tokens.story.js
daverau-optimizely: Good catch although JSX seems to be pretty smart here and know this is text for the node and not an import statement. Works okay on my end right now and since it's a story file no real harm once this lands in Optimizely.git.

.storybook/config.js
circAssimilate: If this was a mistake, can we revert this back to single quotes so it complies with our linter?

.storybook/tokens.story.js
circAssimilate: Ohhh, I totally didn't connect this was a code sample. Carry on!

PR #1143, opened 2019-04-11
Drau/button story

daverau-optimizely:
Quick storybook fixes:
1. Show button props table on first example
2. Simplify Tokens story to a single page
3. Fix tokens copy CamelCaseId

### DIFF
--- a/src/components/Button/Button.story.js
+++ b/src/components/Button/Button.story.js
@@ -20,7 +20,7 @@ stories
   ));
 
 stories
-  .add('customize me!', () => {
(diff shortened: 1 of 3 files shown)


src/components/Button/Button.story.js
```diff
@@ -20,7 +20,7 @@ stories
   ));
 
 stories
-  .add('customize me!', () => {
+  .add('customize me!', withInfo()(() => {
```
ayeshakmaz: Good catch

PR #1144, opened 2019-04-12
Fix display of order when token has order but isn't draggable.

raykhenberg:
This is a simple fix for a problem that I just ran into.

If you create an ordered token without a drag handle (probably a rare case) it doesn't render the number properly.

Screen Shot 2019-04-11 at 5 46 54 PM

Simply adding the class that we add when we have the drag handle solves this problem.

Screen Shot 2019-04-11 at 5 47 59 PM


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ This file is similar to the format suggested by [Keep a CHANGELOG](https://githu
 
 ## Unreleased
 - [Patch] Minor fixes to Storybook Button and Design Tokens stories ([#1143](https://github.com/optimizely/oui/pull/1143))
+- [Patch] Fix token order rendering.
```
ayeshakmaz: Maybe add to this a bit, like 'Fix token order rendering when token is not draggable

src/components/Token/index.js
```diff
@@ -38,7 +38,7 @@ const Token = ({
     'flex': true,
   });
   const tokenToolsClasses = classNames({
-    'oui-token-tool': isDraggable,
+    'oui-token-tool': isDraggable || order,
```
ayeshakmaz: One downside here - the `oui-token-tool` class has `cursor: move`, so having this class without the draggable icon still makes it seem like the token could be re-ordered if they mouse over the number. 

Let me chat with the Design team and see how they want to fix this, and I'll get back to you

ayeshakmaz: Ok one small change - could you:

1. Remove `cursor: move` from the `oui-token-tool` class and
2. Create a new class `oui-token--drag` class that just has `cursor:move` and this class would be added anytime the token is draggable

This way the tokens don't seems draggable when they're not

raykhenberg: Nice catch. Just made these changes.

PR #1146, opened 2019-04-15
feat (OverlayWrapper) Adding attachmentConstraint prop

pauloptimizely:
## Summary

Adding an `attachmentConstraint` prop to the `OverlayWrapper`. This prop will allow a consumer to set how the attachment is constrained to the target.

For example, tethr will reposition the popover on scroll

popover_issue

This fix for this issue , would be to set the `attachmentConstraint` to `bottom`

## Test Plan

Run the test suite `npm test`

I also tested this within [optly-components](https://github.com/optimizely/optly-components) using `npm link`

## Jira Issues

https://optimizely.atlassian.net/browse/MGMT-1917

### DIFF
--- a/src/components/OverlayWrapper/index.js
+++ b/src/components/OverlayWrapper/index.js
@@ -32,7 +32,7 @@ class OverlayWrapper extends React.Component {
       attachment: `${this.props.verticalAttachment} ${this.props.horizontalAttachment}`,
       constraints: [{
         to: 'window',
-        attachment: 'together',
+        attachment: this.props.attachmentConstraint,
         pin: this.props.isConstrainedToScreen,
       }],
     };
@@ -249,6 +249,8 @@ class OverlayWrapper extends React.Component {
 }
 
 OverlayWrapper.propTypes = {
+  /** Determines how the attachment is constrainted to the target */
+  attachmentConstraint: PropTypes.string,
   /** Event to listen to and open the overlay */
   behavior: PropTypes.oneOf(['click', 'hover']),
   /** Element that the `overlay` should attach to */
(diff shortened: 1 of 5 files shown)


src/components/OverlayWrapper/index.js
```diff
@@ -249,6 +249,8 @@ class OverlayWrapper extends React.Component {
 }
 
 OverlayWrapper.propTypes = {
+  /** Determines how the attachment is constrainted to the target */
+  attachmentConstraint: PropTypes.string,
```
ayeshakmaz: Are there a finite number of options for this prop? Maybe consider doing oneOf[] to make it a little more constrained (no pun intended :P )

pauloptimizely: good call!! I'll make that update

ayeshakmaz: Also just noticed a typo in your comment: constrainted instead of constrained

src/components/OverlayWrapper/index.js
```diff
@@ -249,6 +249,8 @@ class OverlayWrapper extends React.Component {
 }
 
 OverlayWrapper.propTypes = {
+  /** Determines how the attachment is constrainted to the target */
+  attachmentConstraint: PropTypes.oneOf(['both', 'element', 'none', 'target', 'together']),
```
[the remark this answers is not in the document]
pauloptimizely: I'll update the tests to use 'none' instead :+1:

pauloptimizely:
@ayeshakmaz sweet!! 🎉 Thanks for the speedy review!

PR #1149, opened 2019-04-26
[FE-950] Upgrade react-popper and fix breaking changes

zachpwr:
# Summary

This PR upgrades the version of `react-popper` from 0.8.2 to 1.3.3. This change is meant to fix an intermittent JS error that was thrown by the old version of `react-popper`.

image

This version bump contains breaking changes. First, the `<Target />` component needed to be changed to `<Reference />`. Second, both `<Popper />` and `<Reference />` components needed to be updated to use render props to pass `ref`s to their children. Finally, certain props passed to `react-popper` components (such as `style`, `data-*`, `className`) are now unsupported and needed to be moved to wrapper `<div />`s.

# Test Plan

Since certain components were renamed in `react-popper` and the component hierarchy of `<Dropdown />` was altered slightly, the recorded portion of the dropdown unit test was modified to reflect these changes.


jamesopti:
YOU ARE A SAINT

LMK when ready for full review

zachpwr:
@jamesopti Thanks! Feel free to leave some feedback if you'd like to review 👍

PR #1151, opened 2019-05-01
feat(sheet/dialog):  Updates to allow for node in subtitle

ayeshakmaz:
Updates to let subtitle be a node or a string in Dialog and Sheet


src/components/Dialog/index.js
```diff
@@ -7,27 +7,35 @@ import Footer from './Deprecated/Footer';
...
-        {props.subtitle && <p className="push--top flush--bottom">{props.subtitle}</p>}
-      </header>
-      <div className="oui-dialog__body overflow-y--auto">
-        {props.children}
+export const DialogNew = props => {
+  let subtitleContent;
+  if (props.subtitle && typeof (props.subtitle) === 'string') {
+    subtitleContent = <p className="push--top flush--bottom">{props.subtitle}</p>;
+  } else {
+    subtitleContent = <div className="push--top flush--bottom">{props.subtitle}</div>;
```
raykhenberg: Just curious, why not use the div for both? Is there some kind of tacit agreement that a div can't have plain text in it without a <p> tag around it?

ayeshakmaz: It's generally better semantically/for accessibility reasons to have any plain text inside a <p> instead of a <div>. Not a huge deal, just slightly more specific I guess

PR #1159, opened 2019-05-07
feat(Textarea) Adding MinHeight prop

Hassan-Khalid-folio3:
# Summary
Added an optional prop to set minimum height of textarea.

### The Following File includes changes
1. `src/components/Textarea/index.js`

### The Following Storybook file includes changes
1. `src/components/Textarea/Textarea.story.js `

# Updated CHANGELOG

### DIFF
--- a/src/components/Textarea/index.js
+++ b/src/components/Textarea/index.js
@@ -41,7 +41,7 @@ class Textarea extends React.Component {
     return (
       /* eslint-disable react/jsx-no-bind */
       <textarea
-        style={ { minHeight: minHeight } }
+        style={{ minHeight: minHeight }}
         data-oui-component={ true }
         className={ classes }
         ref={ (c) => { this._textarea = c; } }
--- a/src/components/Textarea/index.js
+++ b/src/components/Textarea/index.js
@@ -23,6 +23,7 @@ class Textarea extends React.Component {
     isRequired,
     isDisabled,
     maxLength,
+    minHeight,
     onBlur,
     onChange,
     onFocus,
@@ -40,6 +41,7 @@ class Textarea extends React.Component {
     return (
       /* eslint-disable react/jsx-no-bind */
       <textarea
+        style={ { minHeight: minHeight } }
         data-oui-component={ true }
         className={ classes }
         ref={ (c) => { this._textarea = c; } }
@@ -133,6 +135,10 @@ Textarea.propTypes = {
    * Max length of the input value
    */
   maxLength: PropTypes.number,
+  /**
+   * Min height of the text area
+   */
+  minHeight: PropTypes.number,
   /** Form note for the input */
   note: PropTypes.string,
   /**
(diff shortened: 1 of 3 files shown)


src/components/Textarea/index.js
```diff
@@ -165,6 +171,7 @@ Textarea.defaultProps = {
   label: null,
   note: null,
   isOptional: false,
+  minHeight: null,
```
spham1412: Do you think the default value for `minHeight` should not be `null` or `0`? Because that would collapse the Textarea if users forgot to pass it in, should it has at least `20-30px` to accommodate at least 2 lines of font-size 12?

spham1412: A text area should not collapse all the way down because it would make it not clickable to type. Let me know what you think or there's another hidden solution that already solved this.

ayeshakmaz:
@fayyazarshad Let's hold off on this for now, and see if CodeMirror is a better case for having large sections of code (also mentioned this in our thread in slack) Dave and I are hesitant to let people style the height of the textarea to any possible value cc @daverau-optimizely

src/components/Textarea/index.js
daverau-optimizely: Textarea comes in two fixed heights so we have consistent text input throughout the application. Is there a reason we'd not want to use CodeMirror component here @Hassan-Khalid-folio3? That's my preference unless you have some rationale for using textarea instead, I'd love to hear your thoughts here. Thanks! 🙌

Hassan-Khalid-folio3:
Hey @daverau-optimizely . In our case, we can use Code Mirror component but if you look at the code base you will find many examples where `<textarea>` is used with variable heights(min-height). (a few of them are mentioned below). 
1. `src/js/bundles/p13n/sections/plugin_builder/pages/plugin_editor/components/sidebar/edit_schema_field/template.html`
2. `src/js/bundles/p13n/components/editor/sidebar/change_editor_sidebar/base/template.html`
3. `src/js/bundles/p13n/sections/recommender_settings/components/edit_recommender_dialog/index.js`

First two are vue components which use standard html `<textarea>` with a class `min-height--200` and with the current OUI `<Textareat>` we can't achieve that. 

The last one is a react component which uses the standard html `<textarea>` with a class `min-height--200`.

daverau-optimizely:
@Hassan-Khalid-folio3 Let's try a prop that toggles `height--1-1` so you either use the pre-set fixed heights or you use height 100% to fill the available space. 

Can you provide a screenshot example of the issue you're hoping to solve too, just so we have the specific use case shown here. Thanks! 🙌

Hassan-Khalid-folio3:
Hey @daverau-optimizely. I noticed that in the above mentioned files, Code mirror can be used instead of `textarea` because they are used to input `json` and `html` in some cases. So I think the prop for minHeight isn't required anymore. We can use Code mirror instead.

daverau-optimizely:
Closing this in favor of using Code Mirror component.

PR #1165, opened 2019-05-31
patch(DatePicker): Properly export DatePicker and DateRangePicker

circAssimilate:
I made an assumption of how DateRangePicker was defined and mistakenly merged a change that introduced a build error. This will fix that and export both DatePicker and DateRangePicker.


src/components/DatePicker/DatePicker.js
```diff
@@ -148,3 +146,5 @@ DatePicker.defaultProps = {
   keepOpenAlways: false,
   keepOpenOnDateSelect: false,
 };
+
+export default DatePicker;
```
circAssimilate: To me, it seemed like this and `DateRangePicker` default export should be at the bottom.

PR #1169, opened 2019-06-05
feat(Dropdown): Allow render props with hide function

circAssimilate:
**Summary:**
The goal of this PR is to fix a bug where the toggle function is always called when the Dropdown children elements are clicked, and also to also optionally allow a hide function to be used via the children using a [render props](https://reactjs.org/docs/render-props.html) function. The component will provide the children with the `handleHideChildren` function if the child has the type of function, otherwise it will render the them directly as a node. Unless we expect react to make nodes functions, this will be both backwards compatible and future safe.

Pull this branch and run `yarn storybook` to play with the "Custom hide function" story.

image

**Test plan:**
- [x] Unit tests
- [x] Manual testing


src/components/Dropdown/index.js
```diff
@@ -37,7 +37,7 @@ class Dropdown extends React.Component {
   render() {
     const {
       activator,
-      arrowIcon,
+      arrowIcon = 'none',
```
circAssimilate: Moving this to hear and away from defaultProps since this does the same.

src/components/Dropdown/index.js
```diff
@@ -133,7 +134,13 @@ class Dropdown extends React.Component {
                     boxShadow: '0 2px 3px rgba(0,0,0,.1)',
                     ...popperStyle,
                   }}>
-                  {children}
+                  {
+                    withRenderProps ? (
+                      children({ hideFn: this.props.hide })
```
circAssimilate: Check out examples of this within OUI in the FilterPicker component and in the React docs: https://reactjs.org/docs/render-props.html

src/components/Dropdown/index.js
```diff
@@ -205,6 +212,8 @@ Dropdown.propTypes = {
   testSection: PropTypes.string,
   /** Toggle control for isOpen. */
   toggle: PropTypes.func,
+  /* Control for overChildren value */
+  withRenderProps: PropTypes.bool,
```
circAssimilate: To me, this seemed like a good, lightweight option to allow for more control if the consumer needs while not forcing everyone to use renderProps.

src/components/Dropdown/index.js
```diff
@@ -205,6 +212,8 @@ Dropdown.propTypes = {
   testSection: PropTypes.string,
   /** Toggle control for isOpen. */
   toggle: PropTypes.func,
+  /* Control for overChildren value */
```
circAssimilate: Todo: update docstring

src/components/Dropdown/index.js
```diff
@@ -40,15 +54,16 @@ class Dropdown extends React.Component {
       arrowIcon,
       buttonContent,
       children,
-      isDisabled = false,
+      isDisabled,
```
circAssimilate: This felt awkward to only define defaults for props in one method - so I moved all these to defaultProps and also cleaned up some other props things. Wouldn't hurt to have a second pair of eyes check that the updates are safe.

src/components/Dropdown/index.js
```diff
@@ -133,7 +151,13 @@ class Dropdown extends React.Component {
                     boxShadow: '0 2px 3px rgba(0,0,0,.1)',
                     ...popperStyle,
                   }}>
-                  {children}
+                  {
+                    typeof children === 'function' ? (
```
circAssimilate: As mentioned in the description, the component will provide the children with the handleHideChildren function if the child has the type of function, otherwise it will render the them directly as a node. Unless we expect react to make nodes functions, this will be both backwards compatible and future safe.

src/components/Dropdown/Dropdown.story.js
```diff
@@ -126,7 +128,8 @@ stories.add('Z-index', (() => {
           <Dropdown
             isDisabled={ boolean('isDisabled', false) }
             buttonContent={ text('buttonContent', 'Dropdown') }
-            width={ number('width', 300) }>
+            width={ number('width', 300) }
+            zIndex={ number('zIndex', 9999) }>
```
circAssimilate: Odd that this knob or prop wasn't added!

src/components/Dropdown/Dropdown.story.js
```diff
@@ -213,6 +216,58 @@ stories.add('Warnings', (() => {
...
+          left: 'left',
+          'left-start': 'left-start',
+          'left-end': 'left-end',
+        }, 'bottom-start') }
+        shouldHideChildrenOnClick={ boolean('shouldHideChildrenOnClick', false) }
+        style="outline"
+        width={ number('width', 300) }>
+        {
+          ({ handleHideChildren }) => (
+            <div className="soft-double background--white">
```
daverau-optimizely: Personally I'd like to see just `soft` here, no double.

src/components/Dropdown/Dropdown.story.js
```diff
@@ -213,6 +216,58 @@ stories.add('Warnings', (() => {
...
+stories.add('Custom hide function', (() => {
+  return [
+    <p key="sourceLink" className="push--bottom">
+      This story's children use the optional <Link href="https://reactjs.org/docs/render-props.html" newWindow={ true }>render props pattern</Link>. <Link href="https://github.com/optimizely/oui/blob/devel/src/components/Dropdown/Dropdown.story.js" newWindow={ true }>Click here</Link> to see this story's source code.
+    </p>,
+    <Container key="story">
+      <Dropdown
+        arrowIcon={ select('arrowIcon', { up: 'up', down: 'down', left: 'left', right: 'right', none: 'none' }, 'down') }
+        buttonContent={ (
+          <div className="line--tight text--left micro push--right">
```
daverau-optimizely: `line--tight` is like `0.2` too loose in my opinion, 1.2 looks tiiiiiiight — but we don't have this in OUI, so I'll fix that. Results page uses `1.0` which is too tight, so I'll fix these two things and update this story when it's ready. 

No change for you here Derek, just a note about what I'll be changing! 😉

circAssimilate: Ok, thanks! Will leave for now but feel free to update whenever - you got better 👀 for this stuff!

PR #1173, opened 2019-06-17
fix(progress-dots): Prevent wrapping of dots

ayeshakmaz:
Add `nowrap` to the progress dots

See https://files.slack.com/files-pri/T02V23TD2-FKKECNC8P/progressdots.gif for example

JIRA [UXENG-7](https://optimizely.atlassian.net/browse/UXENG-7)


src/components/ProgressDots/index.scss
```diff
@@ -12,6 +12,7 @@
 .#{$namespace}progress-dots,
 .oui-progress-dots {
   display: inline-block;
+  white-space: nowrap;
```
daverau-optimizely: Curious about your thoughts of using the `.nowrap` helper class on the component output vs adding to the overall CSS. I guess my preference would be to push this to the output but perhaps that's harder to reason about when reading a component's CSS file... would be curious to talk a little more about this trade-off as we figure out more react/css-in-js/future-of-sass workflow and best practices for OUI.

PR #1176, opened 2019-06-18
feat(selectDropdown): Add ability to provide a placeholder

ayeshakmaz:
Add `initialPlaceholder` prop for the SelectDropdown cmp

Screen Shot 2019-06-18 at 1 33 09 PM

JIRA [UXENG-12](https://optimizely.atlassian.net/browse/UXENG-12)


daverau-optimizely:
Great change — how have we gone this long without this prop?! 💯

src/components/SelectDropdown/index.js
```diff
@@ -63,7 +67,7 @@ class SelectDropdown extends React.Component {
       PropTypes.string,
       PropTypes.number,
       PropTypes.bool,
-    ]).isRequired,
```
ryn-optimizely: Since `value` is no longer required, there needs to be a default prop for value.

PR #1180, opened 2019-06-21
Initial capitalize all Storybook names

daverau-optimizely:
Changes all story names from `buttons` to `Buttons`


src/components/Code/Code.story.js
```diff
@@ -2,6 +2,7 @@ import React from 'react';
 
 import { storiesOf } from '@storybook/react';
 import { withKnobs, text, boolean, select } from '@storybook/addon-knobs';
+import { withInfo } from '@storybook/addon-info';
```
ayeshakmaz: Curious why this is needed

src/components/Code/Code.story.js
```diff
@@ -44,7 +45,7 @@ stories
       </Code>
     );
   }))
-  .add('with styled copy button', (() => {
+  .add('With styled copy button', withInfo()(() => {
```
ayeshakmaz: We shouldn't need this, since withInfo is added at the global level

src/components/Sheet/Sheet.story.js
```diff
@@ -4,6 +4,7 @@ import { storiesOf } from '@storybook/react';
 import { withKnobs, text, boolean } from '@storybook/addon-knobs';
 import noop from 'lodash.noop';
 import { action } from '@storybook/addon-actions';
+import { withInfo } from '@storybook/addon-info';
```
ayeshakmaz: Same here

src/components/Sheet/Sheet.story.js
```diff
@@ -16,53 +17,51 @@ stories
...
+        </Fieldset>
+      </Sheet>
+    </div>
+  ))
   )
-  .add(
-    'with a link in subtitle',
-    (() => (
+  .add('With a link in subtitle',
+    withInfo()(() => (
```
ayeshakmaz: and here, don't think this is needed

src/components/Code/Code.story.js
daverau-optimizely: Interesting, when I comment this out all of Storybook breaks at this point but throws no console error — no idea what gives here...

Screen Shot 2019-06-26 at 1 30 32 PM

PR #1181, opened 2019-06-27
Remove blue from aqua token

daverau-optimizely:
This PR addresses two issues:
1. Remove the word `blue` from the aqua tokens so the names are consistent with other color tokens
2. Update `yarn storybook` so it updates tokens file on a watch process

Note: this is a breaking OUI change but we are not using this value in the product currently.


raykhenberg:
I was consuming this by importing the file and referencing `tokens.aquaBlueBase`. My stuff is still in progress and won't impact me even if it is broken, but is it possible that someone is consuming this in the wild? Will this break anything?

daverau-optimizely:
> I was consuming this by importing the file and referencing `tokens.aquaBlueBase`. My stuff is still in progress and won't impact me even if it is broken, but is it possible that someone is consuming this in the wild? Will this break anything?

Yeah it's possible — and I'll be doing a search/replace across the app when I bump this version of OUI in the app.

daverau-optimizely:
> > I was consuming this by importing the file and referencing `tokens.aquaBlueBase`. My stuff is still in progress and won't impact me even if it is broken, but is it possible that someone is consuming this in the wild? Will this break anything?
> 
> Yeah it's possible — and I'll be doing a search/replace across the app when I bump this version of OUI in the app.

Follow-up here: I searched the app and we aren't using this variable outside of OUI yet — besides the changes you have in progress now @raykhenberg.

raykhenberg:
Perf. Ship it!

PR #1186, opened 2019-07-03
feat(DateRangePicker)[APPX-636]: Introduce render props option for onDatesChange access

circAssimilate:
**Summary:**
Currently, `DateRangePicker` only allows for the component to have its start and end date input initially set when the component is instantiated (via the constructor). After instantiation, the component controls its own state which poses a problem with custom provided `panelButtons`. Similar to how we did with the Dropdown component (https://github.com/optimizely/oui/blob/e4cf30b420899426d752940c5d2d575a8de7a6bc/src/components/Dropdown/index.js#L157), the `panelButtons` can optionally be a render props function (https://reactjs.org/docs/render-props.html) that gets access to the component's `onDatesChange` method. This will allow implementers like Change History to set the `DateRangePicker` component's state from the parent. This is a **backwards compatible** change and will allow any current users to continue passing `panelButtons` as an array of Buttons. However, if someone needs this extra power, they can simply pass a function that expects the `onDatesChange` function and returns an array of buttons.

**Screenshots:**
image

**Test Plan:**
- [x] Added stories to Storybook
- [x] Added 2 tests
- [x] Updated jest snapshots

**JIRA Issues:** [APPX-636](https://optimizely.atlassian.net/browse/APPX-636)


circAssimilate:
@ayeshakmaz, would love your feedback on this when you get a chance!

src/components/DatePicker/DateRangePicker.js
```diff
@@ -245,14 +256,21 @@ DateRangePicker.propTypes = {
...
    * are also defined, the buttons will be to the right of the calendar.
    * Otherwise, they are directly below the calendar.
+   * Use render prop function that returns an array to get access to component's
+   * onDatesChange method
    */
-  panelButtons: PropTypes.array,
+  panelButtons: PropTypes.oneOfType([
+    PropTypes.array,
+    PropTypes.func,
+  ]).isRequired,
```
ayeshakmaz: I don't think this should be a required prop (maybe copy pasta? :P )

src/components/DatePicker/DateRangePicker.js
```diff
@@ -245,14 +256,21 @@ DateRangePicker.propTypes = {
    * or end date has been selected.
    * The dates are passed as a parameter
    * object with startDate and endDate as keys
+   * Use render prop function that returns and array for panelButtons access
```
ayeshakmaz: Could you re-word this? I think there's some words missing 😛

src/components/DatePicker/DateRangePicker.story.js
```diff
@@ -173,9 +177,26 @@ stories
         startDateInputId='start-date-id-0'
       />
     </Container>
-  ))
-  )
-  .add('With no border on calendar', (() => (
+  )))
+  .add('With panelButtons an array (not render props)', (() => (
```
ayeshakmaz: "as an array" instead of "an array" maybe?

src/components/DatePicker/tests/index.js
```diff
@@ -4,6 +4,9 @@ import { mount } from 'enzyme';
 import DatePicker from '../DatePicker';
 import DateRangePicker from '../DateRangePicker';
 import Button from '../../Button';
+import Dropdown from "../../Dropdown/tests";
```
ayeshakmaz: These all seem like odd imports here... There also shouldn't be any snapshots folder within DatePicker, can you double check why that got added for <Dropdown> tests?

src/components/DatePicker/tests/index.js
```diff
@@ -289,4 +292,108 @@ describe('components/DateRangePicker', function() {
       .containsMatchingElement(<button>Clear</button>)).toBe(true);
   });
 
+  it('should pass onDatesChange via panelButtons when array is used (without render props pattern)', () => {
```
ayeshakmaz: Awesome tests!

src/components/DatePicker/tests/__snapshots__/index.js.snap
```diff
@@ -0,0 +1,272 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
ayeshakmaz: Don't think this file should be here

src/components/DatePicker/DateRangePicker.js
circAssimilate: Copy Pasta supreme! Good catch!

src/components/DatePicker/DateRangePicker.js
circAssimilate: Sure thing. I updated to be 

>* For access in panelButtons row items, use render prop function that returns an array.

src/components/DatePicker/tests/__snapshots__/index.js.snap
circAssimilate: Weird it's for dropdown. I'll delete and see if anything fusses.

circAssimilate:
Great catch on all my hasty mistakes, @ayeshakmaz! Thank you! Should be good for review again.

src/components/DatePicker/tests/index.js
circAssimilate: Thank you!

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This file is similar to the format suggested by [Keep a CHANGELOG](https://github.com/olivierlacan/keep-a-changelog).
 
 ## Unreleased
+- [Feature] Introduce render props option for onDatesChange access ((#1186)[https://github.com/optimizely/oui/pull/1186])
```
ayeshakmaz: Last tweak: 

> [Feature] Introduce render props option for `onDatesChange` access in **DateRangePicker** ([#1186](https://github.com/optimizely/oui/pull/1186))

(url is reverse syntax, i've fixed the one below in another PR)
(Bolded Cmp name, code-blocked prop name)

PR #1188, opened 2019-07-05
Drau/underline button hover

daverau-optimizely:
This PR does two things:
1. adds color to underline button hover states
2. moves all color values to OUI variables that are easier to reference and tokenize later


src/components/Button/index.scss
```diff
@@ -249,6 +249,12 @@
       border-bottom-color: map-fetch($color, ui, medium);
       color: map-fetch($color, background, brand-dark);
     }
+
+    &:hover, &:focus, &:active {
+      & .oui-arrow-inline--down {
+        color: map-fetch($button, type, outline, hover, border);
```
ayeshakmaz: Should we use a token instead here?

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This file is similar to the format suggested by [Keep a CHANGELOG](https://github.com/olivierlacan/keep-a-changelog).
 
 ## Unreleased
+- [Patch] Add hover style for button underline arrow ((#1188)[https://github.com/optimizely/oui/pull/1188])
```
ayeshakmaz: Can you switch the link syntax? ([#1188](link here))

src/components/Button/index.scss
daverau-optimizely: Great call — I'm in the process of wrapping this up now! 😉

PR #1190, opened 2019-07-09
feat(PaginationControls): New component for updated pagination design

ayeshakmaz:
**Summary:**
Add a new component for Pagination, allows for customization through maxSlots prop:

**Visuals:**
Default 9 slots:
Screen Shot 2019-07-09 at 4 49 18 PM

Changed to 11 slots:
Screen Shot 2019-07-09 at 4 49 34 PM

Loading State:
Screen Shot 2019-07-09 at 4 49 26 PM

Also updated `Link` to have ariaLabel prop

**Open questions:** 
1. Should the hrefs be used, rather than void?
2. Should this replace the current `Pagination` cmp?


src/components/PaginationControls/index.js
```diff
@@ -0,0 +1,185 @@
...
+    let sanitizedTotalPages = Math.max(totalPages, 2);
+    let adjustedPageNumber = currentPage;
+
+    if (adjustedPageNumber > sanitizedTotalPages) {
+      adjustedPageNumber = sanitizedTotalPages;
+    } else if (adjustedPageNumber < 1) {
+      adjustedPageNumber = 1;
+    }
+
+    for (let i = 0; i < controlSlots.length; i++) {
```
jamesopti: Prefer a more functional approach here:

```js

return controlSlots.map((val, idx) => {
  // do some logic

  return <li className="push--sides" key={ `link-${idx}` }> ... </li>;
})
```

src/components/PaginationControls/index.js
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  handlePageChange = (newPage) => {
+    return () => {
+      const { goToPage } = this.props;
+      goToPage(newPage);
+    };
+  }
+
+  fillPageSlots() {
```
jamesopti: This method is really hard to understand and I think could be quite simplified.

Wanna pair on doing that?

src/components/PaginationControls/index.js
```diff
@@ -0,0 +1,185 @@
...
+      updatedControlSlots.push(sanitizedTotalPages);
+    } else if (currentPage - buffer >= sanitizedTotalPages - maxSpread) {
+      updatedControlSlots.push(1);
+      updatedControlSlots.push(-1);
+      for (let i = sanitizedTotalPages - maxSpread; i <= sanitizedTotalPages; i++) {
+        updatedControlSlots.push(i);
+      }
+    }
+
+    this.setState({
```
jamesopti: Do we really need state for this?

I could see this being a functional component that computes this each time it renders. Not that expensive and its not a component who's props we expect to change a lot (only on page change).

If it was expensive, we could also memoize the result instead.

src/components/Link/index.js
```diff
@@ -52,6 +54,8 @@ const Link = ({
 };
 
 Link.propTypes = {
+  /** Label to use for screen readers */
+  ariaLabel: PropTypes.string,
```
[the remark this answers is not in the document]
ayeshakmaz: Nah, if it has a default value, that attribute will always show up, which doesn't break anything but is probably not the best. We only want that attribute when we need to enhance the actual Link content

src/components/PaginationControls/PaginationControls.story.js
```diff
@@ -0,0 +1,54 @@
...
+
+const totalPages = 43;
+
+stories
+  .add('Default', (() => {
+    return (
+      <PaginationControls
+        currentPage={ number('currentPage', 1) }
+        totalPages={ number('totalPages', totalPages) }
+        goToPage={ action('page changed') }
```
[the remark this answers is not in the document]
ayeshakmaz: It does in Storybook, just doesn't look like it here

ayeshakmaz:
@jamesopti I think we addressed everything while pairing, let me know watchya think

src/components/PaginationControls/index.js
```diff
@@ -0,0 +1,229 @@
...
+    return {
+      totalPages: sanitizedTotalPages,
+      totalSlots: sanitizedTotalSlots,
+      buffer,
+      remainingSlotsToFill,
+      currentPage: sanitizedCurrentPage,
+    };
+  }
+
+  fillPageSlotsOption2() {
```
jamesopti: I prefer this option as I find it more logical and more flexible to update in the future.

It's also much more functional which I prefer.

src/components/PaginationControls/index.js
```diff
@@ -0,0 +1,189 @@
+/* eslint-disable no-script-url */
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { checkPropIsGreaterThanOrEqualTo } from '../../utils/custom-prop-types';
+import Spinner from '../Spinner';
+import Link from '../Link';
+import _ from 'lodash';
```
quarklemotion: I don't see **full** `lodash` listed in the dependencies in package.json - there are only 2 scoped packages `lodash.debounce` and `lodash.noop`.  We should update package.json rather than rely on 3rd party libraries to pull in lodash.  I think we should `yarn add lodash@4.x.x` and then use `import flatten from lodash/flatten;` here.

ayeshakmaz: @quarklemotion Thanks for catching this, I added full lodash to this PR and updated this line, but haven't gone back to adjust any of the other partial uses. Figure I can do that in a separate PR if need be, but let me know if you'd prefer it in this one

PR #1191, opened 2019-07-10
CodeDiff component

daverau-optimizely:
WIP component with @ayeshakmaz 

```
      <CodeDiff
        oldFilename={ 'Old file' }
        newFilename={ 'New renamed file' }
        oldText={ 'Test' }
        newText={ 'A test with new things added and changed.' }
      />
```

Note: there seems to be a bug with how react-gh-like-diff handles the `matching` prop because in our trials it didn't seem to work the way we expected: https://codesandbox.io/s/react-github-like-diff-example-u4j4h

```matching: matching level: 'lines' for matching lines, 'words' for matching lines and words or 'none', default is none```

cc @circAssimilate @benjaminpadula

### CODE REVIEW

File: src/components/CodeDiff/index.js [hunk 1]
```diff
@@ -0,0 +1,51 @@
...
+  updatedFileName,
+  outputStyle,
+}) => {
+
+  return (
+    <div className="oui-codediff">
+      <ReactGhLikeDiff
+        options={{
+          originalFileName: fileName,
+          updatedFileName: updatedFileName || fileName,
```
circAssimilate: Nice. I like this. default of filteName. Do you think it's also worth having the option of keeping this blank too? I was playing around in CodeSandbox and couldn't really think of a great title and kinda thought the text could be removed altogether.

src/components/CodeDiff/index.js
```diff
@@ -0,0 +1,51 @@
...
+  /** Hook for automated JavaScript tests */
+  outputStyle: PropTypes.oneOf(['side-by-side', 'line-by-line']),
+  /** Hook for automated JavaScript tests */
+  testSection: PropTypes.string,
+  /** Hook for automated JavaScript tests */
+  updatedFileName: PropTypes.string,
+};
+
+CodeDiff.defaultProps = {
+  fileName: '',
```
circAssimilate: I like that this is blank by default. But it's required above.

src/components/CodeDiff/index.js
```diff
@@ -0,0 +1,51 @@
...
+  /** Hook for automated JavaScript tests */
+  newText: PropTypes.string.isRequired,
+  /** Hook for automated JavaScript tests */
+  oldText: PropTypes.string.isRequired,
+  /** Hook for automated JavaScript tests */
+  outputStyle: PropTypes.oneOf(['side-by-side', 'line-by-line']),
+  /** Hook for automated JavaScript tests */
+  testSection: PropTypes.string,
+  /** Hook for automated JavaScript tests */
+  updatedFileName: PropTypes.string,
```
circAssimilate: Let's give this and `testSection` default, empty string props too.

src/components/CodeDiff/index.js
```diff
@@ -0,0 +1,51 @@
...
+};
+
+CodeDiff.defaultProps = {
+  fileName: '',
+  newText: '',
+  oldText: '',
+  outputStyle: 'side-by-side',
+};
+
+export default CodeDiff;
```
circAssimilate: Seems like we just make all props optional and give them defaults.

src/components/CodeDiff/index.js
daverau-optimizely: Great idea, just added this.

src/components/CodeDiff/index.js
daverau-optimizely: @ayeshakmaz and I decided it would be odd to allow optional old/new text since without both the point of the component (at least in change history) is unnecessary. Can you think of a potential future use case where we'd want to be able to leave one blank?

PR #1195, opened 2019-07-18
Add optional fillColorName property to RangeSlider

quarklemotion:
- Add new `fillColorName` property to the `RangeSlider` component.  This optional property allows the customization of the color of the range slider.

Screen Shot 2019-07-17 at 7 20 29 PM


src/components/RangeSlider/index.scss
```diff
@@ -60,13 +60,18 @@
...
     position: absolute;
     left: 12px;
     right: 11px;
   }
+  .range-display svg {
+    display: inline-flex;
+    width: 100%;
+    height: 100%;
+  }
+  .range-display.default-fill-color svg path {
```
ayeshakmaz: Let's check with @daverau-optimizely when he's back to see if we can map this to `brandBlueDark` instead, so we can avoid the new class name

src/components/RangeSlider/tests/index.js
```diff
@@ -3,49 +3,75 @@ import RangeSlider from '../index';
...
+    it('should set a data-test-section attribute from the testSection prop', () => {
+      const testSection = component.find('[data-test-section="myTestSection"]');
+      expect(testSection.length).toBe(1);
+    });
   });
 
-  it('should set a data-test-section attribute from the testSection prop', () => {
-    const testSection = component.find('[data-test-section="myTestSection"]');
-    expect(testSection.length).toBe(1);
+  describe('custom fill color', () => {
```
ayeshakmaz: Thanks for adding tests!

daverau-optimizely: I do worry about allowing arbitrary color values into the mix, is there a good use case for this or just for hack stuff?

quarklemotion: @daverau-optimizely we met with Shane and Ron last week and developed this curated list of allowed rangeslider colors based on their feedback.  There is prop-type validation on the component to ensure only the named curated colors can be used and the way this color token mapping is implemented makes it impossible for arbitrary colors to be specified by the client app.

src/components/RangeSlider/constants.js
```diff
@@ -0,0 +1,28 @@
...
+  redBase,
+  magentaBase,
+  greyBase,
+} from '../../tokens/forimport/index.es';
+
+// map fillColorName prop values to OUI color tokens
+export const FILL_COLOR_MAP = {
+  aqua: aquaBase,
+  amber: amberBase,
+  default: brandBlueDark,
```
quarklemotion: FYI, added a new `default` color so that this can be used as the default `fillColorName` prop value.

ayeshakmaz: Looks great!

src/components/RangeSlider/index.js
```diff
@@ -23,6 +26,14 @@ const RangeSlider = (props) => {
             <label className="oui-label muted flush">100%</label>
           </div>
           <div className="range-display">
+            <svg
+              preserveAspectRatio="none"
+              xmlns="http://www.w3.org/2000/svg"
+              width="256"
+              height="256"
+              viewBox="0 0 256 256">
+              <path fill={ FILL_COLOR_MAP[fillColorName] } d="M0 256h256V0z"/>
```
ayeshakmaz: Would it be possible to safeguard this and switch to using `default` if `FILL_COLOR_MAP[fillColorName]` is not found/doesn't return anything? Right now if I put `pinkish` as the value for `fillColorName`, I get a black range slider. Might be nice to default back to `default` in this case. Let me know what you think

src/oui/_oui-variables.scss
```diff
@@ -368,7 +368,7 @@ $token: (
 
 $switch: (
   background-color: (
-    on                       : map-fetch($brand-color, light-blue-75),
+    on                       : $brand-blue-dark,
```
ayeshakmaz: 🎉

src/components/RangeSlider/constants.js
daverau-optimizely: After @ayeshakmaz showed me this yesterday, I keep thinking about it and really like the way this is handled! Would love to turn this into a more official pattern for OUI components that allow color values. 

Great system-focused work here MiFi! 💯

src/components/RangeSlider/index.js
quarklemotion: @ayeshakmaz I think this is a good idea - I had originally thought that having the enumerated allowed prop values was enough, but I think you are right that we should validate the color string and never allow the black color to be used as a default

PR #1199, opened 2019-07-22
feat(buttonIcon): Add ability to specify iconFill color

ayeshakmaz:
Summary:

Change the color of the icon contained in the button icon


src/components/ButtonIcon/tests/index.js
```diff
@@ -35,6 +35,12 @@ describe('ButtonIcon component', () => {
     );
     expect(component.find('.oui-button--danger').length).toBe(1);
   });
+  it('should fill the icon when iconFill is supplied', () => {
+    const component = mount(
+      <ButtonIcon iconName="add" iconFill="#abcabc"/>
```
daverau-optimizely: I'm hesitant to allow arbitrary hex/color values in here — thoughts on using a predefined set of token values instead?

ayeshakmaz:
@daverau-optimizely can correct me if I'm wrong, but I believe that is still the plan - to merge the two

PR #1202, opened 2019-07-25
Drau/codediff table tightening

daverau-optimizely:
This PR improves the CodeDiff component when nested inside of an OUI table:
- add white background
- add CSS to remove some OUI styles
- prevent OUI padding from cascading past one level of nesting

This uses `noHover=true` and you can see the 2nd table still has hover styles for the individual TRs, but not for the single TR the whole table sits within:
Screen Shot 2019-07-25 at 1 19 51 PM


data/csswhat/oui.css
```diff
@@ -6981,15 +6989,15 @@ body:target *[data-oui-component],
 
 
 .color--good-news {
-  color: #29a65e !important; }
+  color: #087d23 !important; }
```
circAssimilate: Chatted with Dave about this IRL and looks like this may have come through because of someone else's previous commit. Just worth double checking this is intended!

data/csswhat/oui.css
```diff
@@ -6981,15 +6989,15 @@ body:target *[data-oui-component],
...
 
 
 .color--warning {
-  color: #FFD233 !important; }
+  color: #c97b1d !important; }
```
circAssimilate: same as above

src/components/CodeDiff/CodeDiff.story.js
```diff
@@ -43,6 +44,30 @@ stories
       />
     );
   }))
+  .add('Tables nested', (() => {
```
circAssimilate: nit: Could we call this _CodeDiff nested in Table_?

src/components/CodeDiff/CodeDiff.story.js
```diff
@@ -43,6 +44,30 @@ stories
...
+    return (
+      <Table density="loose" tableLayoutAlgorithm="fixed">
+        <Table.THead>
+          <Table.TR>
```
circAssimilate: I think this format could be helpful so a non code diff table row is shown:

```
<Table>
  <Table.THead>
    <Table.TR>
      <Table.TH>
        Change Summary
      </Table.TH>
    </Table.TR>
  </Table.THead>
  <Table.TBody>
    <Table.TR>
      <Table.TD>
        Regular row
      </Table.TD>
     <Table.TR>
    <Table.TR>
      <Table.TD>
        Code Diff
      </Table.TD>
     <Table.TR>
   </Table.TBody>
</Table>
```

data/csswhat/oui.css
daverau-optimizely: These changes aren't changes in my PR but just a reflection of CSS What running against a previous PR. I fixed this issue in another separate PR here:

daverau-optimizely: 👍 Fixed in this PR: https://github.com/optimizely/oui/pull/1203

src/components/CodeDiff/CodeDiff.story.js
daverau-optimizely: 👍 Good call, I added another nested table to the example so we can ensure the nesting looks good.

data/csswhat/oui.css
```diff
@@ -2838,6 +2838,14 @@ li > ol {
   .oui-table td {
     vertical-align: top; }
   
+  .table > tbody > tr:hover:not(.hover--disabled) > td:not(.hover--disabled),
```
circAssimilate: I'm not seeing the `hover--disabled` being able to be added to `TD` through props. Should we remove it from all the `td`'s here?

src/components/CodeDiff/overrides.scss
```diff
@@ -1,3 +1,18 @@
...
+  padding: 0 !important;
+  margin: 0 !important;
+  vertical-align: unset !important;
+}
+
+.d2h-file-side-diff {
+  overflow-x: auto !important;
+}
+
+// .change-history-code-diff-row td {
```
circAssimilate: We may want to fully delete this if we're getting it another way!

data/csswhat/oui.css
daverau-optimizely: Yeah I was leaving this in more as a future thing but good catch, I'll remove so it's not confusing.

src/components/CodeDiff/overrides.scss
daverau-optimizely: 👍

PR #1206, opened 2019-07-29
patch(Table/CodeDiff): Revert Table Styles and Update CodeDiff Styles

circAssimilate:
**Summary:**
OUI version `v44.3.0` introduced a CSS regression for how table styles are applied in the monolith, in a large part due to CSS overrides happening there. This reverts the Component css module to the previous state and also brings in some various monolith styles overrides that were happening (e.g. like for `.CalendarDay`). With that, we should highly prioritize cleaning up the overrides outside of OUI and make OUI the complete source of truth.

_Before:_
image

_After:_
image

**Test Plan:**
- [ ] Manually verify tables and nested tables in Monolith and Program Management


src/components/CodeDiff/CodeDiff.js
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
```
circAssimilate: Just moved this to file to `CodeDiff.js` in troubleshooting to match some of the other component structures.

src/components/Table/Table.story.js
```diff
@@ -17,7 +20,14 @@ stories
 
 stories
   .add('Default', (() => (<div>
-    <Table density="loose" tableLayoutAlgorithm="fixed">
+    <Table
+      density={
+        select('density', {
+          'loose': 'loose',
```
daverau-optimizely: Once I move remaining Optly.css overrides into OUI this property will make more sense in Storybook

PR #1207, opened 2019-08-02
feat(SelectDropdown): Add ability to specify maxWidth

ayeshakmaz:
Truncate dropdowns and add a title for truncated select dropdowns


src/components/SelectDropdown/index.scss
```diff
@@ -0,0 +1,12 @@
+.oui-dropdown-group {
+  .oui-button--outline {
+    background: white;
+  }
+}
+
+.oui-dropdown-group__activator-label {
```
daverau-optimizely: I'll 👍 this so we can land it but I'd like to go back and extend %truncate just so the sass is more modular there.

```%truncate,
.truncate {
  white-space: nowrap;
  overflow: hidden;
  text-overflow: ellipsis;
}
```

PR #1208, opened 2019-08-02
Adds nearest design token colors to Storybook

daverau-optimizely:
This PR adds color analysis tooling (a no-risk change 🤞) to Storybook. It will loop over all non-token color values and finds the nearest color token:

- Original color on the left is a random color found in the CSS, along with number of times used
- Design system color token on the right, along with distance value 0-100 from original color

Screen Shot 2019-08-02 at 12 21 42 PM


.storybook/css.story.js
```diff
@@ -4,6 +4,36 @@ import { storiesOf } from '@storybook/react';
...
+const colorTokens = require('../src/tokens/src/color.json').tokens;
+var tokens = {};
+colorTokens.map(colorToken => (
+  tokens[colorToken.name] = colorToken.hex
+));
+var nearestColor = require('nearest-color').from(tokens);
+
+css.nearestColors = [];
+css.uniqueColors.map(function(c) {
+  if (c.color.length == 7) {
```
ayeshakmaz: 7 seems like a bit of a magic number here (guessing it's hex length?) Could you pull it out into a constant above with a name?

.storybook/css.story.js
```diff
@@ -4,6 +4,36 @@ import { storiesOf } from '@storybook/react';
...
+    var newColor = nearestColor(c.color);
+    if (newColor.distance > 0) {
+      newColor.original = c.color;
+      newColor.count = c.count;
+      newColor.distance = Math.round(newColor.distance);
+      css.nearestColors.push(newColor);
+    }
+  }
+});
+css.oneOfAKindColors.map(function(c) {
```
ayeshakmaz: What's the difference between unique colors and oneofakind?

.storybook/preview-head.html
```diff
@@ -27,4 +27,10 @@
     padding-top: 75px;
     border: 1px dashed #ccc;
 }
+
+/* .oui--swatches span:hover {
```
ayeshakmaz: Should we delete this?

.storybook/css.story.js
daverau-optimizely: This skips over alpha hex values. Good call.

.storybook/css.story.js
daverau-optimizely: unique colors has a count > 1, one of a kind == 1

PR #1212, opened 2019-08-08
feat(NavBar): Added a new Navigation bar component.

zashraf1985:
# Summary:
This PR contains a new component which renders the main left navigation bar. It exposes following components.
1. `Navbar` (The container in which primary links, secondary links and other components will reside)
2. `PrimaryLinks` - multiple instances passed as children
3. `SecondaryLinks` - multiple instances passed as children
4. `CurrentUserMenu` - One instance passed as a child.

Children will always be rendered in order irrespective of the order they were passed in. Primary links will be rendered first, then secondary links and then current user menu.

The idea behind this refactor is to make main navigation generic enough to be used in projects other than FE experimentation. Each project can use this component in its own customized way by implementing a wrapper that provides the necessary state values and customized components and/or dynamic links. This component is stateless and expects state values to be passed from the wrapper. 

Refer to the example in the story to understand how it works.

### For more context and detail please read the design doc
[Refactor Global Nav](https://docs.google.com/document/d/1-JGKxEh9an8d5Ic3Um3A7QpFx2aSSbSYBwAY50PnDWM/edit#)

# Test Plan:
Added Unit tests for `NavBar`, `IconLink`, `AccountSwitcher` and `CurrentUserMenu` components

# JIRA Issues:
[FEI-551](https://optimizely.atlassian.net/browse/FEI-551)


ayeshakmaz:
@zashraf1985 I'm a bit unclear on how we're proceeding with adding this to OUI so I'd love to get some clarity before reviewing more:

My thinking is that everything needed to recreate the Global Nav in any product (monolith, Program Mgt. etc.) would be included in OUI. Right now it feels like this component only gives me the shell, and then I'd have to do extra work to add the different types of links and style them properly. 

Given the breakdown in the [design doc](https://docs.google.com/document/d/1-JGKxEh9an8d5Ic3Um3A7QpFx2aSSbSYBwAY50PnDWM/edit#heading=h.z9tms3arjm1j), I was thinking something like: 

```
<Navbar>
    <Navbar.Header />
    <Navbar.PrimaryLink 
          name="Projects" 
          iconName="list" 
          url="www.google.com" 
         onClick={some function} />
     <Navbar.SecondaryLink 
          name="Help" 
          iconName="help" 
          url="www.help.google.com" 
         onClick={some function} />
      <Navbar.CurrentUserMenu/>
      {etc.}
</Navbar>
```

but it's entirely possible my mental model of this is way off, so I'm curious what @jamesopti and @zachpwr think as well

zashraf1985:
@ayeshakmaz your understanding is absolutetly correct. This is exactly how the final version will look like. This is an interim PR. Our plan is to break it down in multiple PRs, ideally 4, to avoid large diffs for reviewer's ease.
1. Skeleton and Header (This one)
2. Primary Links
3. Secondary Links
4. Current User Menu.

I agree that we should have been explicit about this in the PR description to avoid confusion. Let us know if you want us to do it in one single PR?

src/components/NavBar/index.js
```diff
@@ -0,0 +1,33 @@
...
+    children,
+  } = props;
+  return (
+    <ul
+      className={ classNames({
+        'root-nav': true,
+        'root-nav--open': isNavOpen,
+      }) }
+      data-test-section="p13n-root-navbar">
+      { children }
```
jamesopti: This looks pretty good, but I have a couple questions about how we'll validate & enforce what is rendered:

1. Since were just blindly rendering `{ children }`, what will happen if a consumer puts things in the wrong order?

Will the CSS ensure the layout works as expected?

E.g.:

```js
<Navbar>
  <SecondaryLinks />
  <NavbarHeader />
  <CurrentUserMenu>
    <MyUserComp />
  </CurrentUserMenu>
  <PrimaryLinks />
</Navbar>
```

2. Similarly, what if the consumer passes their own markup. Is that OK? Do we want to support that?
```js
<Navbar>
  <NavbarHeader />
  <div>
    <MyUserComp />
  </div>
</Navbar>
```

3. Another example of a violation might be passing more than one of our 4 section types:

```js
<Navbar>
  <NavbarHeader />
  <NavbarHeader />
  <PrimaryLinks />
  <PrimaryLinks />
</Navbar>
```

The renderProps approach handles this more carefully as it provides a defined interface for each allowed section.

Happy to chat about this on Slack if you want!

zashraf1985: @jamesopti Thanks for the feedback. I have made changes and now passing all information as props. Also added support for primary and secondary links. Now, only thing pending is the bottom section which contains current user related links. It would be great if you can take a look and give feedback

zashraf1985:
@quarklemotion @ayeshakmaz Thank you for the detailed feedback. I have incorporated changes to the best of my understanding. The CurrentUser component is still pending and we will work on that on monday. It would be great if you guys can take another look at this PR and let us know if we are moving forward in the correct direction or not.

src/components/NavBar/index.js
```diff
@@ -0,0 +1,165 @@
...
+};
+
+NavBar.propTypes = {
+  /** Text to appear below title in a badge. */
+  badgeText: PropTypes.string,
+  /* Allowed Children are <PrimaryLink>, <SecondaryLink>, <CurrentUser>. */
+  children: (props, propName) => {
+    const prop = props[propName];
+    let error = null;
+    React.Children.forEach(prop, (child) => {
```
jamesopti: very nice!

src/components/NavBar/IconLink/index.js
```diff
@@ -0,0 +1,162 @@
...
+        style="unstyled"
+        testSection={ `${testSection}-button` }>
+        { this.renderNavLink() }
+      </Button>
+    );
+  };
+
+  render() {
+    const { testSection, type } = this.props;
+    let primaryLink;
```
jamesopti: Can you name this more generically, like `linkToRender`, to avoid confusion with the term `primaryLink`?

src/components/NavBar/CurrentUserMenu/index.js
```diff
@@ -0,0 +1,222 @@
...
+    /** Account Log Out Url */
+    logoutUrl: PropTypes.string.isRequired,
+    /** Function called when Emulate is clicked */
+    onEmulateClick: PropTypes.bool.isRequired,
+    /** Account Profile Avatar Url */
+    profileAvatarUrl: PropTypes.string.isRequired,
+    /** Account Profile Url */
+    profileUrl: PropTypes.string.isRequired,
+    /** Show Emulate Link */
+    showEmulate: PropTypes.bool.isRequired,
```
jamesopti: default false?

src/components/NavBar/CurrentUserMenu/index.js
```diff
@@ -0,0 +1,222 @@
...
+     * isCurrent: Bool */
+    accountSwitcherItems: PropTypes.array.isRequired,
+    /** True if Navbar is Open, False if collapsed */
+    isOpen: PropTypes.bool,
+    /** Account Log Out Url */
+    logoutUrl: PropTypes.string.isRequired,
+    /** Function called when Emulate is clicked */
+    onEmulateClick: PropTypes.bool.isRequired,
+    /** Account Profile Avatar Url */
+    profileAvatarUrl: PropTypes.string.isRequired,
```
jamesopti: default to a generic image?

src/components/NavBar/index.js
```diff
@@ -0,0 +1,165 @@
...
+      error = new Error('There should be only one instance of `CurrentUser`');
+    }
+    return error;
+  },
+  /** Url to navigate to when Brand Logo is clicked. */
+  homeUrl: PropTypes.string,
+  /** Is Navigation Bar open or closed. */
+  isOpen: PropTypes.bool,
+  /** Component containing brand logo. */
+  logo: PropTypes.node,
```
jamesopti: @ayeshakmaz Do we want them to be able to pass their own content here?

daverau-optimizely: Allowing visible options between set choices would be fine. I would prefer to avoid props that let you dump arbitrary things that greatly impact the visual UI.

src/components/NavBar/index.scss
```diff
@@ -0,0 +1,222 @@
...
+  background: var(--grey-25);
+  width: 76px;
+  transition: width $animation-speed ease-in;
+  &:after {
+    transition: width $animation-speed ease-in;
+    content : '';
+    position: absolute;
+    top: 0px;
+    background-color: transparent;
+    bottom: 0px;
```
daverau-optimizely: `0` is preferred over `0px` just as a minor improvement here.

src/components/NavBar/CurrentUserMenu/index.js
daverau-optimizely: The CSS does this now for free, but perhaps it's more clear at the prop level.

src/components/NavBar/index.js
```diff
@@ -0,0 +1,165 @@
...
+    <ul
+      className={ classNames({
+        'root-nav': true,
+        'root-nav--open': isOpen,
+      }) }
+      data-test-section="p13n-root-navbar">
+      <div data-test-section="navbar-header">
+        <li className="root-nav__logo push-double--bottom">
+          <Link href={ homeUrl }>
+            { logo }
```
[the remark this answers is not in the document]
ayeshakmaz: Then we can insure all logos passed in are at least the same height and width and in the right location

src/components/NavBar/CurrentUserMenu/index.js
Hassan-Khalid-folio3: Yes @daverau-optimizely  in the old implementation of NavBar, CSS was handling this by using a default svg. But if I try to use an svg in OUI, i get build errors. So how can we use svgs in OUI?

src/components/NavBar/index.js
```diff
@@ -0,0 +1,165 @@
...
+  /* Allowed Children are <PrimaryLink>, <SecondaryLink>, <CurrentUser>. */
+  children: (props, propName) => {
+    const prop = props[propName];
+    let error = null;
+    React.Children.forEach(prop, (child) => {
+      if (![SecondaryLink, PrimaryLink, CurrentUserMenu].includes(child.type)) {
+        error = new Error('Children should be of type PrimaryLink, SecondaryLink, or CurrentUser.');
+      }
+    });
+    if (!error && React.Children.toArray(prop).filter(child => child.type === CurrentUserMenu).length > 1) {
```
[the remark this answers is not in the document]
Hassan-Khalid-folio3: Thanks for pointing this out. I have modified this check. We will now get an error if there is 0 or more than one CurrentUserMenu component.

ayeshakmaz:
@jamesopti @daverau-optimizely Created https://optimizely.atlassian.net/browse/UXENG-33

src/components/NavBar/CurrentUserMenu/index.js
```diff
@@ -0,0 +1,223 @@
...
+    </span>
+    <div className="flex--1 push-half--left">
+      <span className="oui-arrow-inline--down oui-arrow-inline--small vertical-align--middle" />
+    </div>
+  </div>
+);
+
+Activator.propTypes = {
+  buttonRef: PropTypes.func,
+  onBLur: PropTypes.func,
```
ayeshakmaz: Slight typo with the capital L

src/components/NavBar/CurrentUserMenu/index.js
```diff
@@ -0,0 +1,223 @@
...
+    /** Account Settings Url */
+    accountSettingsUrl: PropTypes.string.isRequired,
+    /** Function Called to Switch Account */
+    accountSwitcherHandler: PropTypes.func.isRequired,
+    /** An array of elements containing the following account data
+     * text: String
+     * url: String
+     * description: String
+     * isCurrent: Bool */
+    accountSwitcherItems: PropTypes.array.isRequired,
```
ayeshakmaz: I think there might be a way to specify the shape of each object you expect in the array

Hassan-Khalid-folio3: Your right. I have used ProtoType.shape()method to specify what type of objects the array should expect.

PR #1214, opened 2019-08-13
patch(Layout)[MGMT-2422] Export layout component

pauloptimizely:
## Summary

Export `Layout` component from main.js

I'm using this component to make the contents of MGMT table items responsive

## Test Plan
N/A

## Jira Issue
https://optimizely.atlassian.net/browse/MGMT-2422

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -29,6 +29,7 @@ export { default as HelpPopover } from './components/HelpPopover';
 export { default as ButtonIcon } from './components/ButtonIcon';
 export { default as Input } from './components/Input';
 export { default as Label } from './components/Label';
+export { default as Layout } from './components/Layout';
 export { default as Link } from './components/Link';
 export { default as ListGroup } from './components/ListGroup';
 export { default as OverlayWrapper } from './components/OverlayWrapper';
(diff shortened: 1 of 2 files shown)


ayeshakmaz:
@pauloptimizely What are you trying to export here? There isn't a singular <Layout> component, but rather <Col>, <Row>, and <Container>, which are already exported

src/main.js
```diff
@@ -29,6 +29,7 @@ export { default as HelpPopover } from './components/HelpPopover';
 export { default as ButtonIcon } from './components/ButtonIcon';
 export { default as Input } from './components/Input';
 export { default as Label } from './components/Label';
+export { default as Layout } from './components/Layout';
```
daverau-optimizely: Hi @pauloptimizely — what drove this change? I wouldn't expect this to export anything since there is no named export called Layout, only the inner components like `<Col>`, `<Row>`, and `<Container>` — is something else at play? Thanks for clarifying!

pauloptimizely: Oh sorry about that, I wasn't aware of the `Col`, `Row` and `Container` components 🤦‍♂. Thanks for the info!!

PR #1220, opened 2019-09-16
WIP fitted sheet

daverau-optimizely:
Fitted sheet aims to solve the problem where long forms push the save/confirm buttons "below the fold" and makes them difficult to discover. This is a work-in-progress attempt to re-use the sheet component without the outer most layout wrapper styles being applied.

fitted_sheet


daverau-optimizely:
Skipping this in favor of DockedFooter

PR #1224, opened 2019-10-03
security bump for lodash

daverau-optimizely:
1. Bump for lodash
1. Also includes running `yarn-deduplicate yarn.lock`

Screen Shot 2019-10-03 at 10 49 34 AM 1

PR #1227, opened 2019-10-09
fix (Security) Update lodash to resolve security issues

kwalker3690:
OUI has a security warning issued:
image

This diff specifies that lodash cannot be set to lower than the known good version. We had the known good version specified in our checked in yarn.lock so this is not too worrisome, but this will prevent any backslip.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "handlebars": "^4.0.14",
     "highlight.js": "9.5.0",
     "immutable": "3.x.x",
-    "lodash": "4.x.x",
+    "lodash": ">=4.17.13",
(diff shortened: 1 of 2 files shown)


daverau-optimizely:
Hey @kwalker3690 — i'm curious what your process for upgrading this was? I wanted to update the contributing docs with some instructions as I tried this two different ways and ran into issues both ways. I was expecting `yarn upgrade lodash --latest`

There was an existing PR for this already (https://github.com/optimizely/oui/pull/1224) so I just wanted to be sure there was more of a process for fixing these types of issues in the future. Thanks!

PR #1228, opened 2019-10-09
fix (Dropdown) Use proper value for maxHeight style in Dropdown.Contents

kwalker3690:
Dropdown.Contents has a prop canScroll (which actually is implemented backwards, it would be more correct to call it cannotScroll), which forces the contents to not scroll when true. One part of this is to override the max height css property of oui-dropdown. This was previously implemented with maxHeight=auto, which is not a valid value for max height, so it wasn't being applied as a style.
 
image (4)

This popped up as an issue in this UI, visible when the hover state was misaligned with the dropdown:
image


benjaminpadula:
Thanks!

PR #1230, opened 2019-10-09
fix [DESIGN-630] TokensInput formatting

pauloptimizely:
## Summary

TokensInput component is visually misaligned. I made the follow UI changes:

1. Centered the tokens-list and input within the bordered container
1. Added margin between the internal-input and the tokens list
1. Added a border radius to the internal-input

**Before**

Screen Shot 2019-10-09 at 5 25 31 PM

**After**

Screen Shot 2019-10-09 at 5 25 45 PM

## Test Plan

Visually tested my changes in storybook

## Jira Issues

https://optimizely.atlassian.net/browse/DESIGN-630


src/components/TokensInput/index.js
```diff
@@ -136,16 +136,17 @@ export const TokensInput = ({
...
   return (
-    <div className="oui-text-input text--left flush">
+    <div className="oui-tokens-input">
       <ReactTagsInput
         addKeys={ addKeys }
         addOnBlur={ addOnBlur }
         addOnPaste={ addOnPaste }
         inputProps={{
-          className: `flex flex--1 ${minWidth} no-border soft-half--ends soft--sides`,
+          className: classNames('oui-tokens-input__input', {
```
daverau-optimizely: No change here, just a comment...
If I were doing this from scratch myself I'd probably favor the more atomic css style of class construction as previous `flex flex--1 ${minWidth} no-border soft-half--ends soft--sides` vs increasing the overall CSS bundle (even if it's a small amount) and that probably shows a need for some documentation around how we approach authoring CSS from scratch vs when to use atomic helper classes.

daverau-optimizely:
@pauloptimizely just curious if you noticed anything odd here that would lead you to discover the root cause of this regression? I'm just thinking how we could have avoided this.

ayeshakmaz:
@daverau-optimizely Visual Regression Testing! This is the exact type of thing it would catch

pauloptimizely:
@daverau-optimizely We recently added a Tokens autocomplete component to Program Management and I noticed this regression when I upgraded from OUI version`44.1.0` to `44.9.0`.

I don't currently have a good solution for automated visual regression testing, but the MGMT FE team will currently perform a manual visual regression test each time we upgrade any of our FE libraries. Usually something minor will popup in the process of upgrading

src/components/TokensInput/index.js
pauloptimizely: I'm up for using atomic css! The reason I migrated to a custom class was b/c, I wasn't able to find an atomic rule for `height: auto` or `border-radius`. I'd be happy to add in those rules if atomic css is preferred

PR #1231, opened 2019-10-09
feat(sortable): allow for dynamic list

ayeshakmaz:
This PR contains changes needed to support a dynamic sortable list and new draggable token design

1) Update Sortable component if default value changes
2) Add option to use hamburger for draggable token


src/components/Token/index.js
```diff
@@ -124,6 +160,9 @@ Token.propTypes = {
 
   /** Hook for automated JavaScript tests */
   testSection: PropTypes.string,
+
+  /** Used to switch drag handle */
+  usesHamburger: PropTypes.bool,
```
daverau-optimizely: I might be inclined to use a different name here instead of `hamburger`, otherwise LGTM

PR #1240, opened 2019-10-25
Ayesha/upgrade storybook 5.2

ayeshakmaz:
Upgrade to storybook 5.2
React to 16.8
React-dom to 16.8


circAssimilate:
You pointed out that the storybook build happens automatically on release. That works for me, as long as we're ensuring this update doesn't break anything!

yarn.lock
```diff
@@ -11051,25 +10996,32 @@ react-with-styles@^3.2.0:
...
     object-assign "^4.1.1"
     prop-types "^15.6.2"
-    scheduler "^0.10.0"
+    scheduler "^0.13.6"
 
-"react@^15.6.2 || ^16.0", react@^16.7.0, react@^16.8.3, react@^16.8.4:
-  version "16.8.6"
-  resolved "https://registry.yarnpkg.com/react/-/react-16.8.6.tgz#ad6c3a9614fd3a4e9ef51117f54d888da01f2bbe"
-  integrity sha512-pC0uMkhLaHm11ZSJULfOBqV4tIZkx87ZLvbbQYunNixAAvjnC+snJCg0XQXn9VIsttVsbZP/H/ewzgsd5fxKXw==
+react@^0.14.0:
```
quarklemotion: I'm curious why these upgrades are somehow leading to an extra **very old** version of react being pulled in?

jamesopti:
> I just wanna do sanity check that even though these are only devDependencies in package.json, no subdependencies will hurt the actual OUI components. 

> LGTM, I assume the added older react version is only a dev dependency and should not increase the bundle size when pulled into the monolith.

Couple thoughts here.

1 - React is most certainly a production dependency here. It is needed at runtime for everything inside `src/components` and as such, I think should be listed here as a peerDependency (read more about that in [this great SO post](https://stackoverflow.com/questions/26737819/why-use-peer-dependencies-in-npm-for-plugins/34645112#34645112)) so that consumers are aware that they must include a version of React that meet's this repo's needs.

2 - Since it's currently listed as a `devDependency`, the React version being used is subject to whatever happens to be installed by the consumer (which ends up being fine for our 2 repos).

> I'm curious why these upgrades are somehow leading to an extra very old version of react being pulled in?

I agree with this - We should figure out what's going on here before merging.

yarn.lock
jamesopti: DId you run `yarn-deduplicate yarn.lock` after this change?

This is indeed strange. We should figure this out.

ayeshakmaz: I did, very odd that this got added, let me try again

quarklemotion: It looks like `@storybook/addon-info` is pulling in `jsx-to-string` which has a react dependency of`^0.14.0`.  I would expect this to dedupe to latest react 0.16.x.  It is possible you may need to add a `resolves` entry in package.json to force the dedupe if it is not happening automatically/naturally.  We had to do this with Immutable.js in the monolith.

quarklemotion:
@jamesopti brings up a good point - react should either be listed as a `peerDependency` or be moved to `dependencies`

ayeshakmaz:
@quarklemotion Let me know if this is what you had in mind

ayeshakmaz:
Alright, not sure what's happening with the tests now 😭

ayeshakmaz:
Storybook has already moved to 5.3, so closing this and making a new PR

PR #1245, opened 2019-11-04
fix(BlockList/FilterPicker): Add hasFauxFocus prop for block list items

ayeshakmaz:


src/components/FilterPicker/FilterPickerListItem/index.js
```diff
@@ -54,6 +55,8 @@ ListItem.propTypes = {
   buttonText: PropTypes.string,
   /* Description of ListItem */
   description: PropTypes.string,
+  /* Boolean to set faux focus of a list item */
+  hasFauxFocus: PropTypes.bool,
```
jamesopti: Since its optional, can we put a default here?

PR #1247, opened 2019-11-06
feature (docked-footer) [FEI-3319] adds new docked-footer component

acacia-m-pappas:
## Summary: 
Optimizely customers have expressed frustration ([FEI-3319](https://optimizely.atlassian.net/browse/FEI-3319)) at not being able to find the `Save Button` on certain pages, like the Metrics Page and Audiences Page. This new `<DockedFooter>` component creates a solution for them. 

The `<DockedFooter>` adds styles to the ButtonRow such that the `Save Button` _sticks_ to the bottom of the browser window if the the content is larger/longer than the window size. If the content is smaller/shorter than the window size, then transforms the footer undocks and transforms into `<ButtonRow>`.

### Pitfalls in Implementing `<DockedFooter>` in the `Optimizely Frontend`:

This component uses event listeners to check scroll height and window size and then dock (_aka stick_ )or not dock. The scroll height check depends on _where_ you set `parentTestSection` id. You will have to set this prop carefully. We noticed in implementing this component on the `Metrics Page` (`/metrics_manager_react/index.js`) that this component works best when the `parentTestSection` is set as an `data-test-section` on the same `<div>` as the `overflow-y-auto` so that the event listeners are set on the same div that is being scrolled. 

**Note**: If you set the `parentTestSection` on a `<div>` that _does not scroll,_ this component will not work as expected.

### Implementing `<DockedFooter>`: 

Add a `data-test-section="docked-footer-on-Metrics-page"` as an html attribute to the parent `div` or component that you would like the `scroll` event listener to be set on. Something like this: 
```
 <div data-test-section="docked-footer-on-Metrics-page">
```

Pass the same string as the `parentTestSection` prop to the `<DockedFooter>`: 

```    
<DockedFooter
   parentTestSection="docked-footer-on-Metrics-page"
   includesMargin={true}
   isDocked={true}>
   {buttons as children passed here}
<DockedFooter/> 
```
Now the `<DockedFooter>` will set the scroll listener wherever you set the `data-test-section`. 

To view it in context, go take a look at the `src/js/bundles/p13n/components/metrics/metrics_manager_react/index.js` in the Optimizely Frontend. 

## JIRA Issues: 
[FEI-3319](https://optimizely.atlassian.net/browse/FEI-3319)


src/components/DockedFooter/index.js
```diff
@@ -0,0 +1,129 @@
...
+      this.shouldDock();
+    } else if (prevState.viewableArea !== this.state.viewableArea) {
+      this.shouldDock();
+    }
+  }
+
+  shouldDock() {
+    let footerToTop, viewableArea;
+    const parentElement = document.querySelector('[data-test-section="' + this.props.parentTestSection + '"]');
+    const footerElement = document.getElementsByClassName('oui-docked-footer')[0];
```
ayeshakmaz: Could we make `testSection` prop required and use that here, instead of hoping there's only one docked footer?

src/components/DockedFooter/index.js
```diff
@@ -0,0 +1,129 @@
...
+        this.setState({isDocked: false});
+      }
+    }
+
+    this.onScroll();
+  }
+
+  onScroll() {
+    const parentElement = document.querySelector('[data-test-section="' + this.props.parentTestSection + '"]');
+    let atBottom = parentElement.scrollHeight - parentElement.scrollTop === parentElement.clientHeight;
```
ayeshakmaz: Could you rename `atBottom` to be more descriptive? What is at the bottom here?

src/components/DockedFooter/index.js
```diff
@@ -0,0 +1,129 @@
...
+
+    if (atBottom) {
+      this.setState({isDocked: false});
+    } else if (!atBottom) {
+      this.setState({isDocked: true });
+    }
+  }
```
ayeshakmaz: This could be simplified to set the state to the opposite of the at bottom prop

src/components/DockedFooter/index.js
```diff
@@ -0,0 +1,129 @@
...
+        const now = (new Date()).getTime();
+        if (now - lastCall < delay) {
+          return;
+        }
+        lastCall = now;
+        return fn(...args);
+      };
+    };
+
+    window.addEventListener('resize', throttle(10, this.shouldDock));
```
ayeshakmaz: Could you pull this constant out to the top?

src/components/DockedFooter/index.js
```diff
@@ -0,0 +1,129 @@
...
+        }) }>
+        <ButtonRow rightGroup={ this.props.children }/>
+      </footer>
+    );
+  }
+}
+
+DockedFooter.propTypes = {
+  /**
+   *  The body of the sheet to request information and data from the user.
```
ayeshakmaz: This prop description doesn't match

src/components/DockedFooter/index.js
```diff
@@ -0,0 +1,129 @@
...
+   */
+  children: PropTypes.node,
+  /**
+  *  Used to determine if there should be top margin.
+   */
+  includesMargin: PropTypes.bool,
+  /**
+   *  Determins if the footer should be docked or not.
```
ayeshakmaz: Typo on `determins`

src/components/DockedFooter/tests/index.js
```diff
@@ -0,0 +1,31 @@
...
+                Confirm
+            </Button>,
+          ] }
+        </DockedFooter>
+      </div>
+    );
+    expect(component.find('.oui-button').length).toBeGreaterThanOrEqual(1);
+  });
+
+});
```
ayeshakmaz: Could you add a few more tests to make sure the appropriate classes are being applied based on props passed in? Would also be ideal to have a test to test the scrolling behavior

src/oui/_oui-partials.scss
```diff
@@ -22,6 +22,7 @@
 @import 'partials/elements/fonts';
 @import 'partials/elements/typography';
 @import 'partials/elements/form';
+@import '../components/DockedFooter/index';
```
ayeshakmaz: Could we move this to the bottom portion of the file for consistency? Is there a reason for adding it here?

src/components/DockedFooter/index.js
```diff
@@ -0,0 +1,129 @@
...
+    const footerElement = document.getElementsByClassName('oui-docked-footer')[0];
+
+    if (footerElement !== undefined && parentElement !== undefined) {
+      footerToTop = footerElement.offsetTop;
+      viewableArea = parentElement.offsetHeight - footerElement.offsetHeight;
+      if (footerToTop >= viewableArea) {
+        this.setState({isDocked: true});
+      } else {
+        this.setState({isDocked: false});
+      }
```
ayeshakmaz: This could also be simplified to one setState call

src/components/DockedFooter/index.js
```diff
@@ -0,0 +1,112 @@
...
+    this.onScroll();
+  }
+
+  onScroll() {
+    const parentElement = document.querySelector('[data-test-section="' + this.props.parentTestSection + '"]');
+    let atScrollBottom = parentElement.scrollHeight - parentElement.scrollTop === parentElement.clientHeight;
+    this.setState({isDocked: !atScrollBottom});
+  }
+
+  throttle(delay, fn) {
```
acacia-m-pappas: @ayeshakmaz should this be moved to _something like_ OUI /utils?

ayeshakmaz: Yeah I could see this being useful! maybe a `throttle.js` file?

acacia-m-pappas: I chose not move it because my component seems to be the only one using event listeners for `resize` and `scroll`, if other components are added using this kind of thing then it would make sense to move it to a place where it can be shared.

src/components/DockedFooter/tests/index.js
```diff
@@ -0,0 +1,113 @@
...
+          <DockedFooter
+            testSection="docked-footer-less-content"
+            parentTestSection="helloTesting"
+            includesMargin={ true }
+            isDocked={ false }>
+            { buttonChildren }
+          </DockedFooter>
+        </div>, { attachTo: window.domNode }
+      );
+      expect(component.find('.oui-docked-footer').length).toBeGreaterThanOrEqual(0);
```
ayeshakmaz: I'm a bit confused how this is passing, shouldn't we be looking for the `is-docked` class, while the docked-footer class is still present?

src/oui/_oui-partials.scss
```diff
@@ -31,6 +31,7 @@
 @import '../components/Select/index';
 @import '../components/Table/index';
 @import 'partials/elements/text';
+@import '../components/DockedFooter/index';
```
ayeshakmaz: Could you actually move this below DiscloseTable on line 67?

src/components/DockedFooter/index.js
```diff
@@ -0,0 +1,113 @@
...
+   */
+  children: PropTypes.node,
+  /**
+  *  Used to determine if there should be top margin.
+   */
+  includesMargin: PropTypes.bool,
+  /**
+   *  Determines if the footer should be docked or not.
+   */
+  isDocked: PropTypes.bool,
```
ayeshakmaz: Is this actually used anywhere?

src/components/DockedFooter/tests/index.js
ayeshakmaz: Actually, is the isDocked prop actually doing anything?

PR #1259, opened 2019-12-03
Migrate Legacy CSS from Optly.git to OUI

daverau-optimizely:
# Background
Since @jamesopti ripped out Classic from Optimizely.git there was never a parallel effort to rip out legacy CSS from the app — ⚡️UNTIL NOW! ⚡️ 

This PR aims to migrate in-use CSS from Optimizely.git into OUI when it makes sense. Some of these are mixin files we should be using inside OUI, some of these are older patterns that are in use that never made it back into OUI, and a few are just helpful utilities that should live in OUI as the source of truth vs the monolith.

To be paired with this [Optimizely.git PR](https://github.com/optimizely/optimizely/pull/11272/) that removes over 7k lines of CSS debt from Classic and P13N, along with adding a few system-wide helpers from the app into OUI for better reuse.

## Migrate reusable helpers & mixins
- .border-dashed--all helper
- Pattern mixin for editor
- utilities (size, square, etc)
- no-select
- math functions
- lists
- keyframes

## Migrate style overrides
- Table defaults


src/oui/partials/sass/_mixins.scss
```diff
@@ -2,6 +2,12 @@
 /// Mixins
 ////
 
+@import 'mixins/keyframes';
```
ayeshakmaz: Loving this new mixin pattern!!

There are many form mixins in /oui/src/oui/partials/elements/_form.scss, should those also be separated out into a new file?

src/oui/partials/sass/mixins/_no-select.scss
```diff
@@ -0,0 +1,9 @@
+@mixin no-select {
```
ayeshakmaz: Might be good to add a comment above here explaining this mixin's purpose (similar to the others)

src/oui/partials/sass/_mixins.scss
daverau-optimizely: Those feel less like generic utilities that are reusable and more like input-specific CSS that we are also allowing to be consumed via mixin so it's easily extended. I would say not in this case.

src/oui/partials/sass/mixins/_no-select.scss
daverau-optimizely: Good call, did some cleanup on header comments.

PR #1261, opened 2019-12-06
feat(DragAndDrop): New component to handle dynamic drag and drop

ayeshakmaz:
New Drag and Drop component using react-beautiful-dnd that allows for dynamic lists, interactive elements inside the draggable items, and keyboard accessibility!

JIRA:
[UXENG-53](https://optimizely.atlassian.net/browse/UXENG-53)


src/components/DragAndDrop/index.js
```diff
@@ -0,0 +1,113 @@
...
+    this.state = {
+      items: this.props.items,
+    };
+    this.onDragEnd = this.onDragEnd.bind(this);
+
+  }
+
+  componentDidUpdate(prevProps, prevState) {
+    if (prevProps.items !== this.props.items) {
+      this.setState({
```
kwalker3690: I'm wondering if we should be keeping track of the order of the items or something. Just thinking that this could override any reordering that has already been done, creating a weird user experience

src/components/DragAndDrop/index.js
```diff
@@ -0,0 +1,113 @@
...
+    newOrderOfItems.splice(destination.index, 0, removedItem[0]);
+
+    this.setState({
+      items: newOrderOfItems,
+    });
+
+    this.props.onDragEnd();
+  }
+
+  renderDraggableItems(item) {
```
kwalker3690: nit: argument should be plural to represent list

src/components/DragAndDrop/index.js
```diff
@@ -0,0 +1,113 @@
...
+  /**
+   * ID used for this Droppable region
+   */
+  idForDroppableRegion: PropTypes.string.isRequired,
+  /**
+   * Array of items to render as DraggableItems
+   */
+  items: PropTypes.array.isRequired,
+  /**
+   * Function to perform an action before item dimensions are captured
```
kwalker3690: What are 'item dimensions'?

src/components/DragAndDrop/index.js
```diff
@@ -0,0 +1,113 @@
...
+
+DragAndDrop.propTypes = {
+  /**
+   * ID used for this Droppable region
+   */
+  idForDroppableRegion: PropTypes.string.isRequired,
+  /**
+   * Array of items to render as DraggableItems
+   */
+  items: PropTypes.array.isRequired,
```
jamesopti: Can we add a definition for the shape of `items`? I believe its an object that at a minimum has an `id` property.

src/components/DragAndDrop/index.js
```diff
@@ -0,0 +1,113 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { DragDropContext, Droppable } from 'react-beautiful-dnd';
+import DraggableItem from './DraggableItem';
+
+class DragAndDrop extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
```
jamesopti: What is the value here in making this component stateful?

If a consumer must react to the order of items changing anyway, I think it makes more sense for this component to be a stateless UI interface for helping present and change the order of a list.

The consumer can be notified of any changes via drag/drop and then store that state, passing the new list in as props.

src/components/DragAndDrop/index.js
```diff
@@ -0,0 +1,113 @@
...
+    const currentItems = this.state.items;
+    const newOrderOfItems = Array.from(currentItems);
+    const removedItem = newOrderOfItems.splice(source.index, 1);
+    newOrderOfItems.splice(destination.index, 0, removedItem[0]);
+
+    this.setState({
+      items: newOrderOfItems,
+    });
+
+    this.props.onDragEnd();
```
jamesopti: How does the consumer know what the new order is?

src/components/DragAndDrop/index.js
```diff
@@ -0,0 +1,113 @@
...
+      });
+    }
+  }
+
+  render() {
+    return (
+      <DragDropContext onBeforeCapture={ this.props.onBeforeCapture } onDragEnd={ this.onDragEnd }>
+        <Droppable droppableId={ this.props.idForDroppableRegion }>
+          {provided => (
+            <div
```
jamesopti: Do we have to nest this inside a div here?

src/components/DragAndDrop/index.js
```diff
@@ -0,0 +1,113 @@
...
+              className="oui-sortable"
+              ref={ provided.innerRef }
+              { ...provided.droppableProps }>
+              <ul>{this.renderDraggableItems(this.state.items)}</ul>
```
jamesopti: Should we require that `renderItem` return an `<li>` as the root component?

yarn.lock
```diff
@@ -886,6 +886,14 @@
     "@babel/helper-plugin-utils" "^7.0.0"
     "@babel/plugin-transform-typescript" "^7.3.2"
 
+"@babel/runtime-corejs2@^7.6.3":
```
jamesopti: reminder to run `yarn-deduplicate yarn.lock` here.

src/components/DragAndDrop/index.js
ayeshakmaz: the dimensions of the draggable item (essentially bounding box around each draggable item so it knows where to put the white space when you drag)

src/components/DragAndDrop/index.js
ayeshakmaz: renderItem doesn't need to be an `li` since the DraggableItem ensures that

src/components/DragAndDrop/index.js
ayeshakmaz: Sadly yes : (

ayeshakmaz:
@kwalker3690 @jamesopti Made updates to no longer require state for this component, let me know what ya'll think, thank you both!

src/components/DragAndDrop/DraggableItem/index.js
```diff
@@ -0,0 +1,46 @@
...
+  render() {
+    const { id, index, item, renderFunc } = this.props;
+    return (
+      <Draggable key={ item.id } draggableId={ id.toString() } index={ index }>
+        {(provided, snapshot) => (
+          <li
+            className="oui-sortable__item flex"
+            ref={ provided.innerRef }
+            { ...provided.draggableProps }
+            { ...provided.dragHandleProps }>
```
davemn: What do you think about conditionally letting children receive `provided.dragHandleProps` so that a consumer can optionally customize which descendent becomes the drag handle? Something like:

```
<li
  className="oui-sortable__item flex"
  ref={ provided.innerRef }
  { ...provided.draggableProps }
  { ...(useCustomDragHandle ? {} : provided.dragHandleProps) }>
  {renderFunc({ item, index, snapshot, dragHandleProps: provided.dragHandleProps })}
</li>
```

ayeshakmaz: Ooo sure! I'll make that update

src/components/DragAndDrop/index.js
```diff
@@ -0,0 +1,100 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { DragDropContext, Droppable } from 'react-beautiful-dnd';
+import DraggableItem from './DraggableItem';
+
+class DragAndDrop extends React.Component {
+  constructor(props) {
```
jamesopti: You can just use a class property with arrow function for this instead of constructor:

`onDragEnd = (result) => {...`

src/components/DragAndDrop/index.js
```diff
@@ -0,0 +1,100 @@
...
+
+    const currentItems = this.props.items;
+    const newOrderOfItems = Array.from(currentItems);
+    const removedItem = newOrderOfItems.splice(source.index, 1);
+    newOrderOfItems.splice(destination.index, 0, removedItem[0]);
+
+    this.props.onDragEnd(newOrderOfItems);
+  }
+
+  renderDraggableItems(items) {
```
jamesopti: This function feels a bit unnecessary and makes the reader have to jump one more step to see whats going on.

Can we just inline it below? I dont think we need to verify its an Array either.

```js
items.map((nestedItem, index) => (
          <DraggableItem
            key={ nestedItem.id }
            item={ nestedItem }
            id={ nestedItem.id }
            index={ index }
            renderFunc={ this.props.renderItem }
          />
);
```

src/components/DragAndDrop/DraggableItem/index.js
```diff
@@ -0,0 +1,46 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { Draggable } from 'react-beautiful-dnd';
+
+class DraggableItem extends React.Component {
```
jamesopti: Can you make this a functional component?

src/components/DragAndDrop/index.js
```diff
@@ -0,0 +1,100 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { DragDropContext, Droppable } from 'react-beautiful-dnd';
+import DraggableItem from './DraggableItem';
+
+class DragAndDrop extends React.Component {
```
jamesopti: same re: Functional component

src/components/DragAndDrop/index.js
```diff
@@ -0,0 +1,100 @@
...
+    if (
+      destination.droppableId === source.droppableId &&
+      destination.index === source.index
+    ) {
+      return;
+    }
+
+    const currentItems = this.props.items;
+    const newOrderOfItems = Array.from(currentItems);
+    const removedItem = newOrderOfItems.splice(source.index, 1);
```
jamesopti: Nit - since we expect exactly 1 result here, i think this is more clear as:

```js
const [removedItem] = newOrderOfItems.splice(source.index, 1);
newOrderOfItems.splice(destination.index, 0, removedItem);
```

src/components/DragAndDrop/DraggableItem/index.js
jamesopti: 👍

src/components/DragAndDrop/DragAndDrop.story.js
```diff
@@ -0,0 +1,44 @@
...
+      showWell={ false }
+    />
+  </div>
+);
+const stories = storiesOf('DragAndDrop', module);
+stories
+  .addDecorator(withKnobs)
+  .addDecorator(story => <div id="root-preview">{story()}</div>);
+
+stories.add('Reordering flat list', () => {
```
jamesopti: I think this is an example that would really benefit from a stateful wrapper (so the dropping part actually takes effect).

Could you add a simple one here?

ayeshakmaz:
@jamesopti Made those updates, thanks for the great suggestions! Could you take another look when you get a chance while I can finish up tests?

src/components/DragAndDrop/index.js
```diff
@@ -0,0 +1,100 @@
...
+const DragAndDrop = ({
+  idForDroppableRegion,
+  items,
+  onBeforeCapture,
+  onDragEnd,
+  renderItem,
+  useCustomDragHandle,
+}) => {
+
+  function updateOrderOfItems(result) {
```
jamesopti: Since this function only has 1 external dependency (`props.onDragEnd`), can you use `useCallback`?

```js
import { useCallback } from 'react';
...
const updateOrderOfItems = useCallback(result => {
  // update items code
}, [onDragEnd])
```

The performance benefit in the wild is likely minimal, but its a good practice :)

src/components/DragAndDrop/tests/index.js
```diff
@@ -0,0 +1,61 @@
...
+
+const itemsWithoutGroups = [
+  { id: 1, type: 'item', text: 'Item A' },
+  { id: 2, type: 'item', text: 'Item B' },
+  { id: 3, type: 'item', text: 'Item C' },
+  { id: 4, type: 'item', text: 'Item D' },
+  { id: 5, type: 'item', text: 'Item E' },
+];
+
+describe('components/DragAndDrop', function() {
```
jamesopti: Ahhh, testing drag and drop components!

Are there any good insights within the tests in this library's source that we could lean on?

Could likely copy some of theirs and tweak them for our wrapper.

ayeshakmaz: I took a look at their testing and they've created their own system for it, I'll come back to the tests, but get this merged so that @davemn can merge his frontend PR

jamesopti: Id prefer we wrote some tests for this before merging as a standard best practice, but will defer to @kwalker3690 to make that call for her team.

PR #1263, opened 2019-12-09
feat(typo): Add class for word-break word only

ayeshakmaz:
Added `word-break--word` to break on words vs. class `word-break` which is set to `break-all`

Added `word-break--all` so we can deprecate the confusing `word-break`

Needed for Jira: [ML-1532](https://optimizely.atlassian.net/browse/ML-1532)

Jira 
[AXIOM-137](https://optimizely.atlassian.net/browse/AXIOM-137)


ayeshakmaz:
@daverau-optimizely Let me know what you think

PR #1264, opened 2019-12-10
fix (Input): Restore `focus` prop that was inadvertently removed

jamesopti:
## Summary

This appears to fix the input issues in the monolith. I forgot to pass the `focus` prop in my refactor.


src/components/Input/index.js
```diff
@@ -11,32 +11,34 @@ import PropTypes from 'prop-types';
  */
 const Input = React.forwardRef((props, ref) => {
   const {
-    label,
+    defaultValue,
```
jamesopti: The only change here is adding `focus` back.

The rest is alpha ordering

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ This file is similar to the format suggested by [Keep a CHANGELOG](https://githu
 ## Unreleased
 - [Patch] Undo parent prop inheritance on Activator in **Dropdown** ([#1262])(https://github.com/optimizely/oui/pull/1262)
 - [Feature] Add typography helper classes `word-break--all` and `word-break--word` ([#1263])(https://github.com/optimizely/oui/pull/1263)
+- [Patch] Restore `<Input>` prop `focus`, which was inadvertently removed in #1256. ([#1264])(https://github.com/optimizely/oui/pull/1264)
```
ayeshakmaz: Quick nit to change `<Input>` to **Input**

PR #1268, opened 2019-12-20
chore (Typescript): Enable Typescript and auto generate declaration files

jamesopti:
## Summary

This PR enables Typescript support in OUI. One of the nice things about Typescript is that you can ease into it at whatever pace you like while still maintaining working code.

### Core Changes

- Upgrade Jest to v25.1 (necessary for proper TS support)
- Enable `.tsx` files to be compiled
-- Changes made to support the core webpack config as well as Storybook
- Add @types packages for core 3rd party libraries being used (like React)
- Convert a simple component (`<Attention>`) to use types
-- This is what the simplest conversion to Typescript would look like
- Add the babel-plugin [typescript-to-proptypes](https://www.npmjs.com/package/babel-plugin-typescript-to-proptypes), which autogenerates PropTypes from Typescrip types
-- This supports backwards compatibility for all consumers who dont use TS
- Generate an `optimizely-oui` declaration file using the `tsc` cli
-- the NPM script `yarn generate-types` runs this command and merges it with manually defined named exports inside the `declare module "optimizely-oui" {` block of `oui-module.d.ts`


types/oui-module.d.ts
```diff
@@ -0,0 +1,73 @@
+// Type definitions for OUI
+// Project: OUI
+// Definitions by: ui-engineering <https://github.com/optimizely/oui>
+
+/// <reference types="optimizely-oui" />
+
```
pauloptimizely: I noticed the following components are missing, I'm assuming they're deprecated, but I'm going to list them anyways, just in case:

Array, CloseButton, Help, Icon, InputIcon, Media, Nav, Search, Stats

data/csswhat/oui.css
```diff
@@ -4414,6 +4414,130 @@ code {
   border: 1px solid #e0e0e0;
   padding: 1px 3px; }
 
+/** START - CodeDiffLegacy styles */
```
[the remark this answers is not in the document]
jamesopti: Im going to revert this. Apparently these dont matter anymore?

types/oui-module.d.ts
jamesopti: GOOD EYE! I copied this list from `src/main.js`

@ayeshakmaz do you know if there is an intentional inconsistency?

types/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+I'd love if there was a cleaner way to do this!
+
+## Adding or Updating Types
+
+DO NOT MODIFY `index.d.ts` directly! It is an autogenerated file per above.
+
+To add/edit component definitions, you should only change the module definition template (`templates/module-declaration.d.ts`).
+
+`oui.d.ts` contains Typescript definitions specfically necessary for working inside this repo.
```
ayeshakmaz: Nit: misspelling of `specifically`

types/oui-module.d.ts
ayeshakmaz: These are all CSS inly "components" so I think it makes sense that they're not included here

PR #1270, opened 2019-12-31
feat(Dropdown): Add ability to render label version easily

ayeshakmaz:
Add ability to easily create labeled Dropdown style:

Screen Shot 2019-12-30 at 4 15 27 PM


ayeshakmaz:
A proposal for expanding the `buttonContent` prop to accept an object containing `label` and `content` -  but not married to this solution, would love feedback/thoughts

PR #1272, opened 2020-01-09
feat(layout): updates to vertically align a row and add scroll to col

ayeshakmaz:
Update `Row` to support verticalAlignment prop

Update `Col` to support overflow prop 

Jira:
[UXENG-62](https://optimizely.atlassian.net/browse/UXENG-62)

### DIFF
--- a/src/components/Layout/Col.js
+++ b/src/components/Layout/Col.js
@@ -83,7 +83,6 @@ const propTypes = {
    * Force overflow scrolling
    */
   overflow: PropTypes.oneOf([
-    'none',
     'overflow-y--scroll',
     'overflow-x--auto',
     'overflow-y--auto',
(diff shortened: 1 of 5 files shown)


src/components/Layout/Col.js
```diff
@@ -112,6 +122,11 @@ const Col = React.forwardRef(
       classes.push(`border--${border}`);
     }
 
+    if (overflow) {
+      classes.push(overflow);
+      classes.push('height--1-1');
```
pauloptimizely: I haven't tested this, so i'm not sure what it would look like, but should we still force full-height if overflow is equal to `none`? @ayeshakmaz

ayeshakmaz: Good call, I'm actually going to make an update to remove none as an option, since it's not a logical class name anyway

PR #1276, opened 2020-01-17
feat (DockedFooter) adds group options and changes component to use ref

acacia-m-pappas:
## Summary: 
This PR addresses requested changes to the **DockedFooter**, namely: 

-  to use a `ref` instead of the `data-test-section`
-  adds content group options for `left` and `center`, or any/all combinations of `right`, `left`, and `center` content grouping. 

Other improvements included in this PR are additional `mousemove`, `wheel` and `keydown` event listeners. These account previously unaccounted for use cases when a user might be using a mouse instead of a scroll pad, might be using vim, and in the case of `mousemove` causes the footer to register whether it should be docked more immediately after the parent component mounts. 

Learn how to implement and pass the `ref` in[ this comment ](https://github.com/optimizely/oui/blob/338ddf5d84d9718efca2ac8b81b8d9182343d532/src/components/DockedFooter/DockedFooter.story.js#L11)in the **DockedFooter Story**.

## Test Plan:

- Uses built in tests

## JIRA Issues: 
[FEI-3438](https://optimizely.atlassian.net/browse/FEI-3438)


src/components/DockedFooter/DockedFooter.story.js
```diff
@@ -1,83 +1,197 @@
...
 import { storiesOf } from '@storybook/react';
 import noop from 'lodash.noop';
 import { withInfo } from '@storybook/addon-info';
+import { action } from '@storybook/addon-actions';
 
 import DockedFooter from './index.js';
 import Button from '../Button';
+import PaginationControls from '../PaginationControls/index.js';
+import Code from '../Code/index.js';
+
```
benjaminpadula: Love the comments here, thank you!

PR #1279, opened 2020-01-23
Fixing issue with search input

kwalker3690:


src/components/Dropdown/index.js
```diff
@@ -47,7 +48,7 @@ class Dropdown extends React.Component {
   };
 
   handleOnBlur = () => {
-    if (!this.props.overChildren) {
+    if (!this.props.overChildren & !this.props.overActivator) {
```
jamesopti: Can you see if this solves the problem by itself?

```js
handleOnBlur = (e) => {
  const { target } = event;
  const targetIsInput = <check target to be an input>;
  if (!this.props.overChildren && !targetIsInput) {
    this.props.hide();
  }
}
```

If that works, we could be more specific by adding an attribute to our SearchPicker `<Input>` to instruct dropdown to prevent hiding on blur.

jamesopti: Wed need to ensure too that the blur is on an element INSIDE this instance of dropdown too, but that wouldnt be hard.

kwalker3690: This works, but only in conjunction with the modifications in SearchPicker itself. Maybe tracking the focused state should be moved to Dropdown itself?

jamesopti: Gonna patch this over now.

PR #1280, opened 2020-01-24
fix (ForwardRef Components): Convert components that use forwardRef to Typescript

jamesopti:
## Summary

The Typescript CLI isnt able to properly generate types for components written in plain ol JS that use React.forwardRef.

See the current type for [Input](https://github.com/optimizely/oui/blob/c3fa42dc4ca8be2971d446a8ac8336601220d905/types/index.d.ts#L265-L275) or the [3 Layout components](https://github.com/optimizely/oui/blob/c3fa42dc4ca8be2971d446a8ac8336601220d905/types/index.d.ts#L714-L728).

This PR converts those few usages to be typed so that the typedefs are correctly generated.

### What about PropTypes?

Fear not! The babel plugin `babel-plugin-typescript-to-proptypes` automatically generates prop types for Typescript components to ensure backward compatibility with non TS consumers.


src/components/Layout/Col.tsx
```diff
@@ -0,0 +1,111 @@
...
+  /** Whether or not this is a reading column layout */
+  isReadingColumn?: boolean;
+  /**
+   * The number of columns to span on large devices (≥992px)
+   *
+   * @type {("fillSpace"|"fitContent"|number|
+   *  { span: "fillSpace"|"fitContent"|number, offset: number, order: number }
+   * )}
+   */
+  large: column;
```
jamesopti: @ayeshakmaz are both `small` and `large` required props?

Per the [current code's proptypes](https://github.com/optimizely/oui/blob/devel/src/components/Layout/Col.js#L80), they are, but should they be? I know we discussed that it works without them.

ayeshakmaz: Yeah neither are required

jamesopti: Awesome, updated these to be optional!

PR #1281, opened 2020-01-24
fix (SearchPicker) Allow for an offset in search picker children

kwalker3690:
In the current SearchPicker, it is not possible to accurately use a keyboard to navigate the child elements if there is any element that can be selected that is not one of the entities returned from the search function. The common example is denoted in the story, adding a Create option to the list of elements. This is solved by passing an offset as a prop to SearchPicker, in conjunction with correctly specifying the faux focus in the SearchPicker children implementation as being offset by the same amount.


src/components/SearchPicker/index.js
```diff
@@ -40,6 +40,11 @@ class SearchPicker extends React.Component {
      * via @keyboardTracker
      */
     handleKeyDown: PropTypes.object.isRequired,
+    /**
+     * Number of extra items shown in the children, used to keep
+     * keyboard selection in the correct place
+     */
+    itemOffset: PropTypes.number.isRequired,
```
jamesopti: So I think this makes sense but that the wrapper `keyboard` tracker is the one that should receive this prop and handle it, as it only pertains to that component. Can you move it there and add a test in `accessibility.js` tests?

Also, I think `additionalItems` is a more clear name.

jamesopti: Or maybe `additionalItemCount`? Naming is hard

src/components/SearchPicker/index.js
```diff
@@ -12,6 +12,12 @@ const INPUT_DEBOUNCE_PERIOD = 120;
 @keyboardTracker
 class SearchPicker extends React.Component {
   static propTypes = {
+    /**
+     * Number of extra items shown in the children, used to keep
+     * keyboard selection in the correct place.
+     * used by @keyboardTracker
+     */
+    additionalItems: PropTypes.number,
```
jamesopti: The searchpicker doesnt need to know about this prop (see below).

src/utils/accessibility.js
```diff
@@ -33,8 +34,9 @@ export const getAssistiveTextFromColorClass = className => {
  */
 export const keyboardTracker = Component => {
   const wrappedComponent = props => {
+    const { additionalItems } = props;
     const [currentFauxFocusIndex, setIndex] = useState(0);
-    const [currentItemCount, setItemCount] = useState(0);
+    const [currentItemCount, setItemCount] = useState(additionalItems);
```
jamesopti: So `case 'ArrowDown':` is the only place where we care about the `currentItemCount`, we dont need to set it in state.

At the top of that case, we can just say `const totalItemCount = currentItemCount + additionalItems`, then use that in the `newIndex` computation.

Then `additionalItems` should be added to the dependency array for this callback.

src/utils/accessibility.js
```diff
@@ -90,5 +93,11 @@ export const keyboardTracker = Component => {
     );
   };
   wrappedComponent.displayName = `withkeyboardTracker(${Component.displayName})`;
+  wrappedComponent.propTypes = {
```
jamesopti: Instead of passing down `additionalItems`, we can just consume it above, then pass everything else to the child:

```js
const { additionalItems, ...rest } = props;
<Component { ...rest }
  ...
```

src/utils/accessibility.js
```diff
@@ -73,8 +75,9 @@ export const keyboardTracker = Component => {
      * current index to 0 if the number has changed.
      */
     const handleSetItemCount = useCallback((count) => {
-      setItemCount(count);
-      if (count !== currentItemCount) {
+      const totalCount = count + additionalItems;
```
jamesopti: Then this can stay unchanged, as `currentItemCount` represents the number of items provided by the child.

PR #1283, opened 2020-01-24
chore (Docs): Add docs for Typescript

jamesopti:
## Summary

Adds some more documentation for using Typescript


CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ This file is similar to the format suggested by [Keep a CHANGELOG](https://githu
 
 ## Unreleased
 
+- [Feature] Enable Typescript compilation and generate type definitions for JS and TSX components. ([#1268](https://github.com/optimizely/oui/pull/1268))
```
circAssimilate: This is for the earlier merge?

CONTRIBUTING.md
```diff
@@ -27,6 +27,34 @@ Every component should contain prop definitions and a robust set of Storybook ex
...
+
+Chances are whatever you're trying to type, someone else has typed it before! There are a myriad of examples and support questions on the internet due to the widespread adoption of Typescript. Google around and you'll likely find excellent help.
+
+When it comes to React components, typing can get a little tricky, but fortunately, React itself has excellent Typescript support. Check out the following resources for extensive examples of React Typing:
+
+[React Basic] - https://github.com/typescript-cheatsheets/react-typescript-cheatsheet#basic-cheatsheet-table-of-contents
+[React Advanced] - https://github.com/typescript-cheatsheets/react-typescript-cheatsheet/blob/master/ADVANCED.md
+[React HOC] - https://github.com/typescript-cheatsheets/react-typescript-cheatsheet/blob/master/HOC.md
+
+**Additional Resources:**
```
circAssimilate: nit: What about `#####` for linkability?

CONTRIBUTING.md
```diff
@@ -27,6 +27,34 @@ Every component should contain prop definitions and a robust set of Storybook ex
...
+
+[Typescript in 5min] - https://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes.html
+[Typescript Handbook] - https://www.typescriptlang.org/docs/handbook/basic-types.html
+[Justification - Why Typescript?] - https://medium.com/@jtomaszewski/why-typescript-is-the-best-way-to-write-front-end-in-2019-feb855f9b164
+
+#### @ts-ignore and `:any`
+
+Don't let Typescript errors slow down your development if you dont want them to. Use `// @ts-ignore` to tell the Typescript compiler to ignore the next line, or specify a type as `:any` if you can't figure out why the compiler is complaining to you. Typescript is fun once you get a hang of it, but sometimes you just wanna see if your code works and come back to typing it later.
```
circAssimilate: Think it's awesome you put this. this can be the most frustrating part of Typescript. What if you also suggested `@ts-ignor`ing and not blocking on it but asking in #frontend to see if someone can help so it creates a learning moment?

README.md
```diff
@@ -37,6 +37,11 @@ The `optimizely-oui` declaration file (`types/templates/module-declaration.d.ts`
 To build the exported declaration file (`types/index.d.ts`), the autogenerated individual component module declarations
 are merged with the main declaration file (see `yarn generate-types`).
 
+### PropTypes
+
+`PropTypes` can still be used for non-Typescript (`.js`) components as well as when more complex validation is needed.
```
circAssimilate: To be clear though, `PropTypes` becomes redundant (outside of edge cases) with Typescript, correct? That's been my experience, and I think we should explicitly call that out first if so.

jamesopti: > To be clear though, `PropTypes` becomes redundant (outside of edge cases) with Typescript, correct? That's been my experience, and I think we should explicitly call that out first if so.

Thats true except for 2 scenarios:
- The repo youre using OUI in _is not typed_ (like the monolith). In this case, PropTypes are your only safety.
- The validation you want to do can only be done at _runtime_ (requires a custom function)

I just tested this in the monolith and while the typescript definitions help with suggestions, they dont enforce incorrect values.

I'd like to enable Typescript in the monolith but not sure how we'd configure it to not show errors everywhere initially. Wanna help me with that?

Screen Shot 2020-01-24 at 3 09 49 PM

CONTRIBUTING.md
jamesopti: I am suggesting `@ts-ignore` here. What else are you suggesting?

Great idea to include reaching out in #frontend!

PR #1284, opened 2020-01-27
fix (SearchPicker) Wait for all queries to complete before showing result summary

kwalker3690:
This resolves the issue as seen in this GIF:
search-picker-4
The summary was being incorrectly populated because of how we were handling the promise then/catch scenarios

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ This file is similar to the format suggested by [Keep a CHANGELOG](https://githu
 
 ## Unreleased
 - [Patch] Update `isReadingColumn` prop on **Col** component to create a nested reading column ([#1285](https://github.com/optimizely/oui/pull/1285))
+- [Patch] Wait for all queries to be complete before showing result summary in **SearchPicker** ([#1284](https://github.com/optimizely/oui/pull/1284))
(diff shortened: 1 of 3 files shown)


types/index.d.ts
```diff
@@ -5194,8 +5193,8 @@ declare module "components/PaginationControls/index" {
             remainingSlotsToFill: number;
             currentPage: number;
         };
-        fillPageSlots(): any[];
-        renderPageNumbers(): JSX.Element[];
+        fillPageSlots(): any;
+        renderPageNumbers(): any;
```
circAssimilate: nit since I know the direction is we can allow any. Was this throwing a compile error? I wonder if `React.ReactElement` would work here.

types/index.d.ts
```diff
@@ -5438,8 +5437,8 @@ declare module "components/SearchPicker/index" {
         static defaultProps: {
             searchOptions: {};
             selectedEntityIds: never[];
-            setItemCount: (...args: any[]) => void;
-            setOnItemSelect: (...args: any[]) => void;
+            setItemCount: any;
```
circAssimilate: Seems like this could just be `Function`, right? Just wanna get these while we can if they're easy (and perhaps they're not!)

PR #1285, opened 2020-01-28
fix(layout): Move reading column to child of Col

ayeshakmaz:
After testing the DockedFooter in the monolith, it became clear that the scrollable container and the reading-column cannot be the same div, therefore this PR separates the concerns


ayeshakmaz:
@davemn I now read minds 😛 Glad this helps!

PR #1293, opened 2020-02-14
feat(SelectDropdown): Add ability to add label to select dropdown

ayeshakmaz:
Needed for [Frontend PR #11841](https://github.com/optimizely/optimizely/pull/11841)

JIRA:
[UXENG-56](https://optimizely.atlassian.net/browse/UXENG-56)


src/components/Input/index.scss
```diff
@@ -23,14 +23,17 @@
...
-  height: spacer(3.5);
 }
 
 .#{$namespace}text-input[type="number"],
 .oui-text-input[type="number"] {
   text-align: right;
 }
 
+.oui-text-input[type="time"] {
+  max-height: map-fetch($button, size, base, height);
```
jessicalc: How come the time input has a max-height explicitly set, but the other inputs don't?

src/components/SelectDropdown/SelectDropdown.story.js
```diff
@@ -146,7 +146,7 @@ stories.add('Default', (() => {
       />
     </Container>
   );
-})).add('ActivatorLabel', (() => {
+})).add('Activator Label differs from Value', (() => {
```
jessicalc: Thanks for making this clearer!

src/components/SelectDropdown/index.js
```diff
@@ -8,6 +8,12 @@ import Link from '../Link';
 
 class SelectDropdown extends React.Component {
   static propTypes = {
+    /**
+     * Button text object with label and current value
```
jessicalc: I think you mean "content" value but maybe not? 
```suggestion
     * Button text object with label and content value
```

src/components/Input/index.scss
ayeshakmaz: The native, out-of-the-box time input is slightly larger due to the arrow buttons, so this forces it to be equal to our other inputs (it's a very annoying HTML thing :P )

PR #1294, opened 2020-02-14
fix(Code): Don't wrap text inside Code component (block mode)

davemn:
## Summary

Updates the `<Code>` component to only wrap child text if the text contains newlines. Previously it was using `pre-wrap` behavior:

> Sequences of white space are preserved. Lines are broken at newline characters, at \<br\>, and as necessary to fill line boxes.

Also includes a couple of small fixes to the related story to allow pasting JSX and a knob for syntax highlighting.

**Background**

Based on what I could find in the repo history, the old behavior was decided in #20 with the note:

> Changing to pre-wrap instead of pre since pre-wrap is the default used by browsers for pre tags.

That's not true anymore; this PR will bring it back in line with current (2020) browser behavior for `<pre>` tags.

## Jira

This came up while I was working on https://optimizely.atlassian.net/browse/MGMT-2905.


src/components/Code/Code.story.js
```diff
@@ -29,13 +29,21 @@ stories
...
       {story()}
     </div>
-  ));
+  ))
+  .addParameters({
+    knobs: {
+      // Allows users to paste JSX into the text knobs (e.g. `<MyComponent>`)
+      // w/o it getting rewritten as `&lt;MyComponent&gt`;
+      escapeHTML: false,
+    },
```
davemn: From https://github.com/storybookjs/storybook/tree/master/addons/knobs:

> It's safe to set [escapeHTML] to `false` with frameworks like React which do escaping on their side.

PR #1295, opened 2020-02-18
fix(SelectDropdown): If `SelectDropdown` is `fullWidth`, set `Dropdown` width to 100%

jessicalc:
- Update `<Dropdown>` component such that if `fullWidth` is provided, the `<Dropdown>` will take on a width of 100%
- Update SelectDropdown story with examples

In updating the Pages UI, I realized that using `<SelectDropdown fullWidth={ true }/>` had no effect on the children:

image

This is an attempt to make `fullWidth` apply to the dropdown contents as well. 

### PREVIOUSLY
* image

### NOW
* If `fullWidth` is passed to the SelectDropdown, the Dropdown items within the dropdown will also be fullWidth. 
* image

* It can be overridden by setting the `minDropdownWidth` prop. 
* image

Open questions:
* In what scenario would you want the activator to be 100% (for example, 600px wide due to the container it's in), but have the dropdown contents be less than that (for example, 300px wide?)


src/components/SelectDropdown/SelectDropdown.story.js
```diff
@@ -180,6 +194,17 @@ stories.add('Default', (() => {
...
+  return (
+    <Container>
+      <SelectDropdown
+        items={ items }
+        fullWidth={ true }
+        initialPlaceholder="Select a value..."
+        onChange={ action('SelectDropdown value changed') }
+      />
+    </Container>
+  );
```
ayeshakmaz: I might be missing it, but is this different from the Full Width Activator with Full Width items story?

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This file is similar to the format suggested by [Keep a CHANGELOG](https://github.com/olivierlacan/keep-a-changelog).
 
 ## Unreleased
+- [Patch] If `SelectDropdown` is `fullWidth`, set `Dropdown` width to 100% [#1295](https://github.com/optimizely/oui/pull/1295)
```
ayeshakmaz: Small nit to add parentheses around the PR link ex:` ([#1295](https://github.com/optimizely/oui/pull/1295))`

src/components/SelectDropdown/SelectDropdown.story.js
jessicalc: Oops, it's the same! :homer_disappearing: I fix. Thanks

PR #1296, opened 2020-02-18
feat(Button): Add "toolbar" style

circAssimilate:
**Summary:**
Working on Experiment Manager Change History and need to port a “toolbar” [button style from the Monolith](https://github.com/optimizely/optimizely/blob/devel/src/www/frontend/src/scss/v2/partials/objects/_toolbar.scss) to OUI. This is kind of like a combination of _toggle_ and _plain_ buttons. Copying over styles from the [ReactToolbar](https://github.com/optimizely/optimizely/blob/devel/src/www/frontend/src/js/react_components/toolbar/index.js) used in Monolith components like the Personalization Campaign Manager.

**Screenshots**
_Personalization Campaign Manager in the Monolith_
image

_With this PR's "toolbar" style_ 
image

**Test Plan**
- [x] Run `yarn storybook` and view http://localhost:9009/?path=/story/button--all-buttons
- [ ] Ensure tests pass


src/components/Button/index.scss
```diff
@@ -227,6 +227,32 @@
...
+
+    &[disabled],
+    &--disabled,
+    &--disabled:hover {
+      @extend %link--disabled;
+      box-shadow: none;
+      border-color: transparent;
+      background: none;
+    }
+  }
```
circAssimilate: This was taken from https://github.com/optimizely/optimizely/blob/devel/src/www/frontend/src/scss/v2/partials/objects/_toolbar.scss

PR #1300, opened 2020-02-21
patch(Button): Update focus and hover states for buttons

ayeshakmaz:
Add border to Plain style on hover, adjust focus and hover states for Underline style


src/components/Button/index.scss
```diff
@@ -183,13 +183,19 @@
...
 
-    &:hover {
+    &:hover,
+    &:focus {
       color: map-fetch($button, type, plain, hover, color);
+      border: 1px solid var(--light-blue-25);
     }
 
     &:active {
       color: map-fetch($button, type, plain, active, color);
```
circAssimilate: Maybe this is what needs updated!?

ayeshakmaz: great catch, thank you!

PR #1304, opened 2020-03-03
Convert Button component to typescript

pauloptimizely:
## Summary

This is a first pass at adding strict type checking to the Button component.

## Test Plan

1. Run the existing tests `yarn test`
1. Performed a sanity check in storybook
1. Used `yarn link` to verify the changes i've made will not impact our internal codebases


src/components/Button/index.tsx
```diff
@@ -1,34 +1,75 @@
...
+  /** Make the button act as a submit button */
+  isSubmit?: boolean,
+  /** When the button adds a spinner, it displays this text */
+  loadingText?: string,
+  /** Function that fires when the button loses focus */
+  onBlur?: (e: React.FocusEvent<HTMLButtonElement|HTMLDivElement>) => void,
+  /** Function that fires when the button is clicked on */
+  onClick?: (e: React.MouseEvent) => void,
+  /** Function that fires when the button is mouse downed */
+  onMouseDown?: (e: React.MouseEvent) => void,
```
pauloptimizely: I wanted to make the function handlers has strict as possible for this PR. I would like feedback on whether or not we should be so strict about return types

src/components/Button/index.tsx
```diff
@@ -1,34 +1,75 @@
...
+  /** Various height and width options */
+  width?: 'default' | 'full',
+};
+
 /**
  * @param {Object} props - Properties passed to component
  * @returns {ReactElement}
  */
 
 const Button = ({
```
pauloptimizely: Notice here, i didn't use `React.FC` or `React.SFC`, that's because default-props don't (currently) work with FC or SFC

src/components/Button/index.tsx
```diff
@@ -37,29 +78,31 @@ const Button = ({
...
-  function handleOnClick(event) {
+  function handleOnClick(event: React.MouseEvent): void {
     if (isDisabled || isLoading) {
       return;
     }
     onClick(event);
   }
 
   if (isLink) {
+    const ref: DivRef = buttonRef;
```
pauloptimizely: I wasn't able to make ref type that mixed different element types, so I created an "AnyRef" type, then casted the "AnyRef" to the current ref type at render time

src/components/Button/index.tsx
pauloptimizely: I might give this another shot based on this thread https://github.com/microsoft/TypeScript/issues/31247

pauloptimizely:
I'll revisit this later

PR #1318, opened 2020-04-21
Removing unused components

pauloptimizely:
## Summary

There are a couple of components in OUI that aren't used in any of our code bases. I created both of these components a while back and ended up not using them.

Removing:
- `AutoComplete`
- `EditableInput`

## Test Plan

Verify `AutoComplete` and `EditableInput` do not exist in any of our frontend code bases


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This file is similar to the format suggested by [Keep a CHANGELOG](https://github.com/olivierlacan/keep-a-changelog).
 
 ## Unreleased
+- [Feature] Remove `AutoComplete` component
```
ayeshakmaz, line 9: Small nit to change both of these to **bold** for the component name

PR #1320, opened 2020-04-22
feat(Input): Add ability for icons on left and right and clear button

ayeshakmaz:
Summary:

Add `leftIconName` and `rightIconName` to allow for icons on either side of an input

Add `hasClearButton` and `onClearButtonClick` to add a clear button to the right of the input

Screen Shot 2020-04-28 at 5 02 41 PM

Jira:

[UXENG-22](https://optimizely.atlassian.net/browse/UXENG-22)


src/components/Input/tests/index.js
```diff
@@ -196,14 +234,56 @@ describe('components/Input', () => {
...
+    expect(component.find('[data-test-section="foo-note"]').text()).toBe(
+      'A short description'
+    );
   });
 
   it('should add a maxLength attribute when maxLength is passed', () => {
     const component = mount(<Input type="text" maxLength={ 50 } />);
     expect(component.find('input').prop('maxLength')).toBe(50);
   });
+
```
ayeshakmaz, line 251: This is the new stuff, everything above is formatting fixes

src/components/Input/index.tsx
```diff
@@ -87,112 +101,172 @@ type InputProps = {
...
-    return (
-      <div data-oui-component={true} className={classNames({ 'oui-form-bad-news': displayError })}>
-        <Label
-          testSection={testSection && testSection + '-label'}
-          isRequired={isRequired}
-          isOptional={isOptional}
-          inputId={id}
+
+    const renderIcon = (iconName) => {
+      return <Icon name={iconName} size="medium" />;
```
acacia-m-pappas, line 166: If I was implementing this, I would most likely want/need this icon to also have a click handler...

image
It would depend on what icon was inserted here but in the case of this right pointing tick, if I clicked it I would expect it to be like hitting ENTER on my keyboard.

src/components/Input/index.tsx
```diff
@@ -87,112 +101,172 @@ type InputProps = {
...
+            {(rightIconName && !hasClearButton) && (
+              <span className="flex oui-input-icon__right">
+                {renderIcon(rightIconName)}
+              </span>
+            )}
+            {hasClearButton && (
+              <span className="flex oui-text-input__clear-button">
+                {renderClearButton()}
+              </span>
+            )}
```
acacia-m-pappas, line 236: Just checking that if `hasClearButton` is set to `true`, it _should_ override a rightIcon. Is that right?

ayeshakmaz: yup! That's right :)

src/components/Input/index.tsx
ayeshakmaz: Ahh this is great feedback, probably not the best icon choice. I'll update to be something that wouldn't look clickable

For now, I think we'll keep it to only allowing a clear button and static icons. Icons can't have click handlers (not a button so not accessible) so opening it up to have anything be a button might be a little too wild west 😛 

tl;dr - I'll make a better example 😄

PR #1323, opened 2020-04-28
feat (SearchPicker) Expose search function to the render prop API

pauloptimizely:
## Summary

**What?**

This PR exposes a search function to the render prop API. 

**Why?**

Attempting to keep search results up to date, as users click through the application. This functionality will allow us to repopulate "suggested" search results when a user re-focuses a search input.


src/components/SearchPicker/index.js
```diff
@@ -300,14 +300,15 @@ class SearchPicker extends React.Component {
...
-    const { renderInput } = this;
+    const { renderInput, handleOnInputDebounced } = this;
     const availableEntities = this.getAvailableEntities();
     return (
       children({
         availableEntities,
         isLoading,
         renderInput,
         searchQuery,
+        search: handleOnInputDebounced,
```
ayeshakmaz, line 311: could we rename this to be a little more clear? I think `search` is such an overloaded term in this context that it's hard to tell without looking at the code what I might use this for

pauloptimizely: We could call it something like: `executeNewSearch`, `executeSearch` or `makeSearchRequest`?

ayeshakmaz: OO I like makeSearchRequest! That definitely helps with clarity

pauloptimizely: Awesome!! makeSearchRequest works for me 👍

PR #1325, opened 2020-04-30
fix(SearchPicker): Ensure defaultResults is at least an empty Array

circAssimilate:
**Summary:**
It appears that code paths like https://github.com/optimizely/optimizely/blob/8450148bd91e9590a2236f828dbb1740ed0b51ab/src/www/frontend/src/js/bundles/p13n/components/audience_search_picker/index.js#L45-L60 aren't actually possible because if the Search API does error, `defaultResults` will be undefined and throw a big 'ol error. This will fix this. Came across this while adding a few tests in https://github.com/optimizely/optimizely/pull/12378/files#diff-9653e815c59649186f731f82db2f214aR189-R222.

**Test Plan:**
Manually verified this work in the Chrome Debugger


CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ This file is similar to the format suggested by [Keep a CHANGELOG](https://githu
 
 ## Unreleased
 
+- [Patch] Set Search defaultResults to an empty array at least to prevent error ([1325](https://github.com/optimizely/oui/pull/1325))
```
ayeshakmaz, line 10: Just a small nit to change `Search` to ` **SearchPicker** ` and maybe 'to at least an empty array when results are undefined to prevent error'

src/components/SearchPicker/index.js
```diff
@@ -112,7 +112,7 @@ class SearchPicker extends React.Component {
     }).then(results => {
       const { searchQuery } = this.state;
       this.setState({
-        defaultResults: results,
+        defaultResults: results || [],
```
jamesopti, line 115: So I would say that the consumer is incorrectly resolving the promise in the function they pass to the search picker.

`resolve()` is an invalid state.

Can you dig into why the consumer is doing that?

I would support handling the rejection here, but I think defaulting to an empty list just covers up the consumers mistake.

jamesopti: Ah, this is when `response.ok` is false.

I think this line should throw then and be caught/logged: https://github.com/optimizely/optimizely/blob/8450148bd91e9590a2236f828dbb1740ed0b51ab/src/www/frontend/src/js/bundles/p13n/components/audience_search_picker/index.js#L49

circAssimilate: Should this be making assumptions that consumers are resolving with arrays? This was throwing both when response.ok was false and when it resolve with nothing. It seems like if we don't fix it here we will run into more implementation issues because someone will forget to catch this search API error and return empty results.

circAssimilate: At the very list, it seems like this component should not be able to blow up the entire page because of an unexpected response.

circAssimilate: @jamesopti, to reproduce the behavior I'm recommending we fix, you can go to the Full Stack Experiment Manager Audiences page, then set a breakpoint at this line and set results to `undefined` on component mount. This will trigger the error boundary for that component - which I think it a little risky with the frequency that search requests are made and the higher possibility of intermittent failure and removal of entire Search Picker.

image

image

jamesopti: I think I'm ok with doing this, but I want to log loud and clear that the consumer is resolving with an invalid value.

This component can be defensive to protect itself, but it should yell out loudly that its getting an unexpected value and that the consumer is responsible for handling their own errors, as only the consumer knows what the shape of their own raw response is.

I think the defensiveness should be to bail/no-op. It protects us from getting in an invalid state, but makes a fuss in the console about it.

```js
}).then(results => {
  if (!searchQuery) {
    this.setState({
      isLoading: false,
    });
  }
  this.verifyResultOrThrow(result);
  ...
})
...

verifyResultOrThrow(result) {
  if (!Array.isArray(results)) {
    throw new Error('SearchPicker props.searchFunction resolved with a value that is not an Array.');
  }
}
```

jamesopti: We should do this in both places that execute `searchFunction` as well.

Then we should then also fix the monolith to throw if the response is not an array and use the `.catch` that already exists to notify the user.

circAssimilate: > as only the consumer knows what the shape of their own raw response is.

@jamesopti, the actual error happens because `.filter` is run on the `this.state.results`. So it's not exactly up to the consumer. If I'm understanding correctly your suggestion just changes an implicit error to an explicit one. If that's the case and we still need to catch and always turn an array where this is implemented, then I'm not too sold on the value of this PR. Is that what you're recommending?

circAssimilate: >I think the defensiveness should be to bail/no-op. It protects us from getting in an invalid state, but makes a fuss in the console about it.

Your code example wouldn't do a noop though would it, it would make the component throw an error and need to be caught like my example above. Right?

jamesopti: > Your code example wouldn't do a noop though would it, it would make the component throw an error and need to be caught like my example above. Right?

Throwing in a promise `.then` moves the Promise chain to the reject path. Since it is an asynchronous operation, it wont break the OUI component.

As the monolith consuming code stands today, the `.catch` will catch the rejection caused by this `throw` and the error dialog will be shown (as well as the Promise error in the console).

This will make it obvious to the consuming developer that they need to handle the case where `.ok` is true, but the value is `undefined`

circAssimilate: Thanks for the discussion. I need to focus on the initial Monolith PR first but will circle back around to this when I get a chance - hopefully around the end of the week or next.

jamesopti: Sounds good! I think it can actually be cleaner by doing this:

```js
}).then(results => {
  this.verifyResultOrThrow(result);
  ...
  // setState after validating results
})
.finally(() => {
  // Always set loading false as long as we don't have another searchQuery
  if (!searchQuery) {
    this.setState({
      isLoading: false,
    });
  }
})
...

verifyResultOrThrow(result) {
  if (!Array.isArray(results)) {
    throw new Error('SearchPicker props.searchFunction resolved with a value that is not an Array.');
  }
}
```

CHANGELOG.md
```diff
@@ -7,132 +7,163 @@ This file is similar to the format suggested by [Keep a CHANGELOG](https://githu
 
 ## Unreleased
 
+- [Patch] Set SearchPicker defaultResults to an empty array at least to prevent error ([1325](https://github.com/optimizely/oui/pull/1325))
+
 ## 46.1.0 - 2020-05-08
+
```
circAssimilate, line 13: This and many other of the following changes are from Prettier.

src/components/SearchPicker/index.js
```diff
@@ -1,10 +1,10 @@
-import { debounce, noop } from 'lodash';
-import PropTypes from 'prop-types';
-import React from 'react';
+import { debounce, noop } from "lodash";
```
circAssimilate, line 1: This an many other of the following changes are from Prettier.

circAssimilate:
@jamesopti, got my other PR working with your Array return suggestion but just incorporated your suggestion if you can take an other look. Thanks!

circAssimilate:
@ayeshakmaz, I looked at the UI Review & Tests and everything looked good to me. There were some changes, but not related to this PR so I accepted.

Also, added some tests, @jamesopti!

PR #1328, opened 2020-05-07
fix(Token): Update spacing between ordering and drag handle

ayeshakmaz:
**Summary:**

Make spacing b/w order number, drag handle, and text consistent

**Before:**
OUITokenOrderDragSpacing

**After:**
Screen Shot 2020-05-07 at 7 46 40 PM

**Jira:**
[UXENG-87](https://optimizely.atlassian.net/browse/UXENG-87)


jessicalc:
What happens if there are more than 2 digits? 

It seems like if we go to 3 digits and the digits are of varying width, the draggers get offset variably. Are we OK with that? If so, maybe we should document it and also add an example in storybook. 

image

ayeshakmaz:
@jessicalc  good call, added a notes in the comments of the CSS file, and will add a note in the Axiom documentation as well (once it's written)

benjaminpadula:
@ayeshakmaz Is the plan still to also add this ticket into this PR? https://optimizely.atlassian.net/browse/MGMT-3178

ayeshakmaz:
@benjaminpadula I'm actually going to keep them separate, 3178 is actually more involved than it sounds, so I want to be careful

PR #1329, opened 2020-05-08
patch(storybook): remove mock date to fix storybook knobs

ayeshakmaz:
**Summary:**

Turns out using mock date to change the system time (for VRT tests) [breaks storybook knobs](https://github.com/storybookjs/storybook/issues/9671) 

So I've adjusted to just use an initial starting date for date pickers instead of mocking the system time

And now the knobs work! And we get to keep chromatic VRT! Yay!!


circAssimilate:
Thanks, @ayeshakmaz!

PR #1331, opened 2020-05-13
feat(DetailSideNav): Initial cmp for OUI side nav

ayeshakmaz:
**Summary:**

Initial component for DetailSideNav (bringing the `Sidebar` component at `src/js/react_components/sidebar/index.js` from the monolith to OUI )

Side features:
- Add ability to use `isFullWidth` prop on **Link** to create a full width link 
- Add `ariaLabel` prop on **Switch** to provide screen readers with context
 
Screen Shot 2020-05-13 at 6 01 51 PM

**Jira:**
[UXENG-75](https://optimizely.atlassian.net/browse/UXENG-75)


src/components/Switch/index.scss
```diff
@@ -95,7 +95,7 @@
 .oui-switch:focus + label, .oui-switch:active + label {
   box-shadow:
     1px 1px 10px rgba(0, 0, 0, 0.2) inset,
-    0 0 0 1px var(--light-blue-25) !important;
+    0 0 0 2px var(--light-blue-25) !important;
```
ayeshakmaz, line 98: Improves focus state of switch

src/components/DetailSideNav/DetailSideNav.story.js
```diff
@@ -0,0 +1,152 @@
...
+import Link from '../Link/index.js';
+
+const stories = storiesOf('Navigation/DetailSideNav', module);
+stories
+  .addDecorator(withKnobs)
+  .addDecorator((story) => <div id="root-preview">{story()}</div>);
+
+stories.add('Default', () => {
+  return (
+    <DetailSideNav>
```
circAssimilate, line 18: I see that `SideNav` is already used (also curious if we actually use that anywhere or could replace), but curious why "Detail" in "DetailSideNav" instead of something like `ManagerSideNav` This could eventually tie into an OUI `Manager` where it does the `Container`, `Col`, `Row`, scaffolding for us (maybe with an option for `ColWithNubbin`?). Getting a head of ourselves, but "Detail" made me think of something that happened more towards the right when I clicked on an item within a single view.

src/components/DetailSideNav/DetailSideNav.story.js
```diff
@@ -0,0 +1,152 @@
...
+      <DetailSideNav.Header
+        hasBackLink={ true }
+        backLinkHref="#features"
+        backLinkText="Features"
+        backLinkOnClick={ action('Back link was clicked') }
+        isFullStackProject={ true }
```
circAssimilate, line 24: Looks like this just determines the class put on the h1 if I'm reading it right. What if someday Web also wants to have keys exposed for their experiments? I wonder if we could call it something like `shouldStylePrimaryTitleAsKey` (or something better) and have more flexibility in the future and keep OUI ignorant of Full Stack vs. Web. Wdyt?

src/components/DetailSideNav/DetailSideNav.story.js
```diff
@@ -0,0 +1,152 @@
...
+            </Link>
+          }
+          rightContent={
+            <Switch
+              ariaLabel="Turn feature on in Production"
+              elementId={ 'prod-toggle' }
+              checked={ true }
+              onClick={ action('Toggled Prod Switch') }
+            />
+          }
```
circAssimilate, line 59: Curious if we think that other leftContent / rightContent use cases will come about or if these props only exist to obfuscate the flex classes on leftContent (which I see the benefit of). Also, it feels kind of strange to have leftContent be the only prop used at times for the items below. Should we make this `NavItem` component accept children in lieu of `leftContent` and change the name of `rightContent` to `additionalContent` (or even keeping it the same prop name is fine)?

src/components/DetailSideNav/HeaderDetails/index.js
```diff
@@ -0,0 +1,59 @@
...
+          'flush--bottom': true,
+          'push--top': ind > 0,
+        }) }>
+        <h5>{item.label}</h5>
+      </div>,
+      <span key={ `value-${item.label}` } className="sidenav__details__item">
+        {item.value}
+      </span>,
+    ];
+  });
```
circAssimilate, line 26: nit: I prefer these inline with the component if only used once - just helps for quicker reading me thinks.

src/components/DetailSideNav/NavItem/index.js
```diff
@@ -0,0 +1,41 @@
...
+      className={ classNames({
+        'nav-list__link': true,
+        'is-active': props.isActive,
+      }) }
+      data-test-section={ props.testSection }>
+      <div className="flex flex-justified--between flex-align--center flex--1 soft--ends soft-double--sides">
+        <div className=" flex flex--1 flex-align--center">
+          {props.leftContent}
+        </div>
+        {props.rightContent}
```
circAssimilate, line 18: See comments on story for some thoughts here!

src/components/DetailSideNav/Header/index.js
```diff
@@ -0,0 +1,107 @@
...
+import classNames from 'classnames';
+
+const Header = (props) => {
+  const {
+    backLinkHref,
+    backLinkOnClick,
+    backLinkText,
+    friendlyTitle,
+    isFullHeight,
+    isFullStackProject,
```
circAssimilate, line 12: See comment on story for some thoughts about this prop!

src/components/DetailSideNav/Header/index.js
```diff
@@ -0,0 +1,107 @@
...
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+const Header = (props) => {
+  const {
+    backLinkHref,
+    backLinkOnClick,
+    backLinkText,
+    friendlyTitle,
+    isFullHeight,
```
circAssimilate, line 11: When wouldn't we want this?

src/components/DetailSideNav/Header/index.js
```diff
@@ -0,0 +1,107 @@
...
+    isFullStackProject,
+    projectName,
+    shouldHeaderScroll,
+    hasBackLink,
+    spaceBelow,
```
circAssimilate, line 16: When would we want space below?

src/components/DetailSideNav/Header/index.js
```diff
@@ -0,0 +1,107 @@
...
+    title,
```
circAssimilate, line 17: What if we named `title` primaryTitle` and `friendlyTitle` `secondaryTitle`?

src/components/DetailSideNav/Header/index.js
ayeshakmaz: Good call, removing this

src/components/DetailSideNav/Header/index.js
ayeshakmaz: In regular use we don't want full height (only as high as content) but in other spots (like results page) we use this to extend the header height to be the entire ssection

src/components/DetailSideNav/DetailSideNav.story.js
ayeshakmaz: Since we have multiple use cases for left and right content already (switches, badges, help popover icons), I'm inclined to leave this as is. What I don't want happening is people missing the `rightContent` prop, and trying to pass everything in within children, and trying to do the flexing themselves (likely to produce inconsistencies). I think forcing people to think in terms of left and right content will make it easier to stay consistent in the future. 

@kwalker3690 and I talked about this and landed on this version, so also curious to hear what she thinks

src/components/DetailSideNav/DetailSideNav.story.js
ayeshakmaz: I like `ManagerSideNav`. Honestly naming this thing was SUPER hard, I went with detail since I think of the pages this lives in to be detail pages (like an experiment detail page, a results detail page, etc.) so DetailSideNav seemed best, but I see what you mean about detail being on the right. I also considered `InnerSideNav` but didn't love that either

@kwalker3690 Any favorite names? How do you like `ManagerSideNav`?

kwalker3690:
> Any favorite names? How do you like ManagerSideNav or InnerSideNav?

ManagerSideNav!

> left and right content

I agree with Derek about the oddness of leftContent in isolation, but I also agree with Ayesha about the potential to abuse the contents of the nav item without left/right content, so imo we should stick with left/right content props

circAssimilate:
Thanks, @ayeshakmaz and @kwalker3690! 

Ayesha, just let me know when you're ready for another review. As for deploying, would we be able to deploy after this is merged? I can bump in Flags if so!

ayeshakmaz:
Thanks @kwalker3690! Made those updates, should be good to go @circAssimilate for another review, and yeah, once it's merged we can release right away and bump

src/components/ManagerSideNav/ManagerSideNav.story.js
```diff
@@ -0,0 +1,225 @@
...
+import ManagerSideNav from './index.js';
+import Switch from '../Switch/index.js';
+import Link from '../Link/index.js';
+
+const stories = storiesOf('Navigation/ManagerSideNav', module);
+stories
+  .addDecorator(withKnobs)
+  .addDecorator((story) => <div id="root-preview">{story()}</div>);
+
+stories
```
circAssimilate, line 16: Perhaps as a follow-up or future improvement, I think it may be helpful to see two more stories that showcase Web and Full Stack (pre-flags) start / stop button options. I believe I see that the Header can accept any children and that's how it would be done, but could be helpful for future discovery.

ayeshakmaz: Ooo yes, great idea! I can add those later this week or next

PR #1332, opened 2020-05-15
feat(Sheet) Add prop to Sheet to center title

ryn-optimizely:
### Summary
Adds a prop to `Sheet` to center the title.

### Tests
Adds unit test

### New designs:
Screen Shot 2020-05-15 at 1 28 50 PM


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ This file is similar to the format suggested by [Keep a CHANGELOG](https://githu
 - [Patch] Adjust spacing between drag handle and order number for **Tokens** ([#1328](https://github.com/optimizely/oui/pull/1328))
 - [Patch] Fix the **SearchPicker** results count ([#1330](https://github.com/optimizely/oui/pull/1330))
 - [Patch] Throw an error in SearchPicker's searchFunction Promise chain if the response is not an array. This will prevent a non-Array value from being set as defaultResults ([1325](https://github.com/optimizely/oui/pull/1325))
+- [Feature] Add prop to **Sheet** to center the title text ([1332](https://github.com/optimizely/oui/pull/1332))
```
ayeshakmaz, line 14: Can you add the prop name here? like "Add `centerHeader` prop to..."

src/components/Sheet/index.js
```diff
@@ -87,6 +97,7 @@ Sheet.propTypes = {
 };
 
 Sheet.defaultProps = {
+  centerTitle: false,
```
ayeshakmaz, line 100: If the title is centered, we should also center the subtitle. Maybe this would be better as `centerHeader` or `centerTitles` so that it's clear it applies to both?

Screen Shot 2020-05-15 at 10 41 41 AM

src/components/Sheet/Sheet.story.js
```diff
@@ -171,4 +171,51 @@ stories
         </Fieldset>
       </Sheet>
     </div>
+  ))
+  .add('With centered title', () => (
+    <div>
+      <p>This is text behind the sheet that is blocked by the overlay.</p>
+      <Sheet
+        title={ text('title', 'This is a Sheet') }
+        centerTitle={ boolean('centerTitle', false) }
```
ayeshakmaz, line 180: Should this be true by default for this particular Story?

src/components/Sheet/index.js
```diff
@@ -16,6 +17,11 @@ const Sheet = props => {
     );
   }
 
+  let titleClasses = classNames({
```
ayeshakmaz, line 20: Similar note below, but I think we should apply this to both the title and subtitle

PR #1338, opened 2020-05-20
patch (Input) Add spellCheck prop

pauloptimizely:
## Summary

**What?**
When a user enters a feature key, such as `target_my_mf_rollout `, the spell-checker mark it has an invalid word.

I'm adding a new prop `spellCheck` to the **Input** component so that we can have the option to disable spell checking when it's necessary. 

**Why?**

The Product team requested we disable spell-checking for feature names

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 This file is similar to the format suggested by [Keep a CHANGELOG](https://github.com/olivierlacan/keep-a-changelog).
 
 ## Unreleased
+  - [Patch] Add `hasSpellCheck` prop to **Input** component ([#1338](https://github.com/optimizely/oui/pull/1338))
 
 ## 46.4.2 - 2020-06-08
 - [Patch] Allow **Disclose** to take a node or a string for `title` with `flex` ([#1354](https://github.com/optimizely/oui/pull/1354))
 - [Patch] Define missing components for typescript declarations ([#1352](https://github.com/optimizely/oui/pull/1352))
-- [Patch] Add `hasSpellCheck` prop to **Input** component ([#1338](https://github.com/optimizely/oui/pull/1338))
 
 ## 46.4.1 - 2020-05-29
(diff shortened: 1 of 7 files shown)


src/components/Input/index.tsx
```diff
@@ -299,6 +304,7 @@ Input.displayName = 'Input';
 Input.defaultProps = {
   note: null,
   isRequired: false,
+  spellCheck: true,
```
pauloptimizely, line 307: The current default behavior is that spell-checking is always true

src/components/Input/index.tsx
```diff
@@ -91,6 +91,9 @@ type InputProps = {
   /** Name of the icon to place on right side of input */
   rightIconName?: string;
 
+  /** Disables spellcheck when true */
+  spellCheck?: boolean;
```
ayeshakmaz, line 95: Should this be “disables spellcheck when false”? Could we rename it to something like hasSpellCheck to help make it more clear?

pauloptimizely: > Should this be “disables spellcheck when false”?

oh yes, i'll make that change

> Could we rename it to something like hasSpellCheck to help make it more clear?

This is sounds good to me :+1:

ayeshakmaz:
Updates look great, but looks like the snapshots need updating too now

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ This file is similar to the format suggested by [Keep a CHANGELOG](https://githu
 ## 46.4.2 - 2020-06-08
 - [Patch] Allow **Disclose** to take a node or a string for `title` with `flex` ([#1354](https://github.com/optimizely/oui/pull/1354))
 - [Patch] Define missing components for typescript declarations ([#1352](https://github.com/optimizely/oui/pull/1352))
+- [Patch] Add `hasSpellCheck` prop to **Input** component ([#1338](https://github.com/optimizely/oui/pull/1338))
```
ayeshakmaz, line 13: Sorry, looks like Github did a bad merge here, could you update this to be in the #Unreleased section

pauloptimizely: ah yes, I thought I had updated this 🤦

PR #1339, opened 2020-05-20
patch(Layout): Rewrite testSection prop

davemn:
In https://github.com/optimizely/optimizely/pull/12612 I ran into an issue with Enzyme tests and our layout components. Copied below:

---

`<Row>` and `<Col>` repeat the props they're given, so JSX like

```jsx
<Col data-test-section="hello">
  <MyComponent />
</Col>
```

ends up mounted in Enzyme like

```jsx
<Col data-test-section="hello" />
  <div data-test-section="hello">
    <MyComponent />
  </div>
</Col>
```

Right now, any test using e.g. `component.find('[data-test-section="hello"]')` is getting 2 results every time -- the parent component and the child div. This fails the tests. Using `.hostNodes()` helps us only match the child div & get these tests running correctly.

I feel like the "right" solution will be to change the implementations of `<Col>` and `<Row>` to rename props on the inner element in some situations.

---

...So this PR hopefully captures that "right" solution. After this change lands I would still need to update the app codebase everywhere to use e.g `<Col testSection="hello" />`.


pauloptimizely:
@davemn I'm having trouble reproducing this in Enzyme, but I remember seeing it in the monolith. What am I doing wrong on this branch: https://github.com/optimizely/oui/compare/paul/davidm/layout-test-sections--review?expand=1?

davemn:
@pauloptimizely Try this out:

```js
describe('component/Layout/Col', function() {
  it('should only render a single tests-section per Col component', function() {
    const component = mount(
      <Row testSection="selected-audiences" removeGutters={ true }>
        <Col small="fillSpace">hello</Col>
        <Col small="fillSpace">hello</Col>
        <Col small="fillSpace">hello</Col>
        <Col small="fillSpace">hello</Col>
      </Row>
    );
    expect(component.find('[data-test-section="selected-audiences"]').length).toBe(1);
  });
});
```

I'm mount `<Row>` with `testSection`, but the expectation is on `data-test-section` to demo the rewrite.

pauloptimizely:
@davemn nice that worked! I think understand what's going with these changes now. i'm going to submit a review soon

src/components/Layout/Col.tsx
```diff
@@ -100,7 +112,20 @@ const Col: React.SFC<ColPropTypes> = React.forwardRef(
     if (!spans.length) {
       spans.push(prefix); // plain 'col'
     }
-    return <Component {...props} children={children} ref={ref} className={classNames(...spans, ...classes)}></Component>;
+    const hostElemProps = {};
+    if (typeof Component === 'string' && testSection) {
+      // Change testSection to data-test-section when the root is a native element.
```
pauloptimizely, line 117: should we also set `props.testSection` to `undefined` here? I'm thinking we only pass in one type of test-section attr.

Let's me know if that's overkill

PR #1341, opened 2020-05-26
fix (DockedFooter) Simplify DockedFooter and update story

kwalker3690:
This PR refactors DockedFooter to be less reliant on internal component state, making it easier to reason about when using it elsewhere. I've also updated the DockedFooter story to reflect a more straightforward approach to implementation.

I've also tested the changes to DockedFooter with the existing story, and the component still works as expected, so these should be considered non-breaking changes


acacia-m-pappas:
Nice catch Kelly! Lgtm 👍

src/components/DockedFooter/DockedFooter.story.js
```diff
@@ -22,53 +22,47 @@ import Code from '../Code/index.js';
...
+      // 2.) Use a ref callback to set the parent ref
+      //     in state once rendered, which triggers an update
+      //     to DockedFooter, so that it can adjust its placement
+      ref={ newRef => setParentRef(newRef) }
+      data-test-section="scroll container"
+      className="height--300 overflow-y--auto">
+      {children}
+      <DockedFooter
+        // 3.) Pass the parent ref to the DockedFooter as a prop
+        scrollRef={ parentRef }
```
ayeshakmaz, line 43: I'm seeing a prop warning in storybook:

`Warning: Failed prop type: Invalid prop scrollRef supplied to DockedFooter, expected a ReactNode.`

Looks like there isn't one great way to do a ref proptype (according to the[ various answers in this stack overflow post)](https://stackoverflow.com/questions/48007326/what-is-the-correct-proptype-for-a-ref-in-react), but we could probably be pretty loose about it with something like 
```

scrollRef: PropTypes.oneOfType([
    PropTypes.func, 
    PropTypes.shape({ current: PropTypes.elemntType })
])
```

or something similar

PR #1343, opened 2020-05-26
feat (Layout) Add ref options for Col and Row

kwalker3690:
I need to use a ref on either Col or Row, so this PR updates the prop types for both to accept a ref, which is passed through to the rendered component


data/csswhat/oui.css
```diff
@@ -4959,6 +4959,10 @@ code {
   .oui-dropdown__block-link:hover .oui-dropdown--descriptive__content {
     color: var(--light-blue-75); }
   
+  .dropdown__block-link:hover .oui-dropdown__block-link--icon,
```
circAssimilate, line 4962: Are the changes in this file intended or automated?

ayeshakmaz: These are automated, I need to update the git ignore to ignore these files, but there's no harm in committing them either

circAssimilate: Thanks! Just wasn't sure since things were actually changing.

kwalker3690:
@circAssimilate changes only need to be made to `module-declaration.d.ts` (step 5) if a new file is added, this is just changes to two existing files. I did already do step 6, which is why there are updates to index.d.ts

PR #1351, opened 2020-05-29
patch(Code): Add maxLineLength prop

davemn:
Adds a new prop, `maxLineLength`, to the `<Code>` component. When specified, the component wraps each line of props.children to be at most `maxLineLength` characters long.

**Background**

Since #1294 was merged, there's no longer an automatic way to have long code samples wrap when passed to `<Code>`. My hope was that passed code would have newlines embedded in it, instead of the `<Code>` component trying to figure out the best wrapping.

That said, I now think opt-in text wrapping can be useful in some circumstances. For example, if you're displaying one code sample in two locations with different widths, there may not be one "correct" embedding of newlines in the string that looks right in both locations.

**Caveats**

This prop easily breaks syntax highlighting when given values that are too small. Fixing this would require the line wrapping function to run after the syntax highlighter, and only modify text nodes in the output. It's do-able but a bigger effort.

### DIFF
--- a/src/components/Code/index.js
+++ b/src/components/Code/index.js
@@ -23,6 +23,37 @@ Highlight.registerLanguage('scss', require('highlight.js/lib/languages/scss'));
 Highlight.registerLanguage('swift', require('highlight.js/lib/languages/swift'));
 Highlight.registerLanguage('go', require('highlight.js/lib/languages/go'));
 
+/* Helper fn. for wrapBlock() */
+function wrapLine(line, limit) {
+  if (line.length <= limit) {
+    return [line];
+  }
+  return [
+    ...wrapLine(line.slice(0, limit), limit),
+    ...wrapLine(line.slice(limit), limit),
+  ];
+}
+
+/**
+ * Given a string with 0+ lines (delimited by \n), break each line into multiple
+ * according to "limit", and returns the broken lines as a single multiline
+ * string.
+ *
+ * @param {String} s - Input string, existing newlines are preserved in the
+ *   output.
+ * @param {Number} limit - Number of characters until a line is wrapped.
+ * @returns {String}
+ */
+function wrapBlock(s, limit) {
+  if (s.length < limit) {
+    return s;
+  }
+  return s
+    .split('\n')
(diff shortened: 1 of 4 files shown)


src/components/Code/index.js
```diff
@@ -23,6 +23,37 @@ Highlight.registerLanguage('scss', require('highlight.js/lib/languages/scss'));
...
+/* Helper fn. for wrapBlock() */
+function wrapLine(line, limit) {
+  if (line.length <= limit) {
+    return [line];
+  }
+  return [
+    ...wrapLine(line.slice(0, limit), limit),
+    ...wrapLine(line.slice(limit), limit),
+  ];
+}
```
davemn, line 35: Can combine these two functions in an iterative style instead of recursive if anyone's concerned about performance. Comparing the two the version in the PR is more readable IMO, but that's subjective.

```js
function wrapBlock(s, limit) {
  if (s.length < limit) {
    return s;
  }
  return s
    .split('\n')
    .flatMap(line => {
      if (line.length === 0) {
        return [line];
      }
      const splits = [];
      let remaining = line;
      while (remaining.length > limit) {
        splits.push(remaining.substring(0, limit));
        remaining = remaining.substring(limit);
      }
      if (remaining) {
        splits.push(remaining);
      }
      return splits;
    })
    .join('\n');
}
```

ayeshakmaz:
This method isn't necessarily very conducive to changes in font size (ie user zooming for accessibility) or responsiveness (resizing the window), is there a reason you went this route instead of creating an option via a prop to do `white-space` `pre` **or** `pre-wrap`?

If I shrink my window in the storybook example, content gets cut off and I'm forced to scroll to see everything, whereas using `pre-wrap` allows me to see all content at all times. Could use both (the maxLineLength and a pre-wrap prop option) to cover all bases: max length for wide screens, but the ability to consume all content w/o scroll in narrow screens

davemn:
> is there a reason you went this route instead of creating an option via a prop to do white-space pre or pre-wrap?

Good point, I think it makes the most sense to close this and will follow up another time with a CSS-based fix.

PR #1358, opened 2020-06-13
feat(Tile): Update dropdownItems to take Dropdown ListItems

ayeshakmaz:
Three changes:

Storybook for reviewers: https://5dcc95c567dd4d002006b897-havuaaytiv.chromatic.com/?path=/story/tile--with-action-items-on-the-right

- Update **Tile** to directly render array of Dropdown.ListItem components for `dropdownItems` prop 
- Add `isDestructive` prop to **Dropdown.BlockLinkText** to render red text 
- Add `removeBorderTop` prop to **Dropdown.ListItem** to remove top border on a list item 

Screen Shot 2020-06-12 at 6 14 07 PM

Jira:
[UXENG-112](https://optimizely.atlassian.net/browse/UXENG-112)


src/components/Tile/index.tsx
```diff
@@ -38,28 +33,20 @@ const renderTileActions = (
...
 };
 
+type ListItemType =  {
+  hardSides?: boolean,
+  hardTop?: boolean,
+  ignoreToggle?: boolean,
+  removeBorderTop?: boolean,
+  role?: string,
+}
+
```
ayeshakmaz, line 49: Not sure if this is ideal, so any feedback welcome!

src/components/Tile/index.tsx
```diff
@@ -38,28 +33,20 @@ const renderTileActions = (
...
-            </Dropdown.BlockLink>
-          </Dropdown.ListItem>
-        ))}
+        {items}
       </Dropdown.Contents>
     </Dropdown>
   );
 };
 
+type ListItemType =  {
```
circAssimilate, line 42: nit: I would maybe name this `ListItemTypeProps`

src/components/Tile/index.tsx
circAssimilate: Should we also define defaults here or below since they're not required?

types/index.d.ts
```diff
@@ -757,6 +757,10 @@ declare module "components/Layout/Col" {
          * )}
          */
         small?: column;
+        /**
+         * How to vertically align self
+         */
+        alignSelf?: 'start' | 'center' | 'end';
```
circAssimilate, line 763: Looks like this is maybe coming through from a previous PR?

ayeshakmaz: yeah this is from a previous one

ayeshakmaz:
Thank for taking a look Derek, made that naming update. As for the default props, that should be taken care of by Dropdown.ListItem itself, since these props aren't actually referenced w/i Tile. However, this might also mean that the type for that prop is not ideal? Not entirely sure with my limited experience w/ Typescript

PR #1360, opened 2020-06-23
[AXIOM-26] Add SummaryBar Component

Elepert:
## Summary

Added new component called `SummaryBar` which allows for highlighting information in an organized fashion.
Screen Shot 2020-06-24 at 5 32 24 PM
It allows for 2 to 5 columns of information and has the option to include tooltips for column headers and content.

Also updated the 11px font variable to be 12px.

##  Test Plan

Added a new automated test file as well as manually tested in Storybook.

## JIRA

[AXIOM-26](https://optimizely.atlassian.net/browse/AXIOM-26)


src/components/SummaryBar/index.scss
```diff
@@ -0,0 +1,20 @@
+
```
ayeshakmaz, line 1: Could you convert this into using[ BEM syntax](https://css-tricks.com/bem-101/)? Might look something like:

`.oui-summary-bar`, `.oui-summary-bar__header-box`, `.oui-summary-bar__help-popover`

src/components/SummaryBar/index.scss
```diff
@@ -0,0 +1,20 @@
...
+    border-bottom: solid ;
+    border-width: 1px;
+    border-color: $smoke-dark;
+    padding-bottom: 15px;
+
+    .header-box {
+        border-bottom: solid ;
+        border-width: 1px;
+        border-color: $smoke-dark;
+        margin-bottom: 15px;
```
ayeshakmaz, line 13: See comments below on how this CSS can all be done using the Row component itself already

src/components/SummaryBar/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+            pushRowsTop={ false }
+            pullRowPadding={ false }
+            paddedContent={ 'around' }
+            fluid={ true }>
+            <Row shouldWrap={false}>
+                {columnHeaders}
+            </Row>
+            <Row shouldWrap={false}>
+                {columnContent}
+            </Row>
```
ayeshakmaz, line 116: Instead of needing to add borders using `border--bottom` and needing to add padding and margin bottom to the wrapper and header box, this first row can have border="top" as a prop, and the second row can add border="bottom"

However, for the spacing, you will need to update the paddedContent prop on row to support "top" and "bottom" and add that corresponding CSS (for rows only for now, don't worry about Col) to `src/components/Layout/index.scss` around lines 8-24

This way your rows would end up looking like:

`<Row shouldWrap={false} border="top" paddedContent="top">
                {columnHeaders}
            </Row>
            <Row shouldWrap={false} border="bottom" paddedContent="bottom">
                {columnContent}
            </Row>`

src/components/SummaryBar/index.scss
```diff
@@ -0,0 +1,20 @@
...
+    }
+
+    .title-help-popover-container {
+        padding-top: 2px;
+        height: 19px;
```
ayeshakmaz, line 18: Where does this 19px come from? Is this necessary?

src/components/SummaryBar/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+};
+
+export const SummaryBar: React.SFC<SummaryBarProps> = React.forwardRef(
+    (props, ref?: React.Ref<HTMLInputElement>) => {    
+    const { title, columns, extraInfo, testSection } = props;
+    const numColumns = columns.length < 6 ? columns.length : 5;
+    let childrenCount = 0 
+    const columnHeaders = columns.map((child) => {
+        if (childrenCount < numColumns) {
+            childrenCount +=1;
```
ayeshakmaz, line 47: I believe you can simplify this logic by using Array.slice() on columns so that you only ever care about the first 5 items:

`let columnsToRender = columns.slice(0, 5);
columnsToRender.map().... `

src/components/SummaryBar/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+            let border;
+            if (childrenCount !== numColumns) {
+                border = 'right'
+            }
```
ayeshakmaz, line 51: you can the instead get the idx from map (`columns.map(child, idx)`) to do this check

src/components/SummaryBar/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+            return (
+                <Col border={border} data-test-section={`${testSection}-col-${childrenCount}-header`}>
+                    <div className={"min-width--100"}>
```
ayeshakmaz, line 54: The storybook alignment seems off here, I think you're missing the `flex` class?

Screen Shot 2020-06-25 at 3 48 43 PM

src/components/SummaryBar/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+        }
+    })
+    childrenCount = 0
+    const columnContent = columns.map((child) => {
+        if (childrenCount < numColumns) {
+            childrenCount +=1;
+            let border;
+            if (childrenCount !== numColumns) {
+                border = 'right'
+            }
```
ayeshakmaz, line 75: Same comment here, this can be simplified by slicing

src/components/SummaryBar/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+            return (
+                <Col border={border} data-test-section={`${testSection}-col-${childrenCount}-content`}>
+                    <div className={"min-width--100"}>
```
ayeshakmaz, line 78: Alignment issue here as well

src/components/SummaryBar/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+        if (childrenCount < numColumns) {
+            childrenCount +=1;
+            let border;
+            if (childrenCount !== numColumns) {
+                border = 'right'
+            }
+            return (
+                <Col border={border} data-test-section={`${testSection}-col-${childrenCount}-header`}>
+                    <div className={"min-width--100"}>
+                        {child.header && <div className={"muted micro push-half--bottom"}>{child.header}</div>}
```
ayeshakmaz, line 55: I think a `weight--bold` class is needed here

src/components/SummaryBar/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+  }
+);
+
+SummaryBar.displayName = "SummaryBar";
+
+SummaryBar.propTypes = {
+    /** Adds an optional label if there is a label provided
+     *  @param {Object} props Object of props
+     *  @returns {Error} Error or null
+     */
```
ayeshakmaz, line 129: Not sure these comments make sense? Maybe copy pasta? 🍝

src/components/SummaryBar/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+    return (
+      <div ref={ref} className={"width--1-1 summary-bar-wrapper"} data-test-section={testSection}>
+          <div className={"header-box flex flex-align--end"}>
+            <div className={"color--base push-half--bottom push-double--right gamma"}>{title}</div>
+            {extraInfo && <p className={"muted micro anchor--right push-half--bottom"}>{extraInfo}</p>}
+          </div>
+          <Container
+            outlineDebug={ false }
+            pushRowsTop={ false }
+            pullRowPadding={ false }
```
ayeshakmaz, line 108: You can remove all these since the defaults are false

src/components/SummaryBar/tests/index.js
```diff
@@ -0,0 +1,122 @@
...
+        .find('[data-test-section="summary-bar-test"]')
+        .find('[data-test-section="summary-bar-test-col-6-header"]')
+        .exists()).toBeFalsy();
+    expect(
+      component
+        .find('[data-test-section="summary-bar-test"]')
+        .find('[data-test-section="summary-bar-test-col-6-content"]')
+        .exists()).toBeFalsy();
+  });
+});
```
ayeshakmaz, line 122: Would be great to add a few tests around the help popovers, just to ensure they render when the prop is provided

types/templates/module-declaration.d.ts
```diff
@@ -59,6 +59,7 @@ declare module 'optimizely-oui' {
   export { default as Sortable } from 'components/Sortable/index';
   export { default as Spinner } from 'components/Spinner/index';
   export { default as Steps } from 'components/Steps/index';
+  export { default as SummaryBar } from 'components/SummaryBar/index';
```
ayeshakmaz, line 62: Don't forget to also run `yarn generate-types`

src/components/SummaryBar/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+    /** Header of column */
+    header?: string;
+    /** Content of help popover for header */
+    headerHelpTooltip?: string;
+    /** Content of column */
+    bodyContent? : {
+        /** Content value */
+        value: string,
+        /** Color of content */
+        color: 'red' | 'green' | 'black';
```
ayeshakmaz, line 21: can we make 'black' the default, so users only have to specify when they want red or green?

src/components/SummaryBar/index.scss
ayeshakmaz: Though, with my other later comments, might not even need any custom CSS after some updates to Row

src/components/SummaryBar/index.scss
Elepert: Nope not at all, this is leftover from me trying some things.

src/components/SummaryBar/index.tsx
Elepert: I need `numColumns` and `childrenCount` for figuring out if a col gets a border on the right or not as well. So I'll do this, but I think that it would only get rid of ```if (childrenCount < numColumns) {```

Elepert: Oh nvm I see your comment below :)

src/components/SummaryBar/index.tsx
Elepert: yep, definitely

src/components/SummaryBar/index.tsx
Elepert: I'm not quite sure how to do that given that this is a type that the actual props uses and it's an array of this type. So it's sort of embedded a couple layers down if that makes sense. I can make that happen when actually describing the classNames though.

ayeshakmaz: Oh there's an example here `src/components/Layout/Col.tsx` where paddedContent is defaulted to "none"

src/components/SummaryBar/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+        return (
+            <Col border={border} data-test-section={`${testSection}-col-${idx}-content`}>
+                <div className={"flex min-width--100"}>
+                    {child.bodyContent &&  <div className={classnames(
+                        'color--base',
+                        {
+                        'beta weight--normal': child.bodyContent.isNumber,
+                        'micro': !child.bodyContent.isNumber,
+                        'color--good-news': child.bodyContent.color === 'green',
+                        'color--bad-news': child.bodyContent.color ==='red'}                        )}>{child.bodyContent.value}</div>}
```
ayeshakmaz, line 79: as discussed in slack, maybe try:

let colorClass = child.bodyContent.color === 'green' ? 'color--good-news' : child.bodyContent.color === 'red' ? 'color--bad-news' : 'color--base';

src/components/SummaryBar/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+    /** Content columns where summary info is displayed. Requires 2-5 columns. */
+    columns: SummaryBarColumn[];
+    /** Test section for element */
+    testSection: string;
+};
+
+export const SummaryBar: React.SFC<SummaryBarProps> = React.forwardRef(
+    (props, ref?: React.Ref<HTMLInputElement>) => {    
+    const { title, columns, extraInfo, testSection } = props;
+    const numColumns = columns.length < 6 ? columns.length : 5;
```
ayeshakmaz, line 43: Could simplify this to `numColumns = Math.min(columns.length, 5)`

ayeshakmaz: and maybe call this variable maxNumColumns so that line 47 and 67 are more clear

src/components/SummaryBar/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+    const columnsToRender = columns.slice(0, 5);
+    const columnHeaders = columnsToRender.map((child, idx) => {
+        let border;
+        if (idx !== numColumns) {
```
ayeshakmaz, line 47: I think there's an off-by-one error here (I'm seeing a right border on the last column)

src/components/SummaryBar/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+                    </div>}
+                </div>
+            </Col>
+        );
+    })
+    const columnContent = columnsToRender.map((child, idx) => {
+        let border;
+        if (idx !== numColumns) {
+            border = 'right'
+        }
```
ayeshakmaz, line 69: Same off-by-one error here

src/components/SummaryBar/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+                        verticalAttachment="middle">
+                            {child.bodyContent.helpTooltip}
+                        </HelpPopover>
+                    </div>}
+                </div>
+            </Col>);
+    });
+
+    return (
+      <div ref={ref} className={"width--1-1 summary-bar-wrapper"} data-test-section={testSection}>
```
ayeshakmaz, line 92: Can remove this class name now

src/components/SummaryBar/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+          <div className={"header-box flex flex-align--end"}>
```
ayeshakmaz, line 93: Can remove `header-box` here as well

src/components/SummaryBar/SummaryBar.story.js
```diff
@@ -0,0 +1,148 @@
+import React from 'react';
```
ayeshakmaz, line 1: Could you update this file to be a .tsx file so we can catch any type errors in the stories? After converting, just double check the console to make sure there are no warnings/type issues

src/components/SummaryBar/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+export const SummaryBar: React.SFC<SummaryBarProps> = React.forwardRef(
+    (props, ref?: React.Ref<HTMLInputElement>) => {    
+    const { title, columns, extraInfo, testSection } = props;
+    const maxNumColumns = Math.min(columns.length, 4);
+    const columnsToRender = columns.slice(0, 5);
+    const columnHeaders = columnsToRender.map((child, idx) => {
+        let border;
+        if (idx !== maxNumColumns) {
+            border = 'right'
+        }
```
ayeshakmaz, line 49: I'm still seeing the extra border on the right when there's 4 columns, I think this needs to be:

```
 const maxNumColumns = Math.min(columns.length, 5);
{...}
if (idx !== maxNumColumns - 1) {
            border = 'right'
        }
```

Elepert: 🔥

PR #1361, opened 2020-06-25
[UXENG-95] Added external icon to navbar

Elepert:
## Summary

Added option for external icon to NavBar elements.
Screen Shot 2020-06-26 at 10 40 14 AM

## Test Plan

Tested in Storybook.

## JIRA

[UXENG-95](https://optimizely.atlassian.net/browse/UXENG-95)


src/components/NavBar/NavBar.story.js
```diff
@@ -92,6 +94,7 @@ stories
           type="button"
           linkLabel="Program Management"
           testSection="program-management"
+          hasExternalIcon={ true }
```
ayeshakmaz, line 97: Could you update the story to only have the prop on the Getting Started and the Program management and Help links?

Since the others (experiment, projects) won't ever be external, I don't want to confuse people 😛

PR #1363, opened 2020-06-29
feat (Table) Add forward ref to table row component

kwalker3690:
This PR adds the ability to forward a ref to the Table.TR component!

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1683,30 +1683,8 @@ declare module "components/Table/THead/index" {
 }
 declare module "components/Table/TR/index" {
     export default TR;
-    function TR({ isActive, isHighlighted, noBorder, noHover, borderStyle, backgroundColor, testSection, children, }: {
-        isActive: any;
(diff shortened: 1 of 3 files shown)


types/index.d.ts
```diff
@@ -1683,30 +1683,8 @@ declare module "components/Table/THead/index" {
...
-            export const isActive: PropTypes.Requireable<boolean>;
-            export const isHighlighted: PropTypes.Requireable<boolean>;
-            export const noBorder: PropTypes.Requireable<boolean>;
-            export const noHover: PropTypes.Requireable<boolean>;
-            export const testSection: PropTypes.Requireable<string>;
-        }
-        export const displayName: string;
-    }
-    import PropTypes from "prop-types";
+    const TR: React.ForwardRefExoticComponent<React.RefAttributes<any>>;
```
davemn, line 1686: 😞  I guess there's limits to how well the Typescript compiler can autodetect types in non-TS source files.

PR #1365, opened 2020-06-30
[UXENG-74] Added built-in sorting to tables

Elepert:
## Summary

Added ability to sort table columns to Table component.
Screen Shot 2020-06-30 at 10 44 52 AM

## Test Plan

Tested in storybook and added automated tests in `Table/tests/`.

## JIRA

[UXENG-74](https://optimizely.atlassian.net/browse/UXENG-74)


src/components/Table/TH/index.js
```diff
@@ -51,6 +89,26 @@ TH.propTypes = {
 TH.defaultProps = {
   isCollapsed: false,
   isNumerical: false,
+  sorting: { canSort: false},
```
spham1412, line 92: missing a space?

src/components/Table/TH/index.js
```diff
@@ -51,6 +89,26 @@ TH.propTypes = {
...
+   *  @returns {Error} Error or null
+   */
+  sorting: function verifyHandleSortExists(props) {
+    if (props.sorting.canSort && !props.sorting.handleSort) {
+      return new Error(
+        'Must pass a sorting function.'
+      );
+    } else if (props.sorting.canSort && !props.sorting.order) {
+      return new Error(
+        'Must pass a sorting direction.'
```
spham1412, line 107: Why not default to `ASC`?

Elepert: This is up to the dev to update as the sorting order is flipped around (so when the user presses the arrow, the dev needs to have a function that changes this from 'asc' to 'desc' and back and forth). If we default to 'asc' then that means that the dev doesn't have a way to update it so the sorting arrow will always point up.

ayeshakmaz:
@Elepert Just one thing to note looking at the storybook - there seems to be weird behavior on the first click of the header, like for Status, for first click sorts properly, but the arrow doesn't change. Then the second click changes just the arrow, then the 3rd click gets it going in the right toggle state. Also the Rate sorting seems not right?

Elepert:
The reason the Rate sorting seems weird is that the values are strings so it's sorting not numerically but via string. In that case, it compares the first element, then the second, etc.. versus the plain numerical value.

PR #1366, opened 2020-07-01
feat(Toolbar): Add option to remove padding

ayeshakmaz:
With this PR, Col properly supports paddedContent="remove" and Toolbar has a new style option "tight" which removes padding

JIRA:
[UXENG-115
](https://optimizely.atlassian.net/browse/UXENG-115)


ayeshakmaz:
@kwalker3690 Here are some updates to support UXENG-115, goes along with this [PR #99](https://github.com/optimizely/flags/pull/99) in Flags showing how to use it, let me know what you think!

PR #1368, opened 2020-07-01
Initial fix for Col with Nubbin

ayeshakmaz:


circAssimilate:
One kinda related things I've run into (that I've captured in the [google doc](https://docs.google.com/document/d/1Y33iOf5mZ7SSNaayLGaQ8xDDHiGj8nPh8IYcuWvrlEc/edit#) but not UXENG Jira yet) that we may be able to tackle here. I've noticed in a three column layout that the third column can have a variable size based off of content width (see examples below). I wonder if we can have set widths for the left SideNav and the ConfigPanel (or whatever we call it) and just leave the middle Column flexible. Wdyt?

Screen Shot 2020-07-02 at 12 33 40 PM

Screen Shot 2020-07-02 at 12 33 45 PM

src/components/Layout/Col.tsx
```diff
@@ -81,17 +98,41 @@ const Col: React.SFC<ColPropTypes> = React.forwardRef(
...
+        nubbinTopPosition = nubbinRef.offsetTop + nubbinRef.offsetHeight / 2;
+      }
+      children = (
+        <>
+          <div
+            className="oui-config-panel__nubbin"
+            style={{ top: nubbinTopPosition }}
+          ></div>
+          {children}
+        </>
```
circAssimilate, line 120: nit: does it work with the nubbin at the end of all these elements? Seems to in the elements panel for me and think that's just a little clearer since it's visually on the right.

src/components/Layout/Templates.story.tsx
```diff
@@ -8,69 +8,139 @@ import Token from '../Token';
...
-                  ref={ this.pointingRef }>
-                  The nubbin should point in the middle of this area
+                <div className="push-double--bottom">
+                  <p>Make the Nubbin point to:</p>
+                  <Radio
+                    name="nubbin"
+                    onChange={this.switchRef.bind(this, 'tile')}
+                    label="the Tile"
+                    defaultChecked={true}
+                  />
```
circAssimilate, line 77: Is there a world where we have a higher order component that can add this nubbin pointer with ref around a specific element and avoid having to set up all these refs manually? Thinking this may be especially important with a dynamic list of things that will need these refs (like rules). 

There is absolutely a better version of this but off the top of my head we could do something like:

```
<div>
  {items.map(item => (
    this.state.selectedItemId === item.id ? withNubbin(<Radio { ...item } />) : <Radio { ...item } />
  })}
</div>
```

ayeshakmaz:
@circAssimilate The column issue width could be fixed by doing <Col small={6}/> on both sides, which would maintain 50% for each at all times, so that they both get wider on large screens. But curious what @zleach thinks about keeping the Config panel a set width ( like 550px? ) It might make more sense to set that on the inner div within the Col, since supplying fixed widths to Col might conflict with the idea of flexbox and the goal of a Col

PR #1379, opened 2020-07-15
RangeSlider: Port to Typescript; fix background color

davemn:
Summary

On backgrounds that aren't white, improve the appearance of the right half of the range slider by using a second SVG for the fade overlay.

JIRA:
[UXENG-136](https://optimizely.atlassian.net/browse/UXENG-136)


ayeshakmaz:
@circAssimilate Good call, made that update - thanks!

PR #1381, opened 2020-07-21
feat(storybook): Re-organize stories into folders

ayeshakmaz:
Organize our components into logical folders for better discovery


src/components/Link/Link.story.js
```diff
@@ -107,23 +107,23 @@ stories
...
-  })
-  .add('Disabled', () => {
-    return (
-      <Link
-        isDisabled={ true }>
-        Faux Link
-      </Link>
-    );
-  });
+// storiesOf('Link (Fake)', module)
```
Elepert, line 110: Should this be uncommented?

ayeshakmaz: Oh I commented this out for now until I can evaluate it, because to me, this is just blue text, and it's misleading to make something look like a Link if it won't actually go anywhere...

PR #1383, opened 2020-07-21
[UXENG-86] Added Icon option to buttons

Elepert:
## Summary

Added option to include an icon in a button. Can specify which side of the button to put the icon.

## JIRA

[UXENG-86](https://optimizely.atlassian.net/browse/UXENG-86)


src/components/Button/index.js
```diff
@@ -74,8 +88,11 @@ const Button = ({
       aria-live="polite"
       title={ title }
       ref={ buttonRef }>
-      { isLoading && <Spinner data-test-section="button-spinner" size="tiny"/> }
+      { isLoading ?
+        <div><Spinner data-test-section="button-spinner" size="tiny"/></div>
```
ayeshakmaz, line 92: Looks like the addition of this div is causing a bug, is this needed?

PR #1386, opened 2020-07-27
[UXENG-131] Add ability to add icons to button links

Elepert:
## Summary

As part of work for [UXENG-131](https://optimizely.atlassian.net/browse/UXENG-131) we need to add the ability to make buttons that double as links have icons. This is a continuation of [this PR](https://github.com/optimizely/oui/pull/1383) which allows regular buttons to have icons.

## Jira

[UXENG-131](https://optimizely.atlassian.net/browse/UXENG-131)


src/components/Button/Button.story.js
```diff
@@ -107,18 +107,17 @@ stories
...
           <Button
-            key="1"
+            key="2"
             style="highlight"
             onClick={ action('I have been clicked') }
             rightIcon='external'>
           Right icon
           </Button>] }
-        rightGroup={ [
+        centerGroup={ [
```
ayeshakmaz, line 118: Might be best to keep these all in the centerGroup, like the story above. Right now the spacing looks funky in storybook

src/components/Button/Button.story.js
```diff
@@ -162,6 +161,20 @@ stories
...
+        rightGroup={
+          [
+            <Link
+              key="4"
+              newWindow={ true }
+              href={ text('href', 'http://google.com') }>
+              <Button
+                style="outline"
+                width='default'
+                isLink={ boolean('isLink', true) }
```
ayeshakmaz, line 173: can you remove the knob from this, and just have it be true

PR #1398, opened 2020-08-11
feat(tokens): Update Color design tokens

ayeshakmaz:
Summary:

- Add new extra light tokens
- Rename amber to Yellow

JIRA:
[UXENG-150](https://optimizely.atlassian.net/browse/UXENG-150)


src/utils/accessibility.js
```diff
@@ -117,7 +117,7 @@ export const keyboardTracker = Component => {
 // map fillColorName prop values to OUI color tokens
 export const FILL_COLOR_MAP = {
   aqua: aquaDark,
-  amber: amberDark,
+  amber: yellowDark,
```
quarklemotion, line 120: is there a specific reason that this `amber` key wasn't changed as well?  I assume maybe due to application code already using this key?

ayeshakmaz: Yeah, I was torn both ways. This is a breaking change already, so I could just update this to yellow... I'll make that update

PR #1403, opened 2020-08-18
patch(HelperClasses): Enforce flex classes

ayeshakmaz:
## Summary

Added !important to flex classes to ensure they are enforced

## Changelog Entry

[Patch] Updated Flex helper classes to ensure they are enforced

## Test Plan

Manual/VRT

## Jira

N/A


jessicalc:
At first glance, this LGTM, but do you have a link to the PR/commit in which the bug was introduced?

ayeshakmaz:
@jessicalc It was this one https://github.com/optimizely/optimizely/pull/13134
The order of CSS changed and now it needs the !important

PR #1417, opened 2020-09-14
fix(SelectDropdown): Fixed Label click firing twice in Multiselect mode

zashraf1985:
## Summary

When `multiSelect` is enabled, the `onClick` event fires twice when checkbox label is clicked. It fires once when actual checkbox is clicked directly. Fixed it by moving the event from `div` to `Checkbox` when `multiSelect` is enabled.

## Changelog Entry
[Patch] Fixed `DropdownBlockLink` firing `click` event twice when Checkbox label is clicked.

## Test Plan
- Manually tested thoroughly.
- Modified existing unit test to test the fix.

## Jira

### DIFF
--- a/src/components/Dropdown/DropdownBlockLink/index.js
+++ b/src/components/Dropdown/DropdownBlockLink/index.js
@@ -44,11 +44,11 @@ export const DropdownBlockLink = props => {
       style={ styleProps }
       { ...(props.testSection ? { 'data-test-section': props.testSection } : {}) }
       { ...(props.trackId ? { 'data-track-id': props.trackId } : {}) }
-      { ...(!props.isMultiSelect ? {onClick} : {}) }
+      { ...(!props.isMultiSelect ? { onClick } : {}) }
       onMouseEnter={ onMouseEnter }
       onMouseLeave={ onMouseLeave }>
       {props.isMultiSelect ? (
-        <Checkbox onChange={onClick} defaultChecked={ props.isItemSelected } label={ props.children } isDisabled={ false } />
+        <Checkbox onChange={ onClick } defaultChecked={ props.isItemSelected } label={ props.children } isDisabled={ false } />
       ) : (
         props.children
       )}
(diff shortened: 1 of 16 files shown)


spham1412 (approved):

src/components/Dropdown/DropdownBlockLink/index.js
```diff
@@ -44,11 +44,11 @@ export const DropdownBlockLink = props => {
       style={ styleProps }
       { ...(props.testSection ? { 'data-test-section': props.testSection } : {}) }
       { ...(props.trackId ? { 'data-track-id': props.trackId } : {}) }
-      onClick={ onClick }
+      { ...(!props.isMultiSelect ? { onClick } : {}) }
       onMouseEnter={ onMouseEnter }
       onMouseLeave={ onMouseLeave }>
       {props.isMultiSelect ? (
-        <Checkbox defaultChecked={ props.isItemSelected } label={ props.children } isDisabled={ false } />
+        <Checkbox onChange={ onClick } defaultChecked={ props.isItemSelected } label={ props.children } isDisabled={ false } />
```
jessicalc, line 51: This looks good to me, but could you please add a test to ensure that when the label is clicked, the onClick method only gets called once?

jessicalc (approved):

src/components/Dropdown/DropdownBlockLink/index.js
zashraf1985: Its a good idea but i am facing a problem here. This problem occured because of the way DOM propagates events. enzyme `simulate` does not work that way. I am not able to get the handler triggered when i simulate clicks on the outer divs.

jessicalc: Ah, interesting, yeah I just found [this](https://github.com/enzymejs/enzyme/issues/1551) which sounds similar. Maybe we can file a ticket/issue to followup on this.

I don't want to block if this is associated w/ a tight deadline, and I validated that the click handler only gets fired once on your PR (and twice on devel!!) in storybook, that's enough for me. 

SHIP IT!

PR #1423, opened 2020-09-28
feat(Badge): Added Background Color support to Badge

zashraf1985:
## Summary
Added Background Color support to `Badge`. I needed a darker version of grey for a project, added support for Background color which can take these additional colors as `backgroundColor`.
1. aqua
2. yellow
3. green
4. orange
5. pink
6. red
7. magenta
8. grey
9. purple
10. black

## Changelog Entry

[Feature] Updated **Badge** with new `backgroundColor` to support additional colors.

## Test Plan
1. Manually tested thoroughly.
2. All existing unit tests pass.

### DIFF
--- a/src/components/Badge/index.js
+++ b/src/components/Badge/index.js
@@ -2,28 +2,45 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 
+import { FILL_COLOR_MAP } from '../../utils/accessibility';
+
 /**
  * Tiny inline component used to draw attention to an item's state or status.
  * @param {Object} props - Properties passed to component
  * @returns {ReactElement}
  */
-const Badge = (props) => {
+const Badge = ({
+  backgroundColor,
+  children,
+  color,
+  testSection,
+}) => {
   let classes = classNames({
     'oui-badge': true,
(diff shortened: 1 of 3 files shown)


src/components/Badge/index.js
```diff
@@ -2,28 +2,45 @@ import React from 'react';
...
+    [`oui-badge--${color}`]: color,
   });
 
+  // ensure valid backgroundColor name
+  // (in the case that propType errors are ignored)
+  const badgeFillColor =
+    backgroundColor && Object.keys(FILL_COLOR_MAP).includes(backgroundColor)
+      ? FILL_COLOR_MAP[backgroundColor]
+      : null;
+
```
spham1412, line 29: Based on this logic, we're using both `color` and `backgroundColor` if both are provided through props. In this case, how are we handling it?

I prefer to use `color` first, if `color` is not provided, then use `backgroundColor`. If both are provided, what happens?

spham1412 (changes requested):

src/components/Badge/index.js
zashraf1985: Added a check to only use `backgroundColor` when its available and `color` is `default`

spham1412 (approved):

PR #1425, opened 2020-10-01
performance (DragAndDrop) Memoize DraggableItem props

pauloptimizely:
## Summary

When there's a list of draggle-items when should only re-render the items that need to get re-render.

Memoizing the draggable-items resulted in a noticeable increase in performance on our pages.

## Test Plan

I tested locally using `yarn link`

## Jira

- [APPX-2251](https://optimizely.atlassian.net/browse/APPX-2251)


circAssimilate (approved):
Nice! Ship it!!

PR #1427, opened 2020-10-02
patch(Tile)[APPX-2261]: Add unsaved changes indicator

circAssimilate:
## Summary

This adds a prop and accompanying story to the Tile component that allows for an unsaved indicator if `unsavedChangesText` is a non-empty string. Whatever text is provided will be shown in a `Poptip`. Direction from design is that when used in places like Flags, this unsaved indicator should only be shown when there is no warning content.

## Screeenshot
image

## Changelog Entry

[Patch] Updated **Tile** with new `unsavedChangesText` prop to indicate a changed item

## Test Plan

- Add unit tests
- Ensure tests pass

## Jira

- [APPX-2261](https://optimizely.atlassian.net/browse/APPX-2261)


src/components/Tile/index.tsx
```diff
@@ -187,6 +194,29 @@ const Tile = ({
...
+          data-test-section={`${testSection}-unsaved-changes-indicator`}
+        >
+          <Poptip
+            content={unsavedChangesText}
+            isAnimated={false}
+            position="bottom"
+            trigger="mouseenter"
+          >
+            <span
+              style={{
```
quarklemotion, line 209: was thinking that a class for this added to one of the CSS files might be better for performance ... but I guess practically there shouldn't be many of these spans being rendered simultaneously.  So I kinda talked myself out of needing a change here 😂

quarklemotion (approved):
LGTM!

src/components/Tile/index.tsx
circAssimilate: Ah, I had not considered the performance difference between style and a class. But I do think you're right that the amount rendered will be less due to them needing changes.

quarklemotion: 👍  the performance issue here just being DOM bloat/duplication having those inline styles repeated for each circle

circAssimilate: Ah, I see.

PR #1428, opened 2020-10-13
fix (Sidebar) Allow sidebar header prop to be a node

kwalker3690:
## Summary

I have a usecase for having a node as the projectName prop in ManagerSideNav, so this PR updates the props to allow a node.

## Changelog Entry

[Patch] Updated **ManagerSideNav** to allow a node for projectName prop

## Test Plan

Tested in conjunction with my changes


circAssimilate (approved):
I think we'll need to re-run `yarn generate-types` since the prop has changed. But approved otherwise!

PR #1432, opened 2020-10-20
update(Sidebar): Allow sidebar to be resizable on horizontal axis

oakbani:
## Summary
This PR makes use of a library called re-resizable and makes our oui Sidebar resizable. 
The sidebar now accepts these new props.
- isResizable
- minWidth
- maxWidth

If the anchor is left, the sidebar would be resizable from right and vice-versa. 

## Changelog Entry

[Patch] Updated Sidebar with new props to make it resizable horizontally.

## Test Plan
- Added unit tests
- Tested Manually

## Jira
- OASIS-7211


src/components/Sidebar/Sidebar.js
```diff
@@ -58,14 +62,46 @@ const Sidebar = (props) => {
...
+        className={ classes }
+        style={ styles }>
+
+        { isOpen && children }
+      </div>);
+  }
+
   return (
-    <div
+    <Resizable
```
spham1412, line 94: Let's add a check for `isResizable`. If that's `false`, then we should return a simple `div` as before.

spham1412 (changes requested):

oakbani:
[Patch] Updated Sidebar with new props to make it resizable horizontally.

spham1412 (approved):

src/components/Sidebar/Sidebar.js
```diff
@@ -58,12 +62,45 @@ const Sidebar = (props) => {
...
+    );
+  }
+
   return (
     <div
       data-oui-component={ true }
       data-test-section={ testSection }
       className={ classes }
       style={ styles }>
+
```
spham1412, line 103: Remove this empty line please

PR #1374, opened 2020-07-09
[UXENG-13] Created proper Fieldset and Form components

Elepert:
## Summary

Created a Fieldset component that is properly accessible. As part of this it was determined that creating a new Form component would help promote the proper use of the Fieldset component (people were misusing it because of the automatic spacing it provided). This PR creates a new Fieldset component and a new Form component. The form component has three sub-components: Section, Row, and Item. The Fieldset component also has the Row and Item components (they're the same one).

## Testing

Wrote new automated tests and tested visually in Storybook. For the Form component, I was able to replicate forms on app.optimizely.com in Storybook. I checked as best as I could the backwards compatibility of replacing the old Fieldset component with this new one and everything should be fine.

## JIRA

[UXENG-13](https://optimizely.atlassian.net/browse/UXENG-13)

PR #1437, opened 2020-11-09
fix (Form) Export form component

kwalker3690:
## Summary

A Form component [was added ](https://github.com/optimizely/oui/pull/1374) a few months ago, but it was never exported from the repo, so it is not currently usable.

## Changelog Entry

[Patch] Exported **Form** from src/main.js

## Test Plan

N/A


circAssimilate (approved):

circAssimilate:
Don't forget to generate types if we need to!

PR #1438, opened 2020-11-13
fix (ButtonIcon) Add type button to ButtonIcon

toddseller:
## Summary

Add `type="button"` to the ButtonIcon. Buttons will default to `type="submit"` when inside a form. This was causing issues where removing a ButtonIcon would cause the form to submit and prevent the user from making additional changes.

## Changelog Entry

[PATCH] Updated **ButtonIcon** to include `type="button"`

## Test Plan

N/A

## Jira

Please list and link the associated JIRA ticket

- [APPX-2239](https://optimizely.atlassian.net/browse/APPX-2239)


circAssimilate (changes requested):
Please run `yarn generate-types` then you're good to go! I'll deploy with my other change once that's approved.

circAssimilate (approved):
Oh, this is my mixup. I was thinking you were adding a new prop but you’re just using an unused one. Ignore me!

PR #1440, opened 2020-11-17
Misc. Design polish for Dropdown and Input

circAssimilate:
## Summary

In order to continue addressing visual bugs that came up from design in the [Flags bug bash](https://docs.google.com/document/d/1ZtTUfcJbSEImiNPR4qxT8BQ82-23gOfUb19ZLaXS-aM/edit#), this PR will make the following changes.

- Updated **DropdownBlockLinkText** with new `isItemSelected` prop to allow for black text
- Updated **Dropdown** with new `isLabelTight` styling option
- Updated **Input** with new `isAutoCompleteEnabled` prop to allow for disabling of autocomplete
- Updated **.oui-tile** class `overflow: hidden` style to prevent content from overlapping

## Test Plan

- Update snapshot
- Ensure tests pass


src/components/Input/index.tsx
```diff
@@ -180,15 +192,16 @@ const Input: React.SFC<InputProps> = React.forwardRef(
...
+        "oui-text-input",
+        { "oui-text-input--read-only": isReadOnly },
+        { "oui-text-input--search": isFilter },
+        { "oui-form-bad-news": displayError },
         { [`text--${textAlign}`]: hasAlignStyle }
       );
 
       let input = (
         <input
+          autoComplete={isAutoCompleteDisabled ? "off" : "on"}
```
kwalker3690, line 204: I think it'd make more sense to have this prop be `enableAutocomplete`, and have it default to `true`. iirc there were guidelines on prop convention for OUI that i think this falls more in line with, but I can't find them written anywhere :/, wdyt?

circAssimilate: I could see that and was thinking about that at first too. I just didn't see any `isXyzEnabled`props around and thought it may be easier to think about disabling things only. I can update it to `isAutoCompleteEnabled` and default to true if you prefer.

kwalker3690: That feels more in line with `isFilter`, `isOptional` `isReadOnly` etc to me!

circAssimilate: Ah, totally. All ready for another review!

kwalker3690 (approved):

PR #1441, opened 2020-11-17
feat(Tile): Replace onResultsClick with resultsLink

circAssimilate:
## Summary

Tile's results link was kind of frustrating to interact with because it didn't show a link preview and the results page is so slow. This update will make the interface simpler and allow for command + click opening in a new tab as an option.

## Screenshot 
image

## Test Plan

- Update tests
- Ensure tests pass


kwalker3690 (approved):
Great improvement, thanks!

PR #1446, opened 2020-12-03
fix (Tile) Fix tile layout

kwalker3690:
## Summary

Before:
image

After:
image

## Changelog Entry

[Patch] Fixed layout issue in **Tile**

## Test Plan

Visual testing


circAssimilate (approved):
Ugh, I totally did this and didn't catch it. I'm sorry and thank you!

PR #1448, opened 2020-12-10
fix(Dropdown):Update padding to be 10px

mikelondon:
## Summary

What did you change and why?
Updated padding to be 10px per the request from design on Flags UI.

## Screenshots 

Before:
image
After:
image

## Changelog Entry

[Patch] Updated **Dropdown** to have 10px padding when isLabelTight === True

## Test Plan

- Ensured tests passed locally 
- Manual tested.


circAssimilate (approved):
LGTM!

PR #1451, opened 2020-12-11
Update Slack notification settings in .travis.yml

circAssimilate:
Update Slack notification settings. This was copied from https://docs.travis-ci.com/user/notifications/ and I believe this is necessary via travis.com and the old format was only working for travis.org.


jeffreychan-optimizely (approved):
Seems legit - interesting that you don't have to specify a room but there's an example in the link you posted where you don't have to and it seems like one wasn't specified before...

circAssimilate:
Thanks, Jeff. That is interesting - and perhaps this won't actually make a difference. I'm going to try it out and see if it helps though.

PR #1455, opened 2021-02-03
update(Tile): Truncate long name and description if it's a string.

spham1412:
## Summary

- This PR added the function `truncate()` to shorten long `name` and `description` props of the `Tile` component. Only truncate if `name` and `description` are of type `string`.
- Also added the `.prettierrc` file to include new rules for lining, mainly changing from double-quoted strings to single-quoted strings.

image

## Changelog Entry

- [Patch] Updated **Tile** with truncated long `name` and `description`.

## Test Plan

- Manually tested.


.eslintrc.js
```diff
@@ -215,14 +212,10 @@ module.exports = {
...
       "nonEmpty": "after-props",
     }],
-    "react/jsx-curly-spacing": ["warn", "always", {"spacing": {
-      "objectLiterals": "never",
-    }}],
     "react/jsx-equals-spacing": ["warn", "never"],
     "react/jsx-filename-extension": ["warn", { "extensions": [".js"] }],
     "react/jsx-first-prop-new-line": ["warn", "multiline"],
     "react/jsx-indent": ["warn", 2],
-    "react/jsx-indent-props": ["warn", 2],
```
pauloptimizely, line 225: Did you run into eslint issues?

.prettierrc
```diff
@@ -0,0 +1,7 @@
+{
+  "jsxBracketSameLine": true,
+  "semi": true,
+  "singleQuote": true,
+  "trailingComma": "all",
+  "tabWidth": 2
+}
```
pauloptimizely, line 7: thanks for adding this in!

pauloptimizely (approved):

jeffreychan-optimizely (approved):

.eslintrc.js
spham1412: No, These deleted rules were duplicated as the linter pointed out from my IDE, that's why I deleted them.

